PR #75, opened 2018-07-31
aws nexrad mirroring

mjip:
A pair of plugins that polls/downloads https://s3.amazonaws.com/noaa-nexrad-level2/index.html


petersilva:
You should include the sample configuration files as well... they help with understanding how to use the plugin.

mjip:
Pushed some examples to demonstrate how to use the plugin.

PR #87, opened 2018-08-20
changed sftp/ftp default file index behaviour

mjip:
Changed ls implementations for SFTP/FTP to take ls index of [8:] to be the name of the listing, instead of the last entry by default. Updated corresponding documentation


mjip:
Added a CHANGE entry in the UPGRADING.rst. Yep, it's to improve on the ls implementation to allow directories with whitespace characters and default/common ls listings to require no config option, but uncommon/older versions will need this specification.

PR #89, opened 2018-08-22
sftp/ftp determines file_index on connection, doesn't need it explicitly set

mjip:
The file_index of an ls listing gets determined upon connection. Adds ~0.2s of overhead with every connection (every time sr_poll wakes up), but resolves any issues with whitespace characters in filenames and there is no longer a burden on the end user to determine where the file index is in the SFTP/FTP server they're connecting to. 

Still backwards compatible with the ls_file_index config option, but can now be safely phased out.

### DIFF
--- a/doc/UPGRADING.rst
+++ b/doc/UPGRADING.rst
@@ -38,12 +38,6 @@ Installation Instructions
 git origin/master branch
 ------------------------
 
-*CHANGE*: sr_poll ls now takes the [8:]th index instead of the -1 index to be 
-          the name, resolves bug with names containing spaces. If SFTP/FTP server 
-          being polled is not an Open BSD variant, an *ls_file_index* might need 
-          to be specified in the poll config to get previous behaviour. See sr\_
-          poll.1.rst for more info.
-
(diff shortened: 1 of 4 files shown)


mjip:
If it was saved as a cache file, and the destination changed between stopping/starting, then potentially the file_index could be outdated from what I understand. I was thinking of using an sr_poll variable instead to store this (self.dest_file_index, or something similar), since ls_file_index also gets used in load_ls_file. Not sure if there would be any conflicts when resetting/cleaning up.

petersilva:
ok... but folks change their software really, really rarely (like every
5-10 years? in terms of a one minute poll, five years is every 2.6 million
polls), and that is on the high end of the frequency expectation.  Some
people want to connect at sub-second intervals. It seems really expensive
to do the check every time... I guess you could do something like check if
the file is more than an hour old, and then re-create... that would still
leave, potentially upto an hour after a software change of ... confusion...
but probably worth it, in terms of replacing huge overhead of a check on
every connection, and also recovering from the extremely rare event that
software changes.

Issue #31, opened 2018-06-02
Extend include directive to support globbing


petersilva:
e.g.
include /etc/sarra/ninjo/subscribe/conf.d/*.conf

From a package maintenance perspective, this feature would be extremely useful as a way of deploying standardized configurations but still allow local customizations with little risk of being clobbered by subsequent upgrades.

imported #62 from sourceforge.net.

PR #93, opened 2018-08-27
feature added: config include directive support for globbing #31

mjip:
Config files can now interpret globbing arguments such as `include *.conf`. Any files that contain *, ?, [ or ] in their names need to be escaped.

### DIFF
--- a/doc/UPGRADING.rst
+++ b/doc/UPGRADING.rst
@@ -38,6 +38,10 @@ Installation Instructions
 git origin/master branch
 ------------------------
 
+*CHANGE*: sr_config include option now supports globbing. If your config filename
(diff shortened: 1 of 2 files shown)


mjip:
By escaped, I mean they need to be encased in brackets. E.g. If a filename contains '?', in order to match on it the '?' must appear as '[?]'.

mjip:
It's set up now to look into the config dir (in the program subdir) for matching files if none are found in the current dir. This can be swapped fairly easily if the other scenario is better form/preferred instead, e.g. look for matching files in ~/.config/sarra/program_dir/, if nothing matches then look in the current dir.

Issue #54, opened 2018-06-04
optimize using extended file attributes


petersilva:
if a file as an extended attribute:

last_sum=<mtime>,s,<value>
then if the mtime of the file is not newer than the one in the attribute, the rest of the field can be used for the checksum field.  This saves reading the file to calculate the checksum.

so all components that write files would need to write this extended attribute, and all components that post would have to check it.

petersilva:
this involves the use of the python3-xattr module.  (yeah, I know, it doesn't work on Windows... sad but don't know a better answer for now.)  for command line usage, need to install the xattr package.

then can make a test file like so:
```

lacklab% echo "hoho" >"hoho".txt
blacklab%  xattr -w user.sr_headers '{ "lala":"hoho" }' hoho.txt
blacklab%  xattr -l hoho.txt
user.sr_headers: { "lala":"hoho" }
blacklab% 


```

petersilva:
There are limitations to the sizes of all the extended attributes, the key is to be *reasonable*... this cannot be used to store arbitrary amounts of data, it's just for a few tags (the headers of a message)  the json module is used (dumps routine) to create the string that should be easily parsed.

petersilva:
the dms use case also fails miserably with partitioning.  That's OK, their files are small.

petersilva:
To fully address the dms use case, would need to add sum/checksum_a.py  which would be an application checksum algorithm.  This thing never calculates anything because the checksums are application defined. To confirm that this works, you have an on_msg plugin that just overwrites the **sum** header with
a,<randint, or uuid> 

sr_subscribe, when it downloads something it looks at the file on disk, calculates the checksum and compares to what it sees in the message received.  In this case the *calculation* from what is on disk is reading the user.sr_headers extended attribute, and extracting the *sum* header from that, which should be a,<something>.

petersilva:
Just had a meeting with the client confirming there is a strong need for this.

petersilva:
merged and released, but perhaps it isn't finished yet.  

I was thinking one would be able to just supply it on the command line
using either -sum a,hoho ... or -header sum=a,hoho
but neither of those work.  is the file supposed to already
have the extended attribute set when it is posted?

Also needs to be added to documentation.

petersilva:
Just had a meeting with Noureddine and the DMS people.  It turns out that they want to advertise, by making their own AMQP message for us to pick up.  The message will contain sum=a,whatever
[...]

PR #121, opened 2018-11-28
Issue54: Commandline config options for setting checksum xattr

mjip:
Includes documentation for usage/background (I wasn't sure where to stick it, let me know if there's a better place or if there's anything that's missing). Continuation of #54, outline of logic in #119.

#### Posting
`sr_post -header sum=a,65537 <fileName(s)> <configName> start|foreground`

#### Downloading
`sr_subscribe -header sum=a,65537 <fileName(s)> <configName> start|foreground`

### DIFF
--- a/doc/sr_post.1.rst
+++ b/doc/sr_post.1.rst
@@ -322,7 +322,18 @@ common settings, and methods of specifying them.
 
   Add a <name> header with the given value to advertisements. Used to pass strings as metadata.
 
+[-header sum=<flag,sum>]
+~~~~~~~~~~~~~~~~~~~~~~~~
 
+  Checksums can be attached to a file by specifying the sum string value in the header on startup with the 
+  'a' (application) scheme indicated::
+
+      sr_post -header sum=a,65537 <fileName(s)> <configName> start|foreground
+
+  where **fileName(s)** can be a list of space separated files or a value containing regex syntax (path must
+  be specified if not located in the current directory). The **user.sr_sum** and **user.sr_mtime** extended 
+  attributes of the files will be updated before being posted. These attributes can also be set using 
+  commandline utilities like xattr. 
 
 
 ADMINISTRATOR SPECIFIC
--- a/doc/sr_post.1.rst
+++ b/doc/sr_post.1.rst
@@ -4,7 +4,7 @@
 =========
 
 ------------------------------------------------
-Publish the Availability of a File to Subcribers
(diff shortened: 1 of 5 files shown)


mjip:
The documentation has been fixed. I removed the little blurb about how extended file attributes work, is there a good place to put that and explain the optimization a little? Or should it just be left out?

When I was testing it was useful to attach xattr to the files before attempting to download duplicates, but I'm not sure what the use case would be for setting them while downloading. I've removed it from the sr_subscribe rst completely. Since the config option is implemented in sr_config it can be technically used with any sarra instance, however useless.

petersilva:
OK, so I merged what was in the branch so far, but I don't understand some things...

I looked at this:
sr_config.py, 1808:

              if key.lower() == "sum":
                           n = len(words[2:]) + 2
                           for xfile in words[2:]:
                              xattr.setxattr(xfile, 'user.sr_sum', bytes(value,"utf-8"))
                              xmtime = timeflt2str(time.time())
          
it looks to me that this assumes:

    -header sum=a,123124 file.py file2.py ...

but what about:

   -header sum=a,123123 -c plugin.conf -debug file.py file2.py ....

blacklab% sr_post -header sum=a,34254 -c local_post.conf --debug uthash.h
2018-12-05 23:57:03,390 [ERROR] sr_config/option 4 Type: <class 'FileNotFoundError'>, Value: [Errno 2] No such file or directory: b'--debug',  ...
2018-12-05 23:57:03,390 [ERROR] problem evaluating option -header

the logic can't assume the rest of the arguments are files... Can leave it there for now, but eventually, we will get a bug report ;-)


Another thing I don't get.. the optimization by storing sums in extended attributes... I think we likely need an option to say *do that all the time*, because it only seems to be happenning when using 'a' ... whereas it should be able to avoid calculating any checksum, not just *a*.  It looks like it is only designed to work with *a*... I don't see the attribute being set for local files using other checksums. and the download logic needs to store the attribute for use by the next what is described in the next paragraph.

For the accelleration part to work, there needs to be something in sr_message, I think... either compute_local_checksum, or the content_should_not_be_downloaded routine, to refer to the stored
xattre.

I imagin we add a config  option *use_xattr*   True/False one...
and when set to true, the file checksum is *cached* by storing it in the file attribute on each download.

petersilva:
I just want to say that this is fairly arcane & detailed stuff.  It might need some conversations with a white board.  It's hard to explain...

PR #124, opened 2018-12-19
Two on_file plugins for further bulletin processing

mjip:
Work done for Noureddine. The first plugin emails all incoming bulletins and associated egc info to whatever emails are included as config options, the second sends the associated egc command over telnet based on the telnet credentials in the credentials.conf (user/host specified in "telnet://user@host" format in the config options, does a password lookup from there). Adaptations of old fx scripts. Not sure how useful they'd be to anyone but Noureddine/DI but I figured I should document them here.


petersilva:
new plugins can't conflict with anything or break anything.

PR #156, opened 2019-02-20
Issue153

benlapETS:
This enhance log rotation as requested.

+ It add a parameters for minutes.
+ It also documents it in both official language

### DIFF
--- a/doc/Admin.rst
+++ b/doc/Admin.rst
@@ -896,9 +896,9 @@ Replace the contents above with::
 where the +1 can be replaced by the number of days to retain. ( Would have preferred to
 use [0-9]{8}, but it would appear that find's regex syntax does not include repetitions. )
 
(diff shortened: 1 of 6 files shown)


petersilva:
I'm changing the comment: 
( default: midnight 5 ).  Please note that  it
         is impossible to specify 'midnight' in parameter to logrotate. It is only accessible when using the default frequency.

to 
(default: 1d 5 ).

the reference to 'midnight' makes no sense without someone having read the documentation for TimedRotatingFileHandle, which is a bit much to ask.  I don't think any more information is provided even with the understanding.

benlapETS:
I dont understand why you were seeing this old version, but it have put a lot of conflict. Please dont merge we will need to understand what happened before.

benlapETS:
Oh I understand, let me resolve the conflicts.

PR #165, opened 2019-03-01
Fixed the problem with v03 breaking pc_clean f92.

benlapETS:
- The problem was that when the sum was missing from
the json object it raised key error which was busting
the retry list processing and was emptying it.

- Also using a dict comprehension to remove the sum element
that would be safer in a multi threaded system.


benlapETS:
Dont merge it i though it was fine (flow_check succeed) but the retry now seems buggy (flow_limit never ends)

petersilva:
looks good! Great relief this is fixed!

PR #169, opened 2019-03-06
Issue160

benlapETS:
- Added diagrams to our design doc (for message consume|publish process and retry logic)
- Moving code to regenerate chksum each for each time integrity is found seemed to make the system less prone (than master) to get various types of OS errors when running flow test in simultaneous on 2 different machines.

### DIFF
--- a/doc/design/uml/Consumer.puml
+++ b/doc/design/uml/Consumer.puml
@@ -1,0 +1,78 @@
+@startuml
+participant ":sr_subscribe.sr_subscribe" as ss
+participant ":sr_consumer.sr_consumer" as scsc
+participant "consumer:sr_amqp.Consumer" as sac
+participant "msg:sr_message.sr_message" as smsm
+participant "parent:sr_config.sr_config" as pss
+participant "plugin:msg_pclean_f92.Msg_Pclean_F92" as mpf
+participant "plugin:post_log.Post_Log" as pl
+
+''''''Gathering message''''''
+[o-> ss: run()
+loop true
+ activate ss
+ ss -> scsc: consume()
+  activate scsc
+  scsc -> sac: ack(raw_msg)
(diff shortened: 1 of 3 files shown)


benlapETS:
I did new tests and I just wonder why is there so many warning from pclean of lags, file not found, etc ?

Also I got the onfly_checksum warning

petersilva:
for *clean* it might be normal to have file not found errors... for others not so much, but there is a fundamentally pathological case:  file with same name published repeatedly, that causes many issues: we publish the first version, it goes through the entire chain, then clean starts deleting, but a second version is partway through the chain...  it is kind of a weakness now that files with static names that get rewritten.

Perhaps we could add a header (at source) to identify such files  'is-rewritten' or 'multi-version' : 'yes' or something, then the flow test could know what to do... this is not fully thought through.  With Michel, we would say 'not baked'  (from the english expression 'half-baked') .... There is perhaps something we could do there, but it needs a little more thought.

I love the diagrams... you should show me the tooling around them to generate them from these sources. It's cool stuff I don't know about.

haven't tested yet...

benlapETS:
the tool is simple, it is call plantuml it uses dot language and graphiz to generate every possible uml diagrams it is directly available on plantuml.com or you may added as plugin to tools like pycharm or even googledoc to generate your diagrams

petersilva:
So I'm unclear on what to do here... I cherry-picked all the the pure diagram commits to master, so they are there, but there is one commit that mixes the diagrams and the code, and I think, from our discussion, the code is wrong, so I didn't pick that.  Then I cherry-picked the last patch, but it doesn't apply cleanly because of the stuff missing from the previous patch.  so I guess the thing to do is for you to  review how I have mangled the diagram in master, and fix it.

Then we can close the pull request.

PR #170, opened 2019-03-07
Issue166

benlapETS:
Completed the v03 spec + some enhancements and cleanup

### DIFF
--- a/sarra/sr_consumer.py
+++ b/sarra/sr_consumer.py
@@ -138,6 +138,7 @@ class sr_consumer:
         if self.raw_msg != None and not self.raw_msg.isRetry : self.consumer.ack(self.raw_msg)
 
         # consume a new one
+        self.get_message()
         self.raw_msg = self.consumer.consume(self.queue_name)
 
         # if no message from queue, perhaps we have message to retry
@@ -165,7 +166,6 @@ class sr_consumer:
 
         # make use it as a sr_message
         # dont bother with retry... 
-        self.get_message()
         try :
                  self.msg.from_amqplib(self.raw_msg)
                  self.logger.debug("notice %s " % self.msg.notice)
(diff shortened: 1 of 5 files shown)


petersilva:
it is a cleanliness thing... https://pythonhosted.org/watchdog/quickstart.html
I think removing the join has no effect, but I worry about leaving crumbs around in the OS because we don't cleanup after ourselves.

I will merge anyways if it passes the test, but rationale needed.

petersilva:
looking at logs from the flow test, I see incorrect field values:

r_winnow_t01_f10_01.log:2019-03-08 09:10:08,390 [INFO] post_log notice=20190308140907.677624 http://dd5.weather.gc.ca /citypage_weather/xml/NS/s0000264_e.xml headers={'mode': '644', 'source': 'WXO', 'to_clusters': 'ALL', 'sum': 'd,fcc4cd41965c81982673b7041bfffb95', 'parts': '1,30237,1,0,0', 'from_cluster': 'WXO', 'mtime': '20190308140907.0737845898', 'atime': '20190308140907.0737845898', 'size': '0'}

Where does this size header come from?  This is a v02 message.  V02 messages should not have v03 headers in them and vice-versa.  I find 'size=0' through lots of logs... should not be there.

v02 should have a *parts* header.   v03 should have only a *size* header if partitioning strategy is 1, and an additional 'block' (blocking? don't remember) if there is inplace or partitioned  *parts* header in the corresponding v02 message.

petersilva:
not merging for now because of  size=0, and also the unexplained loss of observer.join.

benlapETS:
For the observer.join it was unreachable code so it has never been executed and it will never be as is, I saw it as dead code.
For the v02 I need to investigate, I dont understand why it happens...

benlapETS:
Ok both branch issue160 and issue166 are ready to be merge. Be careful as I force updated issue160, also merging issue166 will raise a merge conflict be sure to keep only changes from issue166...

PR #171, opened 2019-03-07
Issue149

benlapETS:
Enabled enough stack trace to gather good info when debugging also changed logging in unit test


petersilva:
great. clean patch. I imagine it is very useful also.

PR #188, opened 2019-04-18
Issue184

benlapETS:
I completed this task and I believe it improves so that flow test components are now started, stop and cleaned by component instead of with sr. This also improved the flow test in term of readability, modifiability.

### DIFF
--- a/sarra/sr_instances.py
+++ b/sarra/sr_instances.py
@@ -221,10 +221,10 @@ class sr_instances(sr_config):
         for x in os.listdir(dpath):
             fpath = dpath + os.sep + x
             try   : os.unlink(fpath)
-            except: self.logger.debug("could not delete %f on cleanup" % fpath)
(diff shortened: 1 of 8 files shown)


petersilva:
the comm: file 1 is not in sorted order est un bobo
```

Adding flow test configurations... 
Initializing with sr_audit... takes a minute or two
OK, as expected 20 queues existing after 1st audit
OK, as expected 28 exchanges for flow test created
comm: file 1 is not in sorted order
NOTE: Extra exchanges: xflow_public	topic
xhoho_public	topic
xl_tsource	topic
xl_tsub	topic
xr_tsource	topic
xr_tsub	topic
xs_tsource	topic
xs_tsource_clean_f90	topic
xs_tsource_clean_f91	topic
xs_tsource_clean_f92	topic
xs_tsource_output	topic
xs_
```
but I merged it anyways.

Issue #179, opened 2019-04-10
Flow test fails with samples between 3000 and 100000


benlapETS:
I am just opening this issue to document and track the progress that i've done so far with bigger samples. The initial problem is set to 3000 and I set an arbitrary limit to this issue at 100 000.

benlapETS:
There are the results I got with a flow limit sample at 100 000:
```
	Maximum of the shovels is: 100318

		TEST RESULTS

                 | hpfx       routing |
test  1 success: sr_shovel		 (100318) t_dd1 should have the same number of items as t_dd2	 (100316)
test  2 success: sr_winnow		 (200634) should have the same of the number of items of shovels	 (200634)
test  3 success: sr_sarra		 (94223) should have the same number of items as winnows'post	 (100105)
test  4 success: sr_subscribe amqp_f30	 (94222) should have the same number of files as sarra		 (94223)
test  5 success: sr_subscribe mqtt_f31	 (94223) should have the same number of msg as amqp_f30		 (94223)
test  6 success: sr_subscribe mqtt_f32	 (92418) should have the same number of files as sarra		 (94223)
                 | watch      routing |
test  7 success: sr_watch		 (331762) should be 4 times subscribe amqp_f30			 (94222)
test  8 FAILURE: sr_sender		 (157022) should have the same number of items as sr_watch	 (331762)
test  9 success: sr_subscribe u_sftp_f60 (140198) should have the same number of items as sr_sender	 (157022)
test 10 success: sr_subscribe cp_f61	 (140202) should have the same number of items as sr_sender	 (157022)
                 | poll       routing |
test 11 FAILURE: sr_poll test1_f62	 (96500) should have half the same number of items of sr_sender	 (157022)
test 12 success: sr_subscribe q_f71	 (96503) should have the same number of items as sr_poll test1_f62 (96500)
                 | flow_post  routing |
test 13 FAILURE: sr_post test2_f61	 (20928) should have half the same number of items of sr_sender 	 (157022)
test 14 success: sr_subscribe ftp_f70	 (20927) should have the same number of items as sr_post test2_f61 (20928)
test 15 success: sr_post test2_f61	 (20928) should have about the same number of items as shim_f63	 (20927)
                 | py infos   routing |
test 16 FAILURE: sr_shovel pclean_f90 (10511) should have the same number of watched items winnows' post	 (100105)
test 17 FAILURE: sr_shovel pclean_f92 (8517) should have the same number of removed items winnows' post	 (100105)
[...]

PR #221, opened 2019-06-14
Issue182

benlapETS:
- Reverted some previous changes, added error handling and diagrams.
- The flow test is close from success at 10 000 and sr_watch doesnt seem to crash anymore.
- The flow test is accumulating delay during high sample runs, this would need to be adress in #179 .


benlapETS:
Why this pr hasn't been merge yet ?

PR #226, opened 2019-06-18
Issue611

Wahaj10:


petersilva:
Please use a setting:
`
file_email_relay smtp.cmc.ec.gc.ca

`
rather than hard coding in the plugin.

petersilva:
Another comment:  I'm concerned about *Rules for TOPIC*, in file_email.py...  topic has a very specific meaning in AMQP, the string TOPIC doesn't appear anywhere in the rest of the plugin...

* topics are NOT regular expressions, but defined by AMQP protocol. ( https://spring.io/blog/2010/06/14/understanding-amqp-the-protocol-used-by-rabbitmq/ ) 

* reference to regular expression is very unhelpful.  It will actively mislead people.

I don't know what those comments are for, but suggest deleting the stanza.

Wahaj10:
Topic was the wrong choice of word, renamed that to file_type and removed unnecessary comments for simplification. `file_email_relay` variable has been added in. 

Added a fallback to `smtp.cmc.ec.gc.ca` relay server in case the setting isn't configured, is that okay, or should the user be forced to define the relay server to avoid putting that string in the code?

sarra/plugins/file_email.py
```diff
@@ -0,0 +1,85 @@
...
+
+                    with open(ipath) as fp:
+                        emsg = EmailMessage()
+                        emsg.set_content(fp.read())
+
+                    emsg['Subject'] = parent.new_file
+                    emsg['From'] = 'sarra-emailer@canada.ca'
+                    emsg['To'] = recipient
+
+                    try:
```
benlapETS: I am not sure this is require to have a fallback can't we trust what the user enter as a parameter here and if he cannot it would fail.

sarra/plugins/file_email.py
```diff
@@ -0,0 +1,85 @@
...
+                        email_relay = parent.file_email_relay[0]
+                        if not email_relay:
+                            raise AttributeError('sr_sender object has no or incorrect file_email_relay')
+                    except:
```
benlapETS: Bare except is a bad coding practice you should know what you are excepting

sarra/plugins/file_email.py
Wahaj10: Yeah, it's probably better this way. I was just trying to make it more convenient for the user.

sarra/plugins/file_email.py
Wahaj10: Oh okay, that's good to know!

sarra/plugins/file_email.py
```diff
@@ -0,0 +1,85 @@
...
+
+                for recipient in emails:
+                    parent.logger.debug('sending bulletin %s to %s' % (ipath, recipient))
+
+                    with open(ipath) as fp:
+                        emsg = EmailMessage()
+                        emsg.set_content(fp.read())
+
+                    emsg['Subject'] = parent.new_file
+                    emsg['From'] = 'sarra-emailer@canada.ca'
```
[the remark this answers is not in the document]
Wahaj10: Yeah, I didn't imagine there would be a use case but it could be nice to give the user an option to choose (and have this as a fall through)?

petersilva:
I don't think the hard coded canada.ca default for the from address is a good idea.  It should be something like os.getuid() @ self.host   ( sarra@ddsr1.cmc.ec.gc.ca , for example) ... something that has some chance of being correct when used elsewhere.

Wahaj10:
Yeah, that's a good point. I removed the @canada domain because the relay server automatically fills in its domain, and went with just sarracenia-emailer as the default sender.

PR #237, opened 2019-07-30
Bug fix: (commit attempt 2) Always outputs error: relay config not de…

Wahaj10:
…fined which was cased by missing line from earlier commit


petersilva:
oops... I guess wahaj was expecting ben to review... I knew it had no effect on flow_tests, so I just merged it... probably should not have done that...

Wahaj10:
Not a problem, it was a very minor change which actually existed in the previous commit as well.

Issue #63, opened 2018-07-09
subprocess file descriptor issues: sr sanity & sr audit fail or outputs disappearing.


petersilva:
Python recommendations have changed multiple times. There was os.system, subprocess.Popen, subprocess.check_call, and then subprocess.run.  when sr is invoked it causes instances to be started, and then the instances may in turn invoke other processes, such as sr_audit.  We end up with command output disappearing, in some cases.  note this:  https://docs.python.org/3/library/os.html#fd-inheritance which has changed things as well.

when we had subprocess.PIPE in start_instance, sr sanity would abort when invoked with run_command(), so I changed that to None, but worried that on older < 3.4.X the PIPE might be necessary... on the older version does sr sanity still crash?  dunno.  When the change is made,
sr sanity runs, (because I can see it restarting crashed instances) but there is no output.

petersilva:
small demo of one effect of file descriptor issue:
https://stackoverflow.com/questions/51234966/python-subprocess-run-if-you-re-direct-stdout-stderr-it-error-exits-instead-of

petersilva:
sr starts -> sr_audit -> sr sanity   
   sr does a sr_instances.build_instances of sr_audit, which does a sr_config.run_command of sr sanity from the hb_sanity plugin.  so two levels of file descriptors inheritance ? or when is it re-assigned?

The problem might also be just that sr_audit's log file is not properly set.

petersilva:
needs to be tested separately on ubuntu 14, 16, and 18, as the behaviour changes over time.

petersilva:
this patch: 32fec7ead3f2cf05e205841114b3757d76e65a60
makes sr sanity run on ubuuntu 18.04, but I don't know where the stdout is going, there isn't any.
(this is when invoked by the sr_audit plugin.)

likely the realy issue is that stdout & stderr, need to be explicitly handled somehow for sr_audit.

petersilva:
so... now if you invoke sr sanity, the terminal you are using is the stdout for all the components you started... which is wrong, should really re-direct stdout and stderr of components to the respective log files.
patch ab8d7536e3fd3a01d296e1cb0dd6fdfb02904eee does this... but sr sanity fails again when called from sr_audit.  to be continued...

petersilva:
so PEP0446 is the source of the confusion, which was part of https://docs.python.org/3/whatsnew/3.4.html  from the start.  so the boundary is >= 3.4

petersilva:
[...]

PR #242, opened 2019-08-14
Issue240

benlapETS:
Removed condition on python version as the fix provide the same code as the old one, also refactored code to PEP8 standards, and fixed an authenticating issue impairing my flow test.

### DIFF
--- a/sarra/examples/cpump/cno_trouble_f00.inc
+++ b/sarra/examples/cpump/cno_trouble_f00.inc
@@ -10,6 +10,8 @@ reject .*/grib2/.*
 reject .*/model_giops/.*
 # radar imagery
 reject .*/GIF/.*
+# auth based data for client
+reject .*/MSC-SAT/.*
 
 # reject moving targets (file overwritten in a matter of secs)
 # this problem eliminated 2018/08 with introduction of SSL broker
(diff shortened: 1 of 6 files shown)


petersilva:
changing the use of subprocess.Popen means testing a lot of things:

* does it work properly on Windows?

* does it work on 14.04 and 16.04 (Popen behaviour seemed different at each release.)

* does sr_audit properly work?  If you read the old bug report ( #63) the Popen was changed in order to have sr_audit sanity runs work. We saw in ops that sr_audit wasn´t restarting processes as required... you can likely test by having a configuration running (including sr_audit), then kill a process, and wait a few minutes (upto 5) to see if gets restarted.  The symptom before was that if sr sanity was run, it would fail to restart things, which is a big problem in ops.  It would work fine when run interactively, but fail when sanity was called run from sr_audit.    That´s from memory, might be slightly off... Unfortunately, that test was not added to the automated suite.

on the good side, I got 22/23 tests successful 2/4 times... the pcleans fail and that messes up sarra...  sometimes... hmm...

benlapETS:
yeah, maybe the fdelay is too short with 30s for now, but it doesnt make sense to wait more than 30s for a msg to fall through the whole test. I'll test with 60 sec.

- For the compat issues I tested 3.4 to 3.7.

- For Windows, do you have a test case, there is no flow test for it...

benlapETS:
I am about to push a branch with 240-223-227 already merged

Issue #240, opened 2019-08-02
sr_subscribe does not return cleanly when run through ssh/dsh


racetted:
Using version: `2.19.04b2` 

When I try to restart subscribers on several host at once, it hangs at the end of the restart. 

My guess is that the stderr pipe is still being written into while the command ends. 

## Sample code 

```
for host in "some list of hosts" ; do    ssh $host sr_subscribe mirror-ss1-from-hall2-cmop restart;  done
2019-08-01 19:03:22,442 [INFO] sr_subscribe mirror-ss1-from-hall2-cmop 01 stopping (pid=32521)
2019-08-01 19:03:22,946 [INFO] sr_subscribe mirror-ss1-from-hall2-cmop 01 stopped
2019-08-01 19:03:22,947 [INFO] sr_subscribe mirror-ss1-from-hall2-cmop 01 starting
2019-08-01 19:03:22,949 [INFO] sr_subscribe mirror-ss1-from-hall2-cmop 02 stopping (pid=32524)
2019-08-01 19:03:23,452 [INFO] sr_subscribe mirror-ss1-from-hall2-cmop 02 stopped
2019-08-01 19:03:23,452 [INFO] sr_subscribe mirror-ss1-from-hall2-cmop 02 starting
^CKilled by signal 2.
``` 

The `ctrl-c` is done manually as the command simply hangs there.

petersilva:
Hey @benlapETS please have a talk with @racetted  and see if you can reproduce the issue.

benlapETS:
What shell are you using when ssh, sh, bash of ksh ?? Is there more informations about the hosts that fails giving up the stderr ?

racetted:
I was using bash.  
I ran dsh on a `rhel 7.1` powerpc host, and the sshs target some `rhel 7.3` powerpc. 

As for what hosts, I didn't want to publish to the world the internal lists of hosts. If that's a non-issue, let me know.

benlapETS:
Have you tried running sr_subscribe interactively, and if so does it hangs ?

racetted:
Yes, and no it doesn't hang, but will return control before the last stderr output, which I think is the crux of the issue why the ssh will not end properly.

benlapETS:
The problem has something to do with trying to run non-interactive script the same way you are running interactive script and also the fact that ssh redirect stdin which comes to conflict with the way we redirect std (something that was used to accommodate analyst complaining about clogging stdout/stderr in a fancy way, try this:

[...]

PR #257, opened 2019-09-12
Issue240 2

benlapETS:
#240 Retrying a merge, I did a force push since the last pull request, so, opened a new PR.


petersilva:
integration testing says all checks have failed... tried running it my laptop, setup fails:
peter@idefix:~/src/sarracenia/test$ ./flow_setup.sh
Cleaning previous flow test logs...
Adding flow test configurations...
Initializing with sr_audit... takes a minute or two
FAILED, expected 22, but there are 0 queues existing after 1st audit
FAILED, expected 29, but there are 13 exchanges for flow test created
comm: file 1 is not in sorted order
Missing exchanges: xcpublic	topic
... 
gets worse...
never get >0 results.

petersilva:
it works about half the time, when I merge master... I think it is good enough that we can continue
from master... merging.

PR #261, opened 2019-09-15
revamping master branch management & release process

petersilva:
deprecating release.sh script, as we are now experienced enough to use git directly, and be savvier in our tagging & versioning. It also documents the master branch management discussed in last week's development meeting.  The second commit adds French version of the changes, rest of the commits
are just result of review when seeing how github.com renders the documents.

The major change:  release.sh script incremented the version... but the release is actually the end of version development, not the beginning.  so now we explicitly use git to tag, the completion of work on a version, this allows us to build release information incrementally during development.  The new process also makes explicit what was originally intended, but only suggested in the prior procedure, and rarely done in practice: immediate version increment after a release has occurred to clearly separate development work from the previous release.

### DIFF
--- a/doc/Dev.rst
+++ b/doc/Dev.rst
@@ -154,7 +154,7 @@ Checklist:
 - **update doc/** manual pages should get their updates ideally at the same time as the code.
 
 Usually there will be many such cycles on a development branch before one is ready
-to issue a pull request. Eventually, we get to `Commits to the Master Branch`
+to issue a pull request. Eventually, we get to `Commits to the Master Branch <Commits to the Master Branch>`
 
 
 Flow Test Description
--- a/doc/Dev.rst
+++ b/doc/Dev.rst
@@ -1018,18 +1018,18 @@ Releasing
 
 When development for a version is complete. The following should occur:
 
-* A tag should be created to identify the end of the cycle::
+A tag should be created to identify the end of the cycle::
 
-  git checkout master
-  git tag -a sarra-v2.16.01a01 -m "release 2.16.01a01"
-  git push
-  git push origin sarra-v2.16.01a01
+   git checkout master
+   git tag -a sarra-v2.16.01a01 -m "release 2.16.01a01"
+   git push
+   git push origin sarra-v2.16.01a01
 
-* then proceed to update the various distribution methods: `PyPI`, and `Launchpad`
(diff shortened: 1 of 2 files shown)


doc/Dev.rst
```diff
@@ -163,6 +166,8 @@ tested. Please add self-tests as appropriate to this process to reflect the new
 
 A typical development workflow will be::
 
+   git branch issueXXX
+   git checkout issueXXX
```
benlapETS: this could also be git checkout -b issueXXX, which is a shortcut...

doc/Dev.rst
```diff
@@ -172,11 +177,12 @@ A typical development workflow will be::
...
-   ./flow_setup.sh  ; *starts the flows*
-   ./flow_limit.sh  ; *stops the flows after some period (default: 1000) *
-   ./flow_check.sh  ; *checks the flows*
-   ./flow_cleanup.sh  ; *cleans up the flows*
-   
+   ./flow_setup.sh  # *starts the flows*
+   ./flow_limit.sh  # *stops the flows after some period (default: 1000) *
+   ./flow_check.sh  # *checks the flows*
+   ./flow_cleanup.sh  # *cleans up the flows*
+   git commit -a  # on the branch...
```
benlapETS: just a thought on the typical workflow, as one commit branch has never been the typical flow for me: 
- this is time consuming to run the flow test like that for each commit when there are a lot... and it doesn't add plus-value to the system quality. Defining our requirements for the issue and adding test case to our test suite (when needed) would improve much more the quality of our system. The flow test would be the final integration test before the pull request...

doc/fr/Dev.rst
```diff
@@ -153,18 +153,22 @@ Sarracenia mais ne signeront pas les changements ou le paquet source::
     debuild -uc -uc -us
     sudo dpkg -i.../<le paquet qui vient d'être construit>>.
 
-Commetre au Dépôt Principale
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Commetre des changements
```
benlapETS: je dirais plus `Soumettre` plutot que `Commettre` ...

benlapETS: Je vais modifier le texte francais moi-meme...

Issue #185, opened 2019-04-17
Logging structure is poorly designed and poorly implemented


benlapETS:
I suspect that the solution will required a long and thorough analysis, with a lot of examples, use cases and test cases. To really measure the size of the problem, I think I would need help of people using it in day to day to get feedback and idea:
- Peter pointed out that it would be a nice to have to be able to activate/deactivate logging in a more finer granular way so we would remove a lot of noise when trying to debug a problem.
- I encountered issues with the fact the components enable logging in different ways depending on the component and the action that seems to be random and which change during the running of the process. We then dont have a full control of what is popping out and this can be really annoying when a lot of process are running (like in the flow test). Also the log option (-l) doesnt work #186 .

benlapETS:
#240 might be a good reference to this design problem...

petersilva:
hey @benlapETS is this done by branch for #issue240 ( issue240_2 ) or is this something different again.  just looking to understand if we can close it as  part of the next release.

benlapETS:
Let me take a look, I had 3 concerns here:
- (de)activating debug in finer granular way
- log option (-l) RESOLVED
- stdout/stderr redirection problem RESOLVED

petersilva:
closing because I think you're saying it is resolved.

PR #262, opened 2019-09-16
Issue240 2

benlapETS:
Small enhancement related to #240 but also #185


petersilva:
passses fine at home... merging.

PR #264, opened 2019-09-20
Issue258

benlapETS:
Passed flow test on my side with pclean rearranged (limit from 100 to 5000)


benlapETS:
This is an improvement but i think i will need to continue on that next week.

PR #266, opened 2019-09-20
notes from 1st try at release done on github.

petersilva:
work in progress... you can see the result of the first attempt to do a proper release on github.com:

https://github.com/MetPX/sarracenia/releases/tag/v2.19.09b2

There are binaries uploaded at the end.
the release notes are there.

It raises the issue of naming of releases... we have not used release code names before.  Could be fun.

### DIFF
--- a/doc/Dev.rst
+++ b/doc/Dev.rst
@@ -933,7 +933,7 @@ request on github.
 ::
    
    git checkout issueXXX
-   vi CHANGES.rst # summarize the changes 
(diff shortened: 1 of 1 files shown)


benlapETS:
Deploying on github is a very good idea :)... and the flow test looks like testing well, with good results... we are approaching an automated release process...

benlapETS:
Could not from my phone... at least i tried...

PR #271, opened 2019-10-07
Should not have removed this function.  Putting it back.

petersilva:
Revert "Removing from_cluster and to_cluster defaulting."

This reverts commit 46a965430c2bc526abeb84ae2624ac9c1ee394dc.


benlapETS:
Should we put back this line too (sr_poll.post#433) ?
``` 
        #if self.cluster != None : self.msg.headers['from_cluster'] = self.cluster
```

petersilva:
I put it back as recommended, and it passes a flow test for me.

PR #274, opened 2019-10-08
Fixed partflg missing in sr_message.

### LINKED ISSUE #273: partflg missing in sr_message
Partflg is missing in subscriber who wants to process message

benlapETS:
This fix #273 Partflg now default to None as it was before v03 changes

### DIFF
--- a/sarra/sr_message.py
+++ b/sarra/sr_message.py
@@ -123,9 +123,10 @@ class sr_message():
         self.to_clusters = []
         self.tbegin = nowflt()
         self.pubtime = None
+        self.partflg       =  None
 
     def change_partflg(self, partflg ):
-        self.partflg       =  partflg 
+        self.partflg       =  partflg
         self.partstr       = '%s,%d,%d,%d,%d' %\
                              (partflg,self.chunksize,self.block_count,self.remainder,self.current_block)
         self.headers['parts'] = self.partstr


petersilva:
This patch is just new code. no root cause has been identified.
You blame v03 work for the problem,. on what basis?

If you do a git bisection looking for when the initialization of partflg disappeared,
( I just grep for partflg = None in sarra/sr_message.py)
I go back to 2.18.10b2, and the initialization was in the routine set_parts_str(
and only disappeared when you re-organized it in this patch: 5007f9e5593bd6f2474343b20b9d4890de4cf548
on March 6th 2018.  The description of that patch is:

```

    Regrouped set parts routines in set_parts methods
    
    + Declared attributes in constructor (for better readability)
    + Recreate sr_message everytime we're done with the last one (and
    avoid dealing with wrong attributes from previous message)

```

can you try reproducing the problem with the release just prior to this patch,
and after, to determine whether it is the root cause of the issue.

msaraga:
The proposed patch successfully avoids this exception being raised:
```
2019-10-22 17:15:09,990 [ERROR] sr_sarra/run going badly with 'sr_message' object has no attribute 'partflg', sleeping for 0.040.
2019-10-22 17:15:09,991 [DEBUG] Exception details:
  File "/usr/lib/python3/dist-packages/sarra/sr_subscribe.py", line 1757, in run
    ok = self.process_message()
  File "/usr/lib/python3/dist-packages/sarra/sr_subscribe.py", line 895, in process_message
    self.msg.set_new()
  File "/usr/lib/python3/dist-packages/sarra/sr_message.py", line 768, in set_new
    if self.partflg == '1' : return
AttributeError: 'sr_message' object has no attribute 'partflg'
```
However, the following exception is raised (with the current proposed patch, ie setting self.partflg to None in sr_message constructor) suggesting that the above mentioned commit https://github.com/MetPX/sarracenia/commit/5007f9e5593bd6f2474343b20b9d4890de4cf548 broke support for a missing partstr header completely.
```
2019-10-22 17:18:54,489 [ERROR] sr_sarra/run going badly with unsupported operand type(s) for +: 'int' and 'NoneType', sleeping for 0.010.
2019-10-22 17:18:54,489 [DEBUG] Exception details:
  File "/usr/lib/python3/dist-packages/sarra/sr_subscribe.py", line 1755, in run
    ok = self.process_message()
  File "/usr/lib/python3/dist-packages/sarra/sr_subscribe.py", line 938, in process_message
    ok = self.__do_tasks__()
  File "/usr/lib/python3/dist-packages/sarra/sr_subscribe.py", line 421, in __do_tasks__
    if not plugin(self): return False
  File "/usr/lib/python3/dist-packages/sarra/sr_subscribe.py", line 1349, in doit_download
    if not self.overwrite and self.msg.content_should_not_be_downloaded() :
  File "/usr/lib/python3/dist-packages/sarra/sr_message.py", line 152, in content_should_not_be_downloaded
    end   = self.local_offset + self.length
TypeError: unsupported operand type(s) for +: 'int' and 'NoneType'
```

msaraga:
As suggested above https://github.com/MetPX/sarracenia/pull/274#issuecomment-540315529, I just tested commit 35733386ad1ab88ec2eb3f4c443f77fe89ac3d07 from March 3rd 2019 for the above described issues. It was capable of handling messages with no partstr header successfully.

Given these observations, it's probably safe to assume that some of the refactoring done during the implementation of v03 support is responsible for the observed breakage. Reverting commits doesn't seem like an option here unless we want to lose v03... so gonna dive into the message processing logic when no partstr is present and figure it out...

Issue #82, opened 2018-08-13
code is ugly, run it all through yapf


petersilva:
yapf - Yet Another Python formatter.

Often run into bits of code where the formatting isn't PEP8 compliant, lots of indenting that isn't 4 spaces, and the like.  It would be easier to read if it were more consistent.  I ran yapf on the whole code and ran a flow test, and it worked fine, but since it has the effect of changing nearly every line in every file, I didn't feel comfortable doing it while there are active branches.  Will wait until things are a bit quieter.

mjip:
I second this! To enforce it going forward, we could set up a Travis CI build so that every further commit/pull request must be formatted by yapf first or it won't pass the build. Just an idea.

michelgrenier:
I agree (and faulty :-) )

benlapETS:
Here is the result of a static analysis done with a pycharm Inspection tool:

![image](https://user-images.githubusercontent.com/11091581/66416211-4febd800-e9cb-11e9-9705-f720dd394a8f.png)

A lot of those problems could be resolve automatically... (read very easily without any impact). Not all of them though, but they should be address to prevent issues like #274 which was related to "Instance attribute defined outside __init__" . Note that we have 888 other warnings like this specific one, so the risk to have other issues like #274 with any change is very high.

benlapETS:
For the yapf thing I think we don't need that if every dev is committed to use an IDE that does those validation automatically as you code, which is the standard to develop in python (or any other high level modern languages.

michelgrenier:
Benoit… just stop complaining and fix them
You dont need to copy me to do so.

Make it pep8 and go through errors first, than the warnings
You can clean the code in a few days 

Good luck

benlapETS:
No comments...

petersilva:
dev environments are notoriously contentious.... obligatory... https://xkcd.com/378/
For one guy using a DE might be great, for another, running a tool might be more comfortable.  We should think hard before telling people they should be uncomfortable.  I think Michel´s point was just that tools create a lot of noise when the code is too messy.  it makes sense to do the broad strokes clean up to get to a reasonable start state... so if we see 800 issues, and we know there are straight-forward fixes for most of it, we just do it and get it out of the way.  we´ll have 200 left in short order.
but I understand also the approach of publishing a baseline.  it´s more rewarding to see the effect of the cleanup.

petersilva:
to run yapf from within vim:

`
:%!yapf
`
[...]

PR #279, opened 2019-10-24
ran yapf on the whole tree, as a start...

petersilva:
so this is not necessarily a pull that I want merged... it's just the easiest way to ask for a review...

so it's after a release, so there are no parallel branches.  This might be a good time to reformat the tree.
take a look try out the analysis tools, and see if it improves the situation with #82


petersilva:
too late now... upstream has changed...

Issue #280, opened 2019-11-01
sr_poll does not tolerate faulty ssl certificates on listing.


kurt2x:
We encountered a problem where we couldn't sr_poll a backup site because it had an invalid ssl certificate on it's https service.

It would be great if some configurable option could be added that would allow us to download from such sites.

petersilva:
tested with www7320.nrlssc.navy.mil
didn't bomb.

petersilva:
We should probably leave this open as a place holder... the original problem is fixed by ignoring bad certs everywhere.  We should likely take the opportunity to expose settings to tune SSL toleration for bad certs.

so will leave open even if the original issue is solved.   the corrent patch is a work-around. real solution should be tunable.

petersilva:
OK added *tls_rigour* option with values:  *lax, medium, strict*  so for this use case, one would
select:

`tls_rigour lax
`
in the config file, to get it to tolerate bad certs, but the default remains to reject it.

PR #281, opened 2019-11-02
Issue280

petersilva:
work-around for bug of sr_poll connecting to SSL http sites with bad certs.  This current work-around tolerates that.  longer term solution should be to have programmable levels of security, adaptable for different situations.


petersilva:
could candidate for a squash merge ...

msaraga:
#280

Issue #245, opened 2019-08-16
sleeping beauties 


michelgrenier:
The migration of sundew pulls to  sr-poll/sr-sarra processes has spread around lots or processes on each node of the sarra cluster. Most of theses, run their task on a single sarra node ... according to the given vip the poll was set to work on, posting locally ... sr_sarra listening locally  fetching locally and reannouncing locally .... etc.   Only a few poll would spread out their messages onto DDSR and their sr_sarra counter part would share the same queue on all nodes hence spreading the fetching through out the sarra cluster.   

For all the sr-poll/sr-sarra that does their thing locally,  a cluster of 8 nodes ends up with  7 pairs of sr-poll/sr-sarra that basically are sleeping beauties.  It would be nice to specify the vip for both processes (now it is only for the poll that it is specified)  and for all the sleeping one  make the code wise enough to strictly uses the resources it needs in this vip sleeping state... Ex.: no amqp connections while we are asleep,  

The iteration would   check for vip,  do the work needed if sleeping (for poll that the matching sr_sarra do not delete... update the ls* files)  ...and go back to sleep...  It the vip is found,  the sr_amqp stuff is set not being asleep (consumer and publisher) and consuming  reactivates the connection and returns a message Iif available) and the publishing  reactivate its connection and publishes...   There was some seeds of this idea earlier in the code where sr_amqp had and was testing against  self.sleeping... but seeds was not enough not to look like dead code and it was removed.  

Because there are a non-negligable amount of sleeping beauties and I have read lately some worries about node resources... I thought it was worth sharing and underlined as a possible feature to implement

petersilva:
+1 for the name :-)

petersilva:
might be worth providing some numbers... on the ddsr's (our main broker cluster), we are running something like 100 polls? so there are 800 polling processes connecting to the broker.   could save 700 processes and just as many connections to the broker.

petersilva:
There is something fundamentally odd about sr_poll.   Everyone expects that when has_vip is false, the process does nothing, but that isn't how it works.   sr_poll was originallly applied to cases where we needed to run on both nodes with and without vip to keep the remote listing (state) uptodate, so on_poll plugin runs whether you have the vip or not.  This is very unexpected.   it would be nice to figure out an elegant way to have it do the right thing for both use cases. Everyone expects that on_poll will only be run if  has_vip is true.  Right now, every on_poll needs to check has_vip again...
[...]

PR #282, opened 2019-11-10
Issue245 - reduce overhead of poll when they don´t have the vip.

petersilva:
We are seeing high load on ddsr nodes since moving the polls from pxatx there.  I think a big part of that is that the polls do not work the way people expect.  what the vip does for sr_poll is stop the nodes from posting, but they still poll. so you have N nodes in a cluster, they all poll, but only one posts.  In many cases, we only need one node to poll (and post) and the state doesn´t matter.  the option added here allows that case.


msaraga:
#245

PR #297, opened 2020-01-13
Fix confusing asterisks in documentation

n-a-t-e:
I found a few cases where there are extra asterisks in the French and English docs which get rendered instead of making the text bold. Makes some config options a bit easier to understand for new users.


petersilva:
Thanks for taking the time!  We have trouble finding reviewers for the documentation, especially in French, the quality is quite uneven.  All improvements welcome.

PR #337, opened 2020-05-27
Fix indent in moth_api_consumer.py

michaeldavie:


petersilva:
fari play indeed. Thanks!

PR #338, opened 2020-06-09
Update subscriber.rst A First Example

danielpronych:
Modifications to additional notes / changes to A First Example additions to subscriber.rst. Clarifies where these changes need to be made, one on command line, either in the source configuration.

### DIFF
--- a/doc/subscriber.rst
+++ b/doc/subscriber.rst
@@ -187,7 +187,7 @@ NOTE:
   
   $ mkdir /tmp/swob_downloads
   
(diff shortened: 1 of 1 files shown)


petersilva:
what is *alue* ?

danielpronych:
Hi, Peter, intended to read "option,value" pair , not alue. Does this clarify? Thanks.

danielpronych:
HI, Peter, since it seems we have determined option value pair is better wording, this has been modified to this wording. Is this better? Thanks.

Issue #356, opened 2020-09-10
Review plugin API through the accel plugin case


benlapETS:
We now have a new plugin API which has been described in #290 and which solved the old unfriendly way from v2 where the import was happening in a different place than the invocation (module context != runtime context), so plugins developper had to import everything he needed inside the routines he was creating. With what has been done in v03 it is much cleaner now and it works well with most of the plugins, which are elegantly callable from flow. Unfortunatly, the approach seems to be uncooperative with plugins involving do_download, do_get, do_put, do_send. Because of a protocol layer of classes between flow and the plugin.

accel_wget plugin is a good example and which is now unusable . Anytime we reactivate it for config/sarra/download/* components,the static flow fails. A good way to reproduce is to run the static test normally and then stop the watch and poll, and just play with foreground of the sarra component. It will crash every time the plugin is invoked.

The problem has something to do with the new API mixed in with do_get call, in the plugin/ directory. What fails first is plugin/accel_wget.do_get . It has "no connect" member in the class, trying to inherit it from sarra.transfer.https.HTTP , but then the constructor fails because the plugin constructor diverges from transfer constructor that take different arguments.

Solutions tried:

1. Splitting into 2 classes in the same file ... the plugin/_init/loadlibrary then complains heavily
2. Building 2 separate plugins, but the problem stayed the same:
    - plugin/accel with the on_message part
    - plugin/accel_wget with the http/download part and this one inherit from transfer.https.HTTP

Solution Proposal:

1. Creating a plugin that inherits from transfer/(sftp,https) and just override the one routine we need. But the question is:
    - How to do an api where you can inherit most from the existing transfer class and just override what you need ?

Complication:

1. Plugins are actually called from flow/_init_ , then, not a parent of transfer

benlapETS:
I just tried the static test with wget activated and sarra failed with that trace:

```
Traceback (most recent call last):
  File "/home/benoit/git/sarracenia/sarra/instance.py", line 207, in <module>
    i.start()
  File "/home/benoit/git/sarracenia/sarra/instance.py", line 199, in start
    self.running_instance.run()
  File "/home/benoit/git/sarracenia/sarra/flow/__init__.py", line 310, in run
    self.do()
  File "/home/benoit/git/sarracenia/sarra/flow/sarra.py", line 31, in do
    self.do_download()
  File "/home/benoit/git/sarracenia/sarra/flow/__init__.py", line 885, in do_download
    ok = self.download(msg, self.o)
[...]

PR #362, opened 2020-09-25
Issue356 1

benlapETS:
Complete the request from #356

### DIFF
--- a/sarra/config.py
+++ b/sarra/config.py
@@ -128,7 +128,6 @@ def declare_plugin_option(option, kind):
        'str'        an arbitrary string value, as will all of the above types, each succeeding occurrence overrides the previous one.
 
     """
-
     if kind == 'count':
         count_options.append(option)
     elif kind == 'duration':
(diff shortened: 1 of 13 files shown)


petersilva:
less magic... but still a little more than ideal...

* I see create still using kwargs / proto mechanism. as said in the issue. I provided examples which do not need __init_subclass__ or kwargs ... integrity and flow classes with factory methods that just iterate using the names of the sub-classes.  I understand the fragility of that, and that something like protos is better, but you could declare a protos or registry property and use that in place of __name__ in the factory and it should work.  The factory method (create?) is outside the class... in the examples, factory was declared as a \@staticmethod within the class. perhaps outside is better (shorter name, and avoid a decorator, so less concepts used.)., any other pluses/minuses of the two options?

* I think you need to pull again... I moved the integrity class out of plugin (it was always strange being in that hierarchy, since it was not a child of sarra.plugin) but when I checked out your branch my integrity subdir is empty... not a huge deal, but would be nice if it were upto date.

* I saw a lot of changes in accel_* and so I thought maybe you had done a lot more than expected in the first patch...When I tried turning on the accellerator in the static_flow, it died... 

```
  File "<frozen importlib._bootstrap_external>", line 783, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/home/peter/Sarracenia/v03_wip/sarra/plugin/accel_wget.py", line 46, in <module>
    class ACCEL_WGET(Plugin, sarra.transfer.sftp.Sftp):
  File "/home/peter/Sarracenia/v03_wip/sarra/transfer/__init__.py", line 122, in __init_subclass__
    schemes = kwargs.pop('schemes')
KeyError: 'schemes'
fractal% 

```

OK... I'm guessing that is what was expected with the first patch.  is that what you expect?

benlapETS:
ok for the review:

- Removed kwargs in __init_subclass_ will try that in the next patch, but it would go against the factory pattern I presented as the solution.
- Factory method was made outside to be consequent with other factory in the code (ie plugin factory)
- For the integrity I couldn't integrate code that you hadnt pushed yet it will be there in the next patch
- As requested accel_* are now useless (and I have not touched it if you would have look at the sum of the diff -- files changed tab at the top of the pull request) as the system is now using Transfer class children (wget, scp) for transfering, then if you want to activate it you don't use the plugin keyword but the import keyword.

petersilva:
after discussion, went a different direction.

Issue #343, opened 2020-06-18
IRIDIUM MSI upload plugin 


petersilva:
debate going on whether to do this with SFTP or SMTP...  Our preference is SFTP, so start with that.

This is a continuation of other Satellite MSI (Marine Safety Information) transmission to ships navigating in the arctic. An extension to existing INMARSAT service.   IRIDIUM should work all the way to the North Pole. This INMARSAT work was done as part of the METAREAS project.

petersilva:
@tysonkaufmann delivered an initial version last week. @EricFunk implemented it on dev. percolating through to ops.
will percolate to public side  post deployment.

petersilva:
@tysonkaufmann can you confirm this is all merged? was cleaning up old branches.
Leaving this one alone for now.

petersilva:
OK.  will delete the branch... can always create a new one should the need arise later.
just tidying up.

PR #367, opened 2021-03-10
Add program_name check for file_index creation

tysonkaufmann:
Adding statement to prevent sr_senders from calling init_file_index.

sr_senders do not use the file_index, and calling it requires write permissions to be granted on the receiver else an error is thrown.

Fixes sends to hosts which do not have write permissions enabled on other files/directories.

Related to: #343


petersilva:
Should say... this issue stemmed from trying to send files to the IRIDIUM safetynet system, which has an odd sftp server.  The interaction is such that requesting the "ls" command on initial connection causes a strange error.   This "ls" behaviour is not desirable or useful in the context of a sender, so putting a guard in front of it, so that it is only done when appropriate (in a poll.)

PR #427, opened 2021-11-03
add Docker capability

tomkralidis:
Adds Docker setup and sample Docker Compose workflow.


petersilva:
um... would prefer "pump" rather than "watch" with the entire config directory mapped and the ability to run a number of configurations in the same container... is it better to acccept the pull and change it after? or should it be changed first?
It still references an individual watch config file instead of the whole directory, and it doesn't persist the logs or statefiles.

petersilva:
thinking more about that... do we want the config files to be outside the container as well?

tomkralidis:
Thanks; I've pushed some changes to address some feedback.  Where are the statefiles usually placed in an sr3 install?

Having said this, the `docker/compose` setup is there as an example.

petersilva:
~/.cache/sr3

Issue #395, opened 2021-08-24
why is memory check missing? perhaps rewrite all v2 hb_* as v3 hk_*


petersilva:
Periodic processing is a feature in Sarracenia, necessary for various maintenance tasks. in v2 it was referred to as heartbeat, which made it sound like it should happen way too often, and as there was a completely separate feature called *pulse* (which has a different interval) ... anyways... in v3 the name is *housekeeping* which is much closer to a good description... it happens often but not too often.

There are some built-in plugins that were in v2, that don't seem to run in v3.  In particular, 
in v2, there is hb_memory... in v3, either old hb_memory should work (upward compad) or there should be a new hk_memory.  This is important because paramiko has a memory leak and if we don't restart periodically, the processes grow without bound and eventually server crashes.

perhaps rewrite all of them (not that many exist.)

petersilva:
look in flow/__init__.py  Flow class: __init__.py  for append( to 'load' thing...

MagikEh:
Run a dynamic flow test, smtp server or something with Parameko (known memory leaks)

MagikEh:
Confirmed that memory is reclaimed via the currently supplied method! I saw a `subscribe/u_sftp_f60` instance memory usage go down after a restart!!

(Had to set `hb_memory_multiplier` to 1.002 to trigger a few times)
`tail -f ~/.cache/sr3/log/*.log | grep Memory | grep -v cpu_times | grep -E " [14-9][\.0-9]+%|restart|$"`

![image](https://user-images.githubusercontent.com/6865027/151257194-ca09a31a-df42-4283-a7b9-732a005a5a15.png)

petersilva:
great!

petersilva:
sounds ready for a pull request.

petersilva:
merged. great!

PR #448, opened 2022-01-31
HeartBeat Plugins into V03 wip

MagikEh:
Changes to heartbeat plugins now referred to as housekeeping plugins. re: #395 

## TODO:
- [ ] Review changes
- [ ] Confirm flow tests pass

## Memory
- [x] re-write to enable functionality on windows and linux
- [x] Adding resources.py to `self.plugins['load']` to enable it during runtime (fixes bug where it just wasn't running) 
- [x] Renamed config options and added to synonyms list
- [x] Renamed `memory.py` -> `resources.py` to better reflect function of the plugin (it also reports CPU times)

## Config type `float`
- [x] Found that the config didn't have a float (decimal) type other than a string, added for more terse configs

## hb_Message
- [x] Printed out lag messages, functionality already in `log.py`

## setup.cfg 
- [x] Enable pep8 linter config option: `max-line-length = 119` as per [Peter's Development Guideline](docs/Contribution/Development.rst#tools-you-need).


MagikEh:
@cleaaum can you comment on the early commits on this branch? Should they be allowed to get pulled into the `v03_wip`?

cleaaum:
The first one accidentally has file_age ported which is not right - it accidentally got put in there. But the other 2 are fine - they are just 2 of those generic plugins

petersilva:
Once I pulled in v03_wip...  file_age went away... it was just that the branch was not upto date with v03_wip.  however, the static flow does not pass... so confused... cannot merge yet.

petersilva:
when you do a git merge.... could you try:

git checkout v03_wip
git pull

I have a feeling the merge merged what was in the v03_wip branch last time a pull happenned...

the 30->300 thing... yeah 30 was an error... don´t know where it came from 300 was the documented default, was corrected.  We can set it to 30 if we want to in default.conf for use in the tests... hmm...

MagikEh:
Looking at [the network graph](https://github.com/MetPX/sarracenia/network) here on GitHub, the hb_branch is the most up to date : 
image

(We can see the `v03_wip` is 4 commits behind `v03_wip_plugins_v2Tov3_hb` and that the most recent commit to the `v03_wip_plugins_v2Tov3_hb` branch is merging in the `v03_wip` branch's changes)

Issue #416, opened 2021-09-19
rabbitmq SASL authentication.


petersilva:
German weather service has deployed their rabbitmq broker with SASL as the authentication mechanism.  We can't connect to it.

reidsunderland:
If a broker supports more than one SASL authentication mechanism, the AMQP library might not choose the correct one. We can override the automatic selection by passing a `login_method` string to the `amqp.Connection` call.

https://github.com/MetPX/sarracenia/blob/85e60d22da9c39cae4f83c0c02ae203fe3175b2f/sarracenia/moth/amqp.py#L187-L191

Reference: https://docs.celeryproject.org/projects/amqp/en/latest/reference/amqp.connection.html

reidsunderland:
We would like to add a `login_method=SOMETHING` parameter to the credentials.conf file, but need to figure out the best way for `moth/amqp.py` to access the credentials.

For example:
```
amqps://username:password@my.broker.ca login_method=PLAIN
```

petersilva:
your branch passed static test for me.  for dynamic, it failed one test, but looking at the log, it has nothing to do with your code. so accepting into v03_main, and producing a separate patch for the other problem.

reidsunderland:
Fixed this issue by adding the ability to define a `login_method` in the credentials.conf file. See documentation [here](https://github.com/MetPX/sarracenia/blob/4601c060cedb56f411540c974a15cb4bc92c01f0/docs/Reference/sr3.1.rst#credential-details).

To connect to the German Weather Service's broker, the `login_method` should be set to `PLAIN` or `AMQPLAIN`.

This required changing the type of object that was passed to the moth classes (and other parts of the code) from `urllib.parse.ParseResult` to `credential_details`, which contains the ParseResult, and other information.

petersilva:
I have succeeded in accessing the DWD (German Weather forecasting service) server with sr3.  Done and Dusted!

petersilva:
uh... buddy... pal... I just noticed we did not update the examples... um.. can you take a look at them and suggest how to do the credential parsing, so they work again?

petersilva:
I'm talking about sr3/sarracenia/examples/*.py
there are a few files there. and then there are the similar jupyter notebooks to update as well.

The problem to consider is that in moth ... they aren't using a config class... so... not sure what exactly...
someone would have to think about it.

PR #449, opened 2022-02-04
#416 Part 2: Fix Examples & Rename Classes

reidsunderland:
This fixes the examples `./sarracenia/examples/moth*.py` and Jupyter notebooks that were broken when #416 was completed.

It also renames the classes inside `credentials.py`:

- `credential_details` --> `Credential`
- `Credentials` --> `CredentialDB`

The static flow tests are passing locally and in GitHub Actions. 😃


petersilva (approved):
great!

PR #459, opened 2022-03-02
V03 wip issue348 2

CodyAu:
Fixed the statehost issue by parsing the config before lr

All static flow tests pass, and flakey + dynamic flow tests pass as well as can be expected


petersilva:
great!

PR #487, opened 2022-03-24
Continue on error for artifacts

reidsunderland:
If the artifacts fail to upload, the whole job shouldn't fail.


petersilva (approved):
Cool!

PR #496, opened 2022-03-31
Artifacts Issue

reidsunderland:
Will compressing the artifacts into a tar.gz file before the upload-artifact step help reduce the errors we're getting?


reidsunderland:
Creating a tar.gz file seems to eliminate or reduce the artifact upload errors that we were getting. Instead of uploading ~30 files, it just has to upload one.

PR #529, opened 2022-06-13
Update flow.yml

reidsunderland:
Don't merge this :) PR will trigger the Actions workflows to run

edit to trigger again

### DIFF
--- a/.github/workflows/flow.yml
+++ b/.github/workflows/flow.yml
@@ -45,10 +45,20 @@ jobs:
         run: |
           cd ${HOME}; pwd; ls ; 
           echo hoho
+          mkdir -p ${HOME}/.config/sr3/cpost
+          mkdir -p ${HOME}/.config/sr3/cpump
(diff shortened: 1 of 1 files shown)


petersilva:
I'm confused... it's a PR that says don't merge this.  It looks like it does fix the booboo... but... I guess people want to fix it some other way?

reidsunderland:
Yes, sorry for the confusion! Creating a PR from a branch causes the GitHub Actions workflow to run for that branch. It let me troubleshoot without committing directly to v03_wip.

It does fix the issue, by creating all the sr3 config subfolders before running setup:
https://github.com/MetPX/sarracenia/blob/0e986a3db7427f10e32306aeb83ed7c3c00c4e30/.github/workflows/flow.yml#L48-L59

That would probably be better to do in sr_insects, or maybe the converter code should check and create the folder if it doesn't exist?

petersilva:
Perhaps even better if the converter ensured the directory exists. before trying to write the file.

PR #530, opened 2022-06-14
V03 wip doc

cleaaum:
Converter doc


petersilva:
In the command line guide... further down... there is an example of running each command.  Can you add a section with a sample run for _convert_ ?

Issue #537, opened 2022-06-27
Message Strategy Stubborn


reidsunderland:
Get message strategy stubborn needs to be properly implemented/fixed. This is used in the Moth classes.

When `message_strategy['stubborn']` is `True`, methods such as `putNewMessage` should block and infinitely try to reconnect to the broker when a connection interruption occurs. This is the current behavior for `getNewMessage`, `__putSetup`, `__getSetup` in `amqp.py`.

When `False`, those methods should not block and instead return control back to their calling methods and somehow report the failure. Other retry logic should handle retrying whatever failed.  

- For example, when `putNewMessage` fails, it should return `False`, and the message should be saved to disk using the post retry DiskQueue and then eventually re-posted at another time.

reidsunderland:
I think this requires at least:

- a way to set `message_strategy['stubborn']` to True or False in the config files.
- implementing changes to `moth/amqp.py getNewMessage` so it checks whether the connection to the broker is connected at the beginning of the method, similar to [this for `putNewMessage`](https://github.com/MetPX/sarracenia/blob/4c9a4c8455a7678fd4ae11149b21438708661013/sarracenia/moth/amqp.py#L448-L456).
- testing and fixing `__putSetup` and `__getSetup` so they function correctly without infinite loops.

Issue #466, opened 2022-03-15
If sender does not succeed in posting a message after sending a file, it should retry.


petersilva:
@habilinour suggested this...
when we are sending to a server,  we complete sending the file, but then need to publish the notification and that part fails... we should (optionally?) retry.

petersilva:
I just looked at the logic in sr3, and it's already implemented there afaict.  if the moth.putMessage( fails, then it gets put on the failed list, and the flowcb/retry module, perrsists that.  so across shutdown & restart, if there are pending posts, it should take care of it... I guess would have to test.

petersilva:
well the retry logic is there, but the logic in the MQP modules to return a failure isn't. I think for now it just locks up trying to post the message looping on re-connection forever. so there is still some work to do.

petersilva:
@reidsunderland  does this make sense to you? Can you run the flow tests? steps:
* not loop trying to re-connect in publish... putMessage needs to accept the connection is broken, and return False.
* the flowcb/retry.py module should take the resulting messages that failed to be published, and persist them to a retry file.
* on restart, when returning to publish, flowcb/retry already seems to add selections from the retry file to the publishing queue.
* It should perhaps simply attempt to connect once if not connected, and give up if it fails, rather than looping as it does today.

I think the key here is to start getting publish to return False, have the

reidsunderland:
`flowcb/retry.py` uses a callback, `after_post` that wasn't implemented. c468cbdc4b7dfbdf1f9089b8096aa63e356ac8a2 enables `after_post`.

> * not loop trying to re-connect in publish... putMessage needs to accept the connection is broken, and return False.

✔️ Forcing a return of False from `putNewMessage` does cause the message to be put in the list of failed messages.

> * the flowcb/retry.py module should take the resulting messages that failed to be published, and persist them to a retry file.

✔️ This works.

```
[...]

PR #538, opened 2022-06-27
For #466: Changes to get post retry DiskQueues working.

reidsunderland:
See details in #466.

- Adds `metrics_report` callback, to allow flow to receive information from plugins.
- Adds `retryEmptyBeforeExit` option, to prevent Sarracenia (particularly post) from exiting while there are still messages in the DiskQueue to be retried. Used in the flow tests, https://github.com/MetPX/sr_insects/pull/11 has already been merged.
- Adds message counting to the DiskQueue class.
- Removes `isRetry` from messages, because it caused a bug and wasn't used.

The use of the DiskQueues for retrying messages that failed to post is enabled by this: https://github.com/MetPX/sarracenia/blob/e1d5de42cf6d05ccd4a098c0f92d2fe988689f39/sarracenia/moth/amqp.py#L546-L551

But this is a temporary fix, and should be configurable by setting `message_strategy['stubborn']` to `False` when #537 is completed.

### DIFF
--- a/docs/source/Reference/sr3_options.7.rst
+++ b/docs/source/Reference/sr3_options.7.rst
@@ -1354,6 +1354,13 @@ the reception cache is also discarded.
 The AMQP protocol defines other queue options which are not exposed
 via sarracenia, because sarracenia itself picks appropriate values.
 
+
+retryEmptyBeforeExit: <boolean> (default: False)
+------------------------------------------------
+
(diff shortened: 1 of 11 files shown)


petersilva:
yay!

petersilva (commented):
This will be an excellent improvement. great!

PR #570, opened 2022-10-13
Changes to sr3 start/stop behaviour to fix #555

### LINKED ISSUE #555: fix restart_server test 6 failure (comparison of downloaded_by_sub_amqp and cfile directories.
With the new flakey_server (now called restart_server) test created by https://github.com/MetPX/sr_insects/issues/14, which stops and starts sr3 itself during the test.  I think this might be a problem with shutdown code causing data loss.  More investigation required.

reidsunderland:
https://github.com/MetPX/sarracenia/issues/555. The restart server test was failing because `sr3 stop` would stop foreground posts (polls started using sr3_post and sr3_cpost.

This changes sr3 so that `sr3 stop` (without specific configs specified) never stops any foreground instances, and `sr3 start` also does not start posts that cannot run as daemons.

Foreground possible cases:
* anything with ``foreground`` in its cmd_line is always foreground.
* anything with ``start`` in its cmd_line is always a daemon (not foreground).
* posts and cposts without ``start`` in the cmd_line are always foreground.
* other components without ``start`` or ``foreground`` in the cmd_line are not foreground.

Note: `sr3 stop component/config` will still stop the specified foreground instance.

Posts/cposts can only run as a daemon when they have a `sleep` value > 0 in the config file, and a path defined in the config file. If these criteria are not met, `sr3 start` will not attempt to start them.

**https://github.com/MetPX/sr_insects/pull/20 should be merged first**

### DIFF
--- a/sarracenia/sr.py
+++ b/sarracenia/sr.py
@@ -1008,6 +1008,8 @@ class sr_GlobalState:
         self._resolve()
         self._find_missing_instances()
         #print('analysis - Done. ', flush=True)
+        # True if the user did ``sr3 action`` with no configs/components specified
+        self._action_all_configs = (config_fnmatches is None or len(config_fnmatches) == 0)
         self._match_patterns(config_fnmatches)
         os.chdir(self.invoking_directory)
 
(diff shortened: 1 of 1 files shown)


petersilva (approved):
I ran the tests locally and everything looks good.
Also read through the patches, and they seem sensible.
Great stuff!!

PR #578, opened 2022-10-21
Stop foreground posts to fix artifact collection

reidsunderland:
I noticed a few actions runs failed to save artifacts because the log files changed. Since sr3 stop doesn't stop posts anymore, I added an explicit `sr3 stop post/* cpost/*`

### DIFF
--- a/.github/workflows/flow.yml
+++ b/.github/workflows/flow.yml
@@ -69,6 +69,7 @@ jobs:
         continue-on-error: true
         run: |
           sr3 stop
+          sr3 stop post/* cpost/*
           cd ${HOME}/.cache/sr3/
           tar -czf ${HOME}/cache_sr3.tar.gz *
       


petersilva:
um... why does that work? I will accept the patch anyways... but I'm confused... I thought it would ignore post and cpost if running in the foreground.  I don't get why the general command is any different from the specific one, but cool I guess.

reidsunderland:
In my change to sr3 stop I added a check if the user called just `sr3 stop` or if they specified any configs. https://github.com/MetPX/sarracenia/blob/v03_wip/sarracenia/sr.py#L1647

I was thinking it would be weird if a user called `sr3 stop cpost/thing` and nothing happened. But if this behavior is more weird/unintuitive then I could change it and maybe just warn the user that the foreground config won't be stopped?

petersilva:
as per discussion:

 * should likely print a warning that ignoring foreground tasks if there are some present.
 
 * should likely kill off foreground stuff if given --dangerWillRobinson option.

Issue #379, opened 2021-07-22
AM protocol driver


petersilva:
To help with decommissioning Sundew, but not lose interop with MM, then we could make a sender/receiver for AM sockets...

petersilva:
@habilinour suggestes we could do an on_file plugin to write to AM.
but requires Sarra on all the client systems, versus still having to run MM on the data hub, and feeding it that way, but same on_file approach would work in both cases.

petersilva:
unable to get python-paramiko on ubuntu 20.04 
  https://github.com/MetPX/Sundew/issues/16

petersilva:
@kurt2x implemented data hub with combo of Sundew and Sarra... he thinks having an AM Sender of some kind would be good.

petersilva:
@kurt2x likely wants the receiver to act like an amtcpserver connection,  where you have  a master that forks individual connections using a single port defined for all connections (different from Sundew, where 1 port was assigned to each configuration.)

(this might be too much for an initial attempt.)... perhaps just start with a script that accepts a connection and processes bulletins received  (separate one for a sender.)

andreleblanc11:
The AM sender is completed and ready to test. One problem though, Sundew is giving me problems when setting up the test receiver on my Azure VM which has Ubuntu 20.04 installed. There seems to be quite a lot of missing dependencies on `Python2`, Especially with `paramiko`. This has already been addressed in [another issue](https://github.com/MetPX/Sundew/issues/16).

petersilva:
I spun up a vm to try things out... the following might work:
```

sudo apt update
sudo apt upgrade
sudo apt install python
curl https://bootstrap.pypa.io/pip/2.7/get-pip.py --output get-pip.py
sudo python get-pip.py
sudo pip install python-paramiko

```

petersilva:
reference: https://linuxize.com/post/how-to-install-pip-on-ubuntu-20.04/

petersilva:
General rule, do apt search python-x to find python library x.  If there isn't one, then install with sudo pip install x.
ubuntu 20, is the last version where even this recipe will work.

petersilva:
OK so I sat with @andreleblanc11  yesterday and we pair programmed for a bit.  He now has a flowcb plugin that works for accepting messages from a single connection, writing them to a file in a configured directory, and posting a message for the resultant local file.  We had to comment out the poll initialization in order to get it to work for our example, because:

[...]

PR #588, opened 2022-10-28
V03 wip issue575 ... adding "flow" component, and flowMain option.

petersilva:
so this doesn't do the whole job of the issue, but I think it is useful on it's own, and it lets @andreleblanc11 's stuff work (he's using it in his branch also.)

it adds:
   -- the ability to declare flow/  component configurations that are vanilla.
   -- the ability to declare flowMain to run arbitrary subclasses of flow making it possible to define all existing components as config files.
   -- include files to implement all the existing components.
   -- English and French documentation of the above.

### DIFF
--- a/docs/source/Explanation/CommandLineGuide.rst
+++ b/docs/source/Explanation/CommandLineGuide.rst
@@ -31,6 +31,7 @@ The components of sarracenia are groups of defaults on the main algorithm,
 to reduce the size of individual components.  The components are:
 
  - cpump - copy notification messages from one pump another second one (a C implementation of shovel.)
+ - flow - generic flow with no default behaviours. Good basis for building user defined components.
  - poll  - poll a non-sarracenia web or file server to create notification messages for processing.
  - post & watch - create notification messages for files for processing.
  - sarra _ - download file from a remote server to the local one, and re-post them for others.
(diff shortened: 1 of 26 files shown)


petersilva:
I didn't merge this because I thought some discussion made sense... it is based on feedback from people about not liking or being confused initially by the concept of "components" ... so when I looked at what #379  that plugin from @andreleblanc11 cannot work with any existing component, so needed vanilla... question is... should we make everything vanilla, so there are no components any more? that would be future work... this just makes "vanilla" (aka: flow/ ) work 

Right now, flow/ is just added as an additional component... but it results in having two of everything... one implementation of each component as a sub-class, and another as a config file.  Some of them are redundant.

reidsunderland (approved):
I like the changes. The explanations make sense. 

To avoid the redundancy, for shovel, winnow, post and watch, we could just use `flowMain shovel`, etc. right? Since we have to keep some subclasses, it doesn't seem terrible to keep the others even if they're just a storage location for the default config options

petersilva:
I didn't think adding three reviewers meant all three have to approve... oh well.  we can discuss on Monday, and if bless it then.  would be nice to have the additional reviews, but not the end of the world.

MagikEh (approved):
Looks good, love the docs changes

Issue #382, opened 2021-08-06
sr cleanup deleting exchanges when it should not.


petersilva:
cleanup done by a client is needed to delete queues when removing a consumer... because it will not be in use for a prolonged period.

current algorithm looks at all user configs known (to the machine & user) and if it finds no other dependencies, it removes the exchange, as well as the queue.

It cannot know if there are other users of the exchange. (affects both v2 and v3)

petersilva:
Also user sr_audit --users foreground.   deleted xs_feeder user.

The configuration has a mix of amqp: and amqps: mixed

if you see the entry in the credentials file, but no declared role... probably should make an error message, and not delete anything.

petersilva:
two weeks ago... also deleted users which were not properly declared... causing issues on dev. server.

MagikEh:
We can consider adding simulate/dry-run to just indicate what actions /would/ be taken if the command were run otherwise?

MagikEh:
Issue #465 is related to this, assigning to Clea, close them together?

petersilva:
Raised priority because it is an issue that prevents use of what should be routine maintenance interventions.  Analysts are doing complicated manual interventions to avoid running audit and cleanup functions.

A new suggestions... the only *dangerous* thing is removing exchanges.  So why not put a --dangerWillRobinson guard around deleting exchange?

in order for cleanup to delete exchanges, one has to provide --exchangeDeclare --dangerWillRobinson

normal clean up just will not remove exchanges (but will still declare them.)
That sound like a reasonable resolution of the tension...

petersilva:
Also... exchange deletion might be better construed as an admin function (using rabbitmqctl or rabbitmqadmin) rather than AMQP calls... since AMQP provides no introspection.  Using an administration interface..  can query for  bindings and examine them...

petersilva:
otoh, @tysonkaufmann just found: 

https://www.rabbitmq.com/amqp-0-9-1-reference.html#exchange.delete

there is a bit "if-unused" that can be set on the operation, that would probably solve this problem entirely, by only deleting exchanges that have no bindings.

petersilva:
The if-unused flag makes it super simple... but need to run the flow tests... to see whether it the existing sr3 cleanup code works there... by confirming exchanges which are created by the flow tests, are deleted at the end.  There may be an issue with order of operations, where all queues deletions, or at least bindings should be done before all exchange deletions.

reidsunderland:
[...]

PR #616, opened 2023-01-30
Add `if_unused=True` to exchange delete for #382

reidsunderland:
Add `if_unused=True` to exchange delete for #382

Ensure that sr3 cleanup doesn't delete exchanges that have bindings.

@tysonkaufmann found the option in the documentation :)

### DIFF
--- a/sarracenia/moth/amqp.py
+++ b/sarracenia/moth/amqp.py
@@ -345,8 +345,12 @@ class AMQP(Moth):
 
         try:
             for x in self.o['exchange']:
-                logger.info("deleting exchange: %s" % x )
-                self.channel.exchange_delete(x, if_unused=True)
+                try:
+                    self.channel.exchange_delete(x, if_unused=True)
+                    logger.info("deleted exchange: %s" % x)
+                except amqp.exceptions.PreconditionFailed as err:
+                    err_msg = str(err).replace("Exchange.delete: (406) PRECONDITION_FAILED - exchange ", "")
+                    logger.warning("failed to delete exchange: %s" % err_msg)
         except Exception as err:
             logger.error("failed on {} with {}".format(
                 self.o['broker'].url.hostname, err))
--- a/sarracenia/moth/amqp.py
+++ b/sarracenia/moth/amqp.py
@@ -346,7 +346,7 @@ class AMQP(Moth):
         try:
             for x in self.o['exchange']:
                 logger.info("deleting exchange: %s" % x )
-                self.channel.exchange_delete(x)
+                self.channel.exchange_delete(x, if_unused=True)
         except Exception as err:
             logger.error("failed on {} with {}".format(
                 self.o['broker'].url.hostname, err))
--- a/sarracenia/sr.py
+++ b/sarracenia/sr.py
@@ -1344,8 +1344,8 @@ class sr_GlobalState:
                         print(' remove %s from %s subscribers: %s ' %
                               (qd[1], x, xx))
                         xx.remove(qd[1])
-                        if len(xx) < 1:
-                            print("no more queues, removing exchange %s" % x)
+                        if o.post_broker and len(xx) < 1:
+                            print("No local queues found for exchange %s, attemping to remove it..." % x)
                             qdc = sarracenia.moth.Moth.pubFactory(
                                 o.post_broker, {
                                     'declare': False,


petersilva:
I think the deleting exchange message a few lines above, should also be amended to reflect that the exchange is only deleted if_unused... because the analysts may understandably panic when they see "deleting exchange"....

reidsunderland:
Good catch.  On second thought as well, I need to test deleting an exchange that *does* have bindings and handle an error if there is one.

petersilva:
to do the above test should be simple... like we did before, in static_flow:
```

./flow_setup.sh declare
sr3 cleanup something/with_a_post_exchange.

```

petersilva (commented):
good message...

reidsunderland:
I don't think the logging needs to be changed for v2.

When deleting an exchange that is in use (has queues bound):

<pre>
$ sr_poll sftp_f62 cleanup
2023-02-03 20:13:39,638 [WARNING] extended option logEvents = ['after_accept,after_work,on_housekeeping,post,after_post'] (unknown or not declared)
2023-02-03 20:13:39,638 [WARNING] extended option set = ['sarracenia.moth.amqp.AMQP.logLevel info', 'sarracenia.moth.mqtt.MQTT.logLevel info'] (unknown or not declared)
2023-02-03 20:13:39,638 [WARNING] extended option timezone = ['EST5EDT'] (unknown or not declared)
2023-02-03 20:13:39,638 [WARNING] extended option nodupe_fileAgeMax = ['1000w'] (unknown or not declared)
2023-02-03 20:13:39,638 [WARNING] Should invoke 3: /home/reid/.local/bin/sr_poll [args] action config
2023-02-03 20:13:39,638 [INFO] sr_poll sftp_f62 cleanup
2023-02-03 20:13:39,638 [INFO] Using amqp module (AMQP 0-9-1)
<b>2023-02-03 20:13:39,650 [ERROR] could not delete exchange xs_tsource_poll (tsource@localhost): Exchange.delete: (406) PRECONDITION_FAILED - exchange 'xs_tsource_poll' in vhost '/' in use</b>
</pre>

and when the exchange is not in use:

<pre>
$ sr_poll sftp_f62 cleanup
2023-02-03 20:15:26,533 [WARNING] extended option logEvents = ['after_accept,after_work,on_housekeeping,post,after_post'] (unknown or not declared)
2023-02-03 20:15:26,533 [WARNING] extended option set = ['sarracenia.moth.amqp.AMQP.logLevel info', 'sarracenia.moth.mqtt.MQTT.logLevel info'] (unknown or not declared)
2023-02-03 20:15:26,533 [WARNING] extended option timezone = ['EST5EDT'] (unknown or not declared)
2023-02-03 20:15:26,533 [WARNING] extended option nodupe_fileAgeMax = ['1000w'] (unknown or not declared)
2023-02-03 20:15:26,533 [WARNING] Should invoke 3: /home/reid/.local/bin/sr_poll [args] action config
2023-02-03 20:15:26,533 [INFO] sr_poll sftp_f62 cleanup
2023-02-03 20:15:26,533 [INFO] Using amqp module (AMQP 0-9-1)
<b>2023-02-03 20:15:26,546 [INFO] deleting exchange xs_tsource_poll (tsource@localhost)</b>
</pre>

I don't know what `restore_exchange` is. There's a call to `exchange_delete` in `restore_clear(self)` in `sr_ampq.py` and I added `if_unused=True` there as well. But I don't know how to test that.

There's also `exchange_delete` in [dd_subscribe](https://github.com/MetPX/sarracenia/blob/5155427b3fd5fc96467997087e95e9aa90cb4603/sarra/dd_subscribe#L199). Should the change be made there as well?

petersilva:
dd_subscribe should not exist anymore it doesn't even support v2 format properly, much less v3.

When the exchange delete fails, does that make it loop retrying?

petersilva:
OK, I review the v2 source to remind what restore_exchange is about. ... in v2, there were *save* and *restore* commands.... basically "save" would use *outlet json* and write messages to a save file, in order to drain queues... so it would take all the messages in the rabbitmq input queue, and write them as json to a local file.     Then you would start the component up with the "restore" command, which pointed at a file containing json messages to be put back in the input queue.    This is basically manual retry queues.... functionality that was present before retry_queues were a thing...   retry_queues are kind of an automated way of doing save & restore.

Now this reminds me of why we don't do what @tysonkaufmann was talking about for retry queues... in terms of using broker queues to store retries.  the "save" and "restore" commands that were implemented early on we created because having large queues in a broker is bad. large queues slow down the broker, and there are brokers that have queue size maximums to prevent slowing down processing for everyone using the broker.

So yeah, when there are failures, the retry queues can get very, very large, (hundreds or thousands or millions long) and so using a broker for them was a bad thing. Not to say it can't be revisited, but that was the reasoning at the time.  I think at the time, rabbitmq was purely in memory, and since then I think rabbitmq has the ability to overflow to disk... not sure... anyways, one thing to think about in relation to storing retry queues on brokers.

An alternative to using brokers would be somthing like memcached, which likely has a higher capacity.... probably a more promising way to do shared retry queues.  another thing is once we do that for retry queues, could do it for duplicate suppression also... then the need for winnowing disappears when that is used.  Could be fun!

petersilva:
@reidsunderland as long as it does not loop trying to delete, what you did looks fine.

petersilva:
memory... yeah, before we had retry queues, we just did not read the input queues if we were unable to write to the output ...
we would just stop... so that the queues would back up on the broker, which seemed way simpler and obvious at first... just applying "back pressure" ... that way don't need a second queue at all, just don't read from the input queue unless you can successfully process it.

but when faillures happenned those queues would get too large, and processing on the broker would slow down drastically.  That's why save/restore was implemented in the first place, which was in turn automated by the retry_queue mechanism.

petersilva:
opened #620 to explain this better.

PR #621, opened 2023-02-06
Fix double // on relPath download

### LINKED ISSUE #619: weird // showing up... chdir bug..
* in a case with s3-cloud-publisher example... when it comes to the subscriber.. there seems to be a problem with // showing up in the download process at some point...

perhaps introduced recently by @petersilva (root of all randomness.)

tysonkaufmann:
Short relPath's were generating unnecessary /'s - not an issue on filesystem but causes issues on http(s) downloads.

Changed download to use os.path calls which will fix issues with short paths.

closes #619

### DIFF
--- a/sarracenia/flow/__init__.py
+++ b/sarracenia/flow/__init__.py
@@ -1491,11 +1491,12 @@ class Flow:
         else:
             logger.debug("%s_transport download relPath=%s" % (self.scheme, msg['relPath']))
 
-            # split the path to the file, and the file
-            pathname, remote_file = os.path.split(msg['relPath'])
+            # split the path to the file and the file
+            # if relPath is just the file remote_path will return empty
+            remote_path, remote_file = os.path.split(msg['relPath'])
             # relPath does not contain a prefix / , add it for cdir
-            cdir = '/' + pathname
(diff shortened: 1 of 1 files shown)


petersilva:
question @tysonkaufmann can you try using  os.path.join on windows ... I think we were avoiding using os.path on non-local stuff because I'm confused about what os.path does on windows (where path separator is backslash) vs. everywhere else.  Your patch looks fine for everywhere but windows.  It might be fine there also, would just be worth double checking...

petersilva:
the paths in the messages are OS independent (use '/' ) but when you download or manipulate files locally, they will use the local path separator... this could get highly confusing.

tysonkaufmann:
@petersilva ran os.path.join on window, you are correct, it does incorporate a backslash.

ex:
```
>>> import os.path
>>> dir = "dir1/dir2/dir3"
>>> file = "file1"
>>> print(os.path.join(dir,file))
dir1/dir2/dir3\file1
```

tysonkaufmann:
Added a new commit to remove the os.path.join change, it is unnecessary at this time (was just added to make things cleaner) - however windows doesn't like it.

tysonkaufmann:
oh whoops, I'll doublecheck that next time

Issue #620, opened 2023-02-04
about retry queues, what they are and why they exist.


petersilva:
in the early days of Sarracenia, there were no retry_queues.  There were simply the queues, and when one was having difficulty processing a file, the sarracenia subscriber just kept working on delivering what it had already downloaded the corresponding messages for. Why were retry queues built?

issues:

Queues for long Outages are too big for a broker
----------------------------------------------------------------

when a remote server we are sending to is down, and they are subscribed to a lot of messages, then the queue will build up quickly.  After a few hours, a feed to a storm prediction centre, for example the backlog of messages and corresponding files will be hundreds of thousands, and rapidly climb to millions,  the climat archive, for example, had multiple 7 day outages. Many clients want the data they missed, so the easiest solution is to let the queue build up, 

It is a general tenet of all message passing systems that queues are extremely undesirable.  The preferred status of a queueing system is that all messages are delivered to interested parties and processed by them so that as few as possible are queued at any one point. Rabbitmq literature warns about large queues, and the performance of a broker is affected by the total number of items in all queues on a given broker.  So when a single user suffers an outage, the queue will build up for that user, and all users will see a performance drop.

At any event, when the system is returned the service, the strategy is that, we just what they didn't get in the order it was queued for them, which leads to the second issue:

After long outages, recent data is more important than old data
------------------------------------------------------------------------------------

2 day old weather data is no long weather, it has passed into climatology.  Many clients are in the real-time weather business where the latest data is the most important.  Just having a standard queue build up and then seeing a "catchup" period that could be hours long when our connection to the client is bandwidth limited, just prolongs the outage for them.   What the client wants is for the current data to be fed, and, when there are gaps in current data (so that we are caught up.) then send some old data, but only as long as we aren't blocking any new data.

Some failures are data specific
-----------------------------------------

We have cases where a particular file cannot be successfully transferred.  Reasons are such as that the file exists on the destination, and we are not allowed to delete it, or the target directory is not writable for the sending process.  So one file, or one in a thosand, needs to be retried.  If we just relied on the standard FIFO queue, we just get stuck at this point trying to send this file over and over again... and we never send anything else... This is what Sundew does.  Sarra tries three times, then puts it in a retry queue.
The rest of the data still gets delivered, and we check once in a while, whether delivering the problem file will work later.

How the retry queues work
------------------------------------

[...]

Issue #618, opened 2023-02-03
option to prevent stopping until download retry is empty...


petersilva:
for safe reduction of instances, avoid abandoning transfers that are in need of retry, so refuse to stop until it is empty.

reidsunderland:
I implemented `retryEmptyBeforeExit` for the post retry queue, see https://github.com/MetPX/sarracenia/commit/102620a21f856687ade8f9c39ecf0923af86df4e

petersilva:
should we just augment that to include download_retry as well (use the same setting to cover both?)

petersilva:
background #620

petersilva:
Another thing we could do:  The housekeeping code for the retry module could look at:
* the number of instance retry files extant indicating outstanding retries.
* the number of instances configured.
* the number of instance pid files present... to ensure "bad" instances aren't running.

If it finds instance files related to instances that are beyond the currently configured number, then it could add those items to it's own retry list.  This could be done by each instance on a module instances basis... e.g.  if instances 4 then instance 1 will look for instance 5,   instance 2 will look for 6, etc...

When it finds a higher instance file, it then checks the next modulo... so instance 1 checks for 5, if it finds a n instance 5 file, ir then looks for instance 9.  It processes files from the high numbers back down to low.

This way no change in running code is needed... only an extran operation at maintenance intervals.

petersilva:
also this would solve the problem in all cases, not just cloud.

reidsunderland:
It should be really easy to add the download retry check before exiting as well. retry.py reports the number of messages in each queue to flow

https://github.com/MetPX/sarracenia/blob/7a89be129870532ed7815bed80e10d47b5338b0b/sarracenia/flowcb/retry.py#L105-L111

I like the modulo idea! Instances can shut down without having to wait and the retry queues still get dealt with.

petersilva:
the only routine needing adjustment is on_housekeeping(), and all it has to do is look for the higher numbered retry files, and catenate them to the lower numbered ones before calling the normal housekeeping routine...

petersilva:
The thing with forcing an empty retry is that, the usual reason for retries is that a remote server is down, so forcing the retries to empty means waiting until the remote server is back in service, which could take a very long time, it doesn't seem like a practical option in real life.

petersilva:
so... in the dev meeting, consensus was that this is a blocker... but it's weird because:

[...]

Issue #393, opened 2021-08-24
consider replacing manual queues with more pythonic ones.


petersilva:
* Consider replacing manual queues in moth with more pythonic Queue.
  How to decide.

* Replace sarracenia.flowcb.retry (transliteration of v2) with more pythonic.
  write an alternate module for each implementation.
  Would want to implement multiple options and examine performance
  in high load situation. Options:
   * Queue + sequential file writing
   * persist-queue (single task like today.)
   * persist-queue (shared among all instances?)

* use this to implement cache sharing (copying of cache from processes with
  vip to those without.) General feature makes switchovers much cleaner.
  peer_broker, peer_exchange.

  advertise  shovel/config/recently_received/ and have non vip holders subscribe...

  or permit substitution with memcached?
  Make a peerqueue wrapper class, wait... just have peers subscribe to peer output.

Issue #247, opened 2019-08-22
posting retries


petersilva:
A client has a case where there are multiple nodes downloading, using the same configuration.  A shared file system goes stale on 1 node, so a the messages picked from the queue by that node that try to write to that file system end up in the retry queue. The client is asking that there be a method of bring the retry_queue back to the exchange so that some other process participating in the download can take over.

It's interesting.  Retry was implemented to offload the broker, so that millions of messages don't pile up in the broker queue, this involves handing stuff back to the queue.  needs to be used sparingly and carefully.

petersilva:
This can currently be done manually by stopping the subscriber, and using a shovel configured to -restore from the .retry file.  That´s fairly involved.  Perhaps first step would be to outline that manual process.

PR #644, opened 2023-04-11
Redis implementation of the Retry Queue

G-regL:
In a distributed, containerized, deployment of sr3, each instance might not know about any others, so there has to be a way to share retry queues between them. Files could be mapped/mounted/managed, but that gets out of hand quick quickly.

It was decided that Redis would likely be the best way to go about it, as any and all instances of a given config can share the same queue, and if any of them crash, or get restarted, other instances can pick up the items automatically.

Relates to #620, #618, #393, and #247.

Using this implementation is relatively simple, requiring only a couple options:

- `redisqueue_serverurl`

A server URL for the Redis instance and database you wish to use.
Follows the IANA [`redis`](https://www.iana.org/assignments/uri-schemes/prov/redis) or [`rediss`](https://www.iana.org/assignments/uri-schemes/prov/rediss) URL schemes, as used by the Python `redis` module's [`from_url`](https://redis-py.readthedocs.io/en/stable/connections.html#redis.Redis.from_url) method.
Default is '', meaning it must be set to work.

- `retry_driver`
Needs to be set to `redis`
Default is `disk`, which is the way sr3 has always worked


petersilva:
@G-regL ... .. you changed something in the baseline stuff... now all the checks without redis fail...
I'm seeing cdnld  being stuck and never draining it's queue.  Can you do some more tests with your branch running the flow tests, but with redis queue stuff turned off?  We still need the baseline diskqueue stuff to work.

G-regL:
I replied via Teams, but will include it here as well.

I did some basic testing, and I can confirm that things seems *not* to work with this code in the repo, but *not* using it (neither `retry_driver`, or `redisqueue_serverurl` set in `default.conf`). I can't begin to guess why that is.

Unfortunaly, this is as far as I can take the PR, as I feel like I've contributed a LOT already, and won't have as much time going forward.
I'm happy to implement whatever fix is required to make it work, but finding that fix is beyond my ability and knowledge with sr3.

petersilva:
It was just that the imports were done in a way that meant one must install redis to run sr3 at all.
shuffled things around, so it is only required when in redisqueue is in use.

PR #659, opened 2023-05-03
default wrong for poll & sender on conversion.

petersilva:
Reid found it in senders, and I noticed it in polls on Monday... this needs to be corrected.


petersilva:
Will it pass if one person reviews it? or does it need all of you... was hoping for >= 1.

reidsunderland (approved):
Looks good to me. I don't think it will stop you from merging a PR even if nobody reviews it, at least with our current settings.

Issue #661, opened 2023-05-04
python api example broken. 'dry_run' 


petersilva:
```

fractal% python3 moth_api_consumer.py
options: {'acceptUnmatched': True, 'batch': 25, 'bindings': [('xpublic', ['v02', 'post'], ['#'])], 'broker': None, 'exchange': None, 'expire': None, 'inline': False, 'inlineEncoding': 'guess', 'inlineByteMax': 4096, 'logFormat': '%(asctime)s [%(levelname)s] %(name)s %(funcName)s %(message)s', 'logLevel': 'info', 'messageDebugDump': False, 'message_strategy': {'reset': True, 'stubborn': True, 'failure_duration': '5m'}, 'message_ttl': 0, 'topicPrefix': ['v02', 'post'], 'tlsRigour': 'normal', 'auto_delete': False, 'durable': True, 'exchangeDeclare': True, 'prefetch': 25, 'queueName': 'q_anonymous_fractal_SomethingHelpfulToYou', 'queueBind': True, 'queueDeclare': True, 'reset': False, 'subtopic': [], 'vhost': '/'}
2023-05-04 07:24:46,987 [ERROR] sarracenia.moth.amqp __getSetup connecting to: q_anonymous_fractal_SomethingHelpfulToYou, durable: True, expire: None, auto_delete=False
2023-05-04 07:24:46,987 [ERROR] sarracenia.moth.amqp __getSetup AMQP getSetup failed to hpfx.collab.science.gc.ca with 'dry_run'
2023-05-04 07:24:46,987 [INFO] sarracenia.moth.amqp __getSetup Sleeping 2 seconds ...

```

petersilva:
work-around for people until an upgraded, fixed versiion is available.  Add

'dry_run': False, 

to the set of options supplied to the constructor.

petersilva:
keep this open until there is a CI/CD test to catch this regression.

petersilva:
released in 3.00.38

PR #662, opened 2023-05-04
fixes for #661 examples of python consumer failing.

petersilva:
over time python3 api examples broke.  
quick fix.  I just put one reviewer because in the last pull request, I saw it wants approval from all before mergeing...


reidsunderland (approved):
This change isn't the reason the tests are failing. I think it's good to merge

Issue #667, opened 2023-05-12
after_accept name change not properly propagating in some cases (shovel/winnow/watch/poll/flow)


petersilva:
in cases where an after_accept plugin is used to modify the file name.  if no download occurs, then the new_file value will not be used to update new_relPath, and the the file rename will not carry over into the published messages.  This case will occur in any component that does not transfer a file: e.g flow.

to reproduce:

try to run the flow/scheduled_noaa_tsunami.conf configuration in the https://github.com/MetPX/sr3-examples/tree/main/empty-amqp-pump

when the flow is run, the messages that will result with have constant file names, instead of being pre-pended with datestamps.
sample messages posted:

```

2023-05-12 01:16:19,845 [INFO] 1826220 sarracenia.flowcb.log after_post posted {'_format': 'v03', '_deleteOnPost': {'post_format', 'subtopic', 'new_dir', 'new_subtopic', 'exchange', 'new_baseUrl', '_format', 'new_file', 'local_offset', 'new_relPath'}, 'exchange': ['xs_NOAA-TSUNAMI_public'], 'local_offset': 0, 'pubTime': '20230512T051619.843795538', 'new_file': '20230512051619_PAAQAtom.xml', 'post_format': 'v03', 'new_baseUrl': 'https://www.tsunami.gov/events/xml/', 'new_relPath': 'PAAQAtom.xml', 'new_subtopic': [], 'relPath': 'PAAQAtom.xml', 'subtopic': [], 'baseUrl': 'https://www.tsunami.gov/events/xml/'}
2023-05-12 01:16:19,845 [INFO] 1826220 sarracenia.flowcb.log after_post posted {'_format': 'v03', '_deleteOnPost': {'post_format', 'subtopic', 'new_dir', 'new_subtopic', 'exchange', 'new_baseUrl', '_format', 'new_file', 'local_offset', 'new_relPath'}, 'exchange': ['xs_NOAA-TSUNAMI_public'], 'local_offset': 0, 'pubTime': '20230512T051619.843834877', 'new_file': '20230512051619_PHEBAtom.xml', 'post_format': 'v03', 'new_baseUrl': 'https://www.tsunami.gov/events/xml/', 'new_relPath': 'PHEBAtom.xml', 'new_subtopic': [], 'relPath': 'PHEBAtom.xml', 'subtopic': [], 'baseUrl': 'https://www.tsunami.gov/events/xml/'}

```
note that :

* new_relPath and new_file paths do not agree.  
[...]

PR #668, opened 2023-05-12
V03 issue667

petersilva:
https://github.com/MetPX/sarracenia/issues/667

fix for that...


petersilva:
ugh... I released thinking this had been merged... only to realize it hadn't after the fact.
for this one time, I will merge it... but work towards having proper peer review in future.
(not having this merged breaks some demos in the example course, and I need them.)

Issue #671, opened 2023-05-15
sr3 crashes parsing a config with a bad regex


petersilva:
really strange... 

```

fractal% cat ~/.config/sr3/subscribe/hungry.conf
broker amqp://tsub@localhost

exchange xs_tsource_public

subtopic fruits.#

# print log messages for every file rejected.
logReject on

reject *.qty

# make directories to match the source.
mirror 

# root of the directory where files will be placed.
directory ${HOME}/hungry
fractal%

```
```

fractal% sr3 status
Traceback (most recent call last):
  File "/home/peter/.local/bin/sr3", line 33, in <module>
    sys.exit(load_entry_point('metpx-sr3', 'console_scripts', 'sr3')())
  File "/home/peter/Sarracenia/sr3/sarracenia/sr.py", line 2601, in main
    gs = sr_GlobalState(cfg, cfg.configurations)
  File "/home/peter/Sarracenia/sr3/sarracenia/sr.py", line 1102, in __init__
    self._read_configs()
  File "/home/peter/Sarracenia/sr3/sarracenia/sr.py", line 324, in _read_configs
    cfgbody.parse_file(cfg,c)
  File "/home/peter/Sarracenia/sr3/sarracenia/config.py", line 1389, in parse_file
    self.masks.append(self._build_mask(k, line[1:]))
  File "/home/peter/Sarracenia/sr3/sarracenia/config.py", line 894, in _build_mask
    regex = re.compile(arguments[0])
  File "/usr/lib/python3.10/re.py", line 251, in compile
    return _compile(pattern, flags)
  File "/usr/lib/python3.10/re.py", line 303, in _compile
    p = sre_compile.compile(pattern, flags)
  File "/usr/lib/python3.10/sre_compile.py", line 788, in compile
    p = sre_parse.parse(p, flags)
  File "/usr/lib/python3.10/sre_parse.py", line 955, in parse
    p = _parse_sub(source, state, flags & SRE_FLAG_VERBOSE, 0)
  File "/usr/lib/python3.10/sre_parse.py", line 444, in _parse_sub
    itemsappend(_parse(source, state, verbose, nested + 1,
  File "/usr/lib/python3.10/sre_parse.py", line 669, in _parse
    raise source.error("nothing to repeat",
re.error: nothing to repeat at position 0
fractal% 

```

petersilva:
ugh... it's just bad user input... have to wrap in a try/except and print a reasonable message... no big deal.

[...]

PR #672, opened 2023-05-15
fix for #671 wrap regcomp in try/except to clearly deal with bad regex.

petersilva:
It's a fix, not much to say... offering to a lot of people to review...


MagikEh:
Spoke about it during the meeting, see #671 nd #688 for discussion points.

PR #676, opened 2023-05-23
Unit testing

gcglinton:
Defined framework and patterns for how to do unit/functional testing using [pytest](https://docs.pytest.org/en/7.3.x/index.html).

Tests written include:
- Retry (using both Disk and Redis)
- DiskQueue
- RedisQueue
- part of NoDupe

There's also a Github Actions workflow defined, but it has not been tested yet.


petersilva:
this is going to take a while to review... 
I have to fix the flow tests with redis first.
then I can run these.

gcglinton:
Just looked at the redis tests again, and I don’t think the Redis service is starting properly. 

I’ll have a look at it if I get time.

Issue #677, opened 2023-05-23
Default Configuration File Not at Expected Location Following Ubuntu Install


jilott:
It appears that sr3 is not looking for the example or default files in the correct location after installation using the Debian/Ubuntu package. To test, please follow the [tutorial ](https://metpx.github.io/sarracenia/Tutorials/Install.html) on a fresh install of Ubuntu 22.04LTS.

When you reach the instruction to call: __sr3 foreground subscribe/data-ingest_ the configuration folders will be created in the user's home directory, but sr3 will return immediately. Similarly, calls to sr3 edit or s3 subscribe will return without launching the editor. It appears that sr3 is not looking in the install location (_/usr/lib/python3/dist-packages/sarracenia/examples/_) for these files.

Is this path configurable with an environment variable?

petersilva:
I agree that example is poor... it was meant to represent that... if you had a configuration that invoked an AMQP broker, that error message should be generated, but if the configuration itself does not exist, then it will not behave as described.

OK, will add building the example configuration file prior to invocation.

petersilva:
#684 is fixed, website updated.

PR #682, opened 2023-05-24
V03 issue677

petersilva:
see #677 for details.


petersilva:
this was supposed to have one commit about installation guide... not all this stuff.
I messed up somewhere.

PR #683, opened 2023-05-24
documentation correction for #677

petersilva:
See #677


petersilva:
had created pull vs. main... which was wrong...

petersilva:
@MagikEh good points,  added a patch.

Issue #681, opened 2023-05-23
python producer api broken...


petersilva:
example poisting got broken at some point...

```

blacklab% python3 examples/moth_api_producer.py
cfg: <sarracenia.config.Config object at 0x7f38684e3c10>
2023-05-23 16:48:10,570 [INFO] sarracenia.moth.amqp __putSetup exchange declared: xpublic (as: amqp://tfeed@localhost)
Traceback (most recent call last):
  File "/home/peter/Sarracenia/sr3/sarracenia/examples/moth_api_producer.py", line 39, in <module>
    m = sarracenia.Message.fromFileData(sample_fileName, cfg, sarracenia.stat(sample_fileName))
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/peter/Sarracenia/sr3/sarracenia/__init__.py", line 469, in fromFileData
    m = sarracenia.Message.fromFileInfo(path, o, lstat)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/peter/Sarracenia/sr3/sarracenia/__init__.py", line 514, in fromFileInfo
    msg.updatePaths(o, os.path.dirname(path), os.path.basename(path))
  File "/home/peter/Sarracenia/sr3/sarracenia/__init__.py", line 682, in updatePaths
    if options.post_topicPrefix:
       ^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'Config' object has no attribute 'post_topicPrefix'. Did you mean: 'topicPrefix'?
blacklab% 
```

fix the problem, and add the examples to the CI/CD so we will notice immediately if it breaks again.

petersilva:
1c81c183bc124195f79d1d2f9972050df66162a9 fixes to examples to work with new
tests in sr_insects that test the API.

petersilva:
OK I was imagining that the CI/CD was fixed... after much faffing:
* the API for publishing had    (broker, properties) as arguments, but often properlties contained 'broker' and looking through the api, there were implementation weirdnesses where sometimes the broker was used, and sometimes the broker property was used.  It is confusing and redundant, so rationalized it.
* now the only argument to the factory methods is properties/options which must include a 'broker' field.  instead of a separate 'broker' argument.
[...]

PR #685, opened 2023-05-24
V03 issue617 (part of it.) "topic" depends more on payload format, than message protocol.

petersilva:
move topic generation away from message protocols and into the payload format code.
the WIS format has a different scheme for determining topic hierarchy, so need a way for the postformat/wis.py to return the "topic" for published messages.

I think the universal idea is that one should be able to determina an appropriate topic from the content of the message.

### DIFF
--- a/sarracenia/__init__.py
+++ b/sarracenia/__init__.py
@@ -679,7 +679,9 @@ class Message(dict):
                 logger.error('missing post_baseUrl setting')
                 return
 
-        if options.post_topicPrefix:
+        if options.post_format:
+            msg['post_format'] = options.post_format
+        elif options.post_topicPrefix:
             msg['post_format'] = options.post_topicPrefix[0]
         elif options.topicPrefix != msg['_format']:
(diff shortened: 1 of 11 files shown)


petersilva:
las patch fixes the breakage at in the publish API ( #681 )

andreleblanc11:
Looks good. Do we all need to approve the pull request for it to merge?

petersilva:
hey @andreleblanc11 it looks like you have not approved it.. you just put a note in it...
that's a good question... how much review is enough, and what a reviewer should do...
I was just going on if one person reviewed it then that was good enough... but it doesn't show you  reviewed it... so I'm still waiting...  I think you could also "rebase and merge" which is the kind of approval that just makes it go... or you can leave that to me... but some kind of approval would be good.

andreleblanc11:
I think one reviewer is probably enough. We are all so busy with other stuff that having all of us review would be a hassle lol. As for what the reviewer should do.. that is a good question. Maybe there should be some sort of standard to be set? (check the commits, insert appropriate comments, etc.). We can add it to the agenda for the next dev meeting

Issue #684, opened 2023-05-24
sphinx document generation deeply broken.


petersilva:
the website https://metpx.github.io/sarracenia is updated by using sphinx in a github action.  This was put in place a year or more ago. in the documentation generation github CI/CD actions, we are using 3rd party defined actions, which have version tags.  They had @v2 everywhere... which, I now realize was referring to python v2... so all those procedures were gradually breaking over the past little while.  A week or two ago,  .github/workflows/doc.yml just stopped working.   So now the documentation web site is no longer being updated.

tried fixing it inline but after 30 commits... I think it needs it's own issue.

petersilva:
with latest merge, all working again.

PR #686, opened 2023-05-25
V03 issue684  (website updating broken.)

petersilva:
This is for #684, naturally.

flailed about trying to figure out why it was broken.  Eventually...
* removed setup-python... it was setting up an alternate python and all the depedency installations would miss it, stuff would fail with missing packages.
* to fix that, I install the dependencies with apt beforehand.
* Also at some point gh-pages got corrupted. 
   * I was getting permission errors, then later:
   * I got something like "your branch is out of sync pull and"... which made no sense since it was in an Action, the clone could not have been fresher. 
* I deleted the gh-pages branch and then it started working again.

leftover: there is a v2 job... there... the current version is v3, but when I put v3 it crashes... It works for now... leaving it at v2.


petersilva:
discussed in weekly dev meeting.

Issue #617, opened 2023-01-31
Generify topic/binding posting.... so it works for multiple networks, not just Sarracenia


petersilva:
when dealing with sarracenia encodings,  the topic tree provides the post encoding format.
but the WMO specifically rejected that, so to support foreign encodings, need to be able too
specify outside the topic tree.  .... so have a post_encoding option that allows one  to override what would normally be provided by post_topicPrefix.

petersilva:
This also means we need to accomodate different topic schemes.  hmm...

petersilva:
more broadly, the point is that we implement a lot of invented conventions in Sarracenia, that work great and make life easier when using sarracenia, but if no-one else uses them in interop cases, they get in the way.

Sarracenia Conventions:

* exchange declaration and naming:    how exchange names are built and permissions assigned.
  *  exchange_suffix, post_exchangeSplit, exchangeSplit...the xs_username convention....  the public permission thing.
  * the concept of exchange does not even exist in mqtt, the sarracenia convention for that is to use the exchange name as the top of the hierarchy.

* topic hierarchies... take the topicPrefix (which defines the encoding) and catenate the folders from the relPath as hierarchical topics.   
  * is the topic a protocol characteristic, or an encoding characteristic, or a combination of both? or just make some new flowcb entry points with defaults that implement the conventions.

Probably need to make all these conventions pluggable, rather than hard-coded as at present.

petersilva:
so for post_encode itself... that was based on the use of encode as the name of the class for the parent class that builds on the wire messages from the internal application representation.  thinking about it,  *encoding" is the wrong word, because:

* encoding typically calls to mind a lower level continuous sort of thing... like character sets and such...
* there is overload of the term in python3 with character encoding which is fairly omnipresent.
* it is also overloaded by the encoding of base64 or utf8 used in message inlining, which are better, more obvious uses of the word.

So changed the name of the "encoding" class to be "PostFormat" ... the PostFormat class accepts an application message, and outputs a structure appropriate to be fed to a protocol API.

There was also a conceptual confusion because there was a "version" tag in the messages which indicated which encoding is in use. so change the "version" property in the message to "_format" , and now the option to implement to control for format posted becomes "post_format" so that option is now implemented.

petersilva:
[...]

Issue #615, opened 2023-01-30
WIS encoding support.


petersilva:
the generic encoding module is there (as per #514) ... this will be a tracking issue for a particular encoding.

not really sure what the format is anymore:

I found this:

https://github.com/wmo-im/wis2-topic-hierarchy

https://github.com/wmo-im/wis2-notification-message

Not sure when it will start making sense to start coding for this... probably want to download wis2box
from time to time... ( https://github.com/wmo-im/wis2box ) and attempt to subscribe... see what the messages actually look like, and work from there.

petersilva:
#698 should resolve this.

petersilva:
included in v3.0.41

PR #698, opened 2023-06-10
V03 issue615 - WIS format messages, as proof of concept of using non-sarracenia ones.

petersilva:
OK, so finally all the pieces are in place to allow support of non-sarracenia messages (talk to external systems).  This is about #615, and #617 mostly.

The work is mostly around this file:

postformat/wis.py   *postformat* has two other classes in it, v02.py, and v03.py for the two native sarracenia formats.  

In order to demonstrate encapsulation of foreign notification formats, using the wis one as an example.
It is defined here:

* https://github.com/wmo-im/wis2-notification-message/
* and there is some code to generate them here: https://github.com/wmo-im/pywis-pubsub

So what had to change?   The generation of a topic in Sarracenia is structured to be nearly automatic.  In other systems,  the way the *topic* field is generated will depend on the network itself, as well as some settings that the user may provide.  So that meant that we now need to supply the options (settings) to the postformat encoding class (where it was formerly self-containted.)

Also had to delegate the topic generation to the postformat itself... because it v03.py messages will have a different topic header than wis.py ones, even though both can be sent over MQTT (the same messaging protocol.)

With this branch... someone can subscribe to a sarracenia datapump and post to a WMO pump and vice versa, exchanging messages in the format appropriate for the pump (v03 for Sarracenia, wis for the WMO pump.)

### DIFF
--- a/.github/workflows/flow_basic.yml
+++ b/.github/workflows/flow_basic.yml
@@ -20,7 +20,7 @@ jobs:
     runs-on: ${{ matrix.osver }}
   
     name: Maintenance test on ${{ matrix.osver }}
-    timeout-minutes: 30
+    timeout-minutes: 10
     
     steps:
       - uses: actions/checkout@v3
--- a/.github/workflows/flow_basic.yml
+++ b/.github/workflows/flow_basic.yml
@@ -20,7 +20,7 @@ jobs:
     runs-on: ${{ matrix.osver }}
   
     name: Maintenance test on ${{ matrix.osver }}
-    timeout-minutes: 10
+    timeout-minutes: 20
     
     steps:
       - uses: actions/checkout@v3
--- a/.github/workflows/flow_mqtt.yml
+++ b/.github/workflows/flow_mqtt.yml
@@ -5,7 +5,7 @@ on:
     types: [opened, edited, reopened]
   push:
     branches:
-      - turned_off_for_now_always_fails
+      - v03_wip
 
   workflow_dispatch:
     inputs:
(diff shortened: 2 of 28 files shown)


petersilva:
new issue #699 to addressing missing type field within links field of wis format.

petersilva:
If folks are interested in trying stuff out... I have added a branch to sr3-examples that uses an mqtt broker instead of the rabbitmq one, and publishes wis messages to this broker, and then uses a subscriber to consume them.

https://github.com/MetPX/sr3-examples/tree/sr3_issue615/empty-mqtt-pump

I think there is sufficient stuf in the branch to close off #615 #617 and #699

petersilva:
sigh... when I added the "contentType" generation, that added a dependency on a package that one obtains in python using "import magic"... the package is named "python-magic" in pypi (why not plain magic?) and python3-magic in the repos for ubuntu 22.04.  but if you go to ubuntu 20.04... it seems around then there were two python packages named python-magic, and one got renamed (they are different packages with different APIs unrelated to each other.)  so it seems in older OS's It' hard to figure out what package to have installed.

not clear on what to do about this right now.

petersilva:
summary of discussions at dev meeting:
* the failure is happenning during the apt remove of the previous version, and/or pip install -e of the current version happenning before the pip install of the deps.
* conclusion was to move the package replacement to after a point where all the deps are taken care of.

petersilva:
yay! it passes now!  Someone should merge it now... (preferably squash and merge)

reidsunderland (approved):

reidsunderland:
it got merged into main... do you want me to revert it?

PR #708, opened 2023-06-21
Fix object has no attribute 'onfly_checksum'

reidsunderland:
This tries to use `parent.onfly_checksum` when it should use `msg.onfly_checksum`.

### DIFF
--- a/sarra/plugins/part_check.py
+++ b/sarra/plugins/part_check.py
@@ -51,7 +51,7 @@ class PartCheck(object):
           lag= now-then
 
           if msg.onfly_checksum != msg.checksum :
-             msg = "checksum differ - %s - %s  msg %s" % (msg.new_file, parent.onfly_checksum,msg.checksum)
+             msg = "checksum differ - %s - %s  msg %s" % (msg.new_file, msg.onfly_checksum, msg.checksum)
              if lag > parent.part_check_lag_threshold :
                  logger.warning("part_check might just be referring to an older version of file, but " + msg)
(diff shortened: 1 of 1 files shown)


reidsunderland:
it doesn't work after changing that. The checksum in the message gets replaced before the plugin is called, so it doesn't do what it's supposed to.

petersilva:
unhelpful note:

that's one of the really confusing things about v2... the difference between parent.x, and parent.msg.x is quite unclear.  also parent.x  could be a setting, or a working variable, or a field in the message, or ...  hard to know.

fwiw.,.. I've noticed the dynamic is failing for v2. in case it isn't obvious, the cause of that is elsewhere. I'm studying it.

petersilva:
I'm confused... do you want this merged? or are you saying it does not solve any problem?

petersilva:
oh... was there a crash with a value not found or something... which precipitated the patch?
what do you mean by "it doesn't work?"

reidsunderland:
Please don't merge it :) 

It was crashing, and I "fixed" that and thought it was good so I created a PR. But when I looked closer at the logs, I realized the plugin doesn't do anything useful because as far as I can tell, the msg.checksum has been replaced by the onfly_checksum before the on_data plugins get called.  

Maybe somewhere here: https://github.com/MetPX/sarracenia/blob/v2_dev/sarra/sr_subscribe.py#L302-L324



I haven't figured it out yet, so that's why this is a draft now.

Issue #712, opened 2023-07-04
Crash when plugin declares a duration option without a default value


reidsunderland:
When a plugin declares a duration option without a default value (and the option isn't set in the config file):

```
2023-06-29 21:01:38,044 [DEBUG] 56762 sarracenia.flow loadCallbacks details:
Traceback (most recent call last):
  File "/home/ib/dads/ras/sr3/sarracenia/flow/__init__.py", line 222, in loadCallbacks
    plugin = sarracenia.flowcb.load_library(c, self.o)
  File "/home/ib/dads/ras/sr3/sarracenia/flowcb/__init__.py", line 255, in load_library
    plugin = class_(opt)
  File "/users/dor/aspy/ras/.config/sr3/plugins/poll/poll_NASA_CMR.py", line 195, in __init__
    self.o.add_option('temporalNowMinus',    kind='duration') # TODO document, test
  File "/home/ib/dads/ras/sr3/sarracenia/config.py", line 1001, in add_option
    setattr(self, option, durationToSeconds(v,default_value))
  File "/home/ib/dads/ras/sr3/sarracenia/__init__.py", line 265, in durationToSeconds
    if str_value.lower() in [ 'none', 'off', 'false' ]:
AttributeError: 'NoneType' object has no attribute 'lower'
```

Plugin code that caused the issue:

```python
class Poll_nasa_cmr(FlowCB):
    def __init__(self, options):
        super().__init__(options, logger)
        logger.debug("plugin: Poll_NASA_CMR __init__")

        self.o.add_option('temporalNowMinus',    kind='duration')
```

If `durationToSeconds` receives `str_value = None`, should it return 0.0? 

https://github.com/MetPX/sarracenia/blob/459cc4e726e5a84d76db9f1930a3634d1456f966/sarracenia/__init__.py#L251-L284

petersilva:
looks like a good fix. Please submit a pull request.

PR #713, opened 2023-07-05
Resolves issue #712

Elhossein:
Added an if statement to check if `str_value` is not a string 

Previously, the code only checked for `None` as a string. Now it checks for None as a `NoneType` by making sure `str_value` is a `str` type. This could've been implemented by 
`if str_value == None:`
         `return 0`
However, checking that `str_value` is not a string covers more cases.


reidsunderland:
I changed the base branch to v03_wip. But it looks good to me. Thank you @Elhossein

PR #719, opened 2023-07-15
"possibly undeclared" option message should only be printed once.

petersilva:
while working on OpenDCS integration, need to declare an entire database (hundreds of lines of configuration) and every occurrence of the plugin specific module, when in debug results in t "possibly undeclared option" which is odd, but really only need to produce that message once, not for every time the option shows up.

So only print the message the first time each option is encourntered, not every time.


petersilva:
the mqtt failure has nothing to do with the pull.

andreleblanc11 (approved):
Change looks good.

Issue #415, opened 2021-09-11
mqtt over websockets support.


petersilva:
One of the weaknesseses of Message queueing protocols (MQP) is that it is a little different from HTTP.
The easiest/most usual or common type of service to validate for security is HTTP and validating any other protocol through vetting organizations can be a barrier.  It can also mean bulkier mobile apps, having to support additional protocols.

seamless web messaging is a long time ask.  the mosquitto mqtt broker now includes websockets, and so making Sarracenia speak to that as a client would be great.

petersilva:
https://www.youtube.com/watch?v=hWF0nuqMJaM&ab_channel=SteveCope

petersilva:
it looks like one can just add an option "transport=websockets" on the existing mqtt client... so might not need a whole driver... just a way to express new url "mqttws" for web services...

PR #723, opened 2023-07-23
V03 issue415

petersilva:
As per #415, adding support for mqtt over websockets (changes connect verb in use of paho-mqtt library)  This is done by adding w or ws to the broker URL  (mqtt:// is pure mqtt protocol, mqttws:// is mqtt over websockets (using SSL))

Also noticed why the flow tests were failing for MQTT, there was a mistake in handling of exchangeSplit, fixed in this branch.  MQTT might completely work after this is merged.


petersilva:
All the MQTT stuff works... dunno what is going on with redis but this is way more positive tests than any branch before it.
It's better than v03_wip now.

Issue #724, opened 2023-07-23
 sr3 status crashes:   if len(self.states[c][cfg]['instance_pids']) >= 0: KeyError: 'instance_pids'


petersilva:
no sr3 commands complete when some status files are missing:
```

fractal% sr3 status
Traceback (most recent call last):
  File "/home/peter/.local/bin/sr3", line 33, in <module>
    sys.exit(load_entry_point('metpx-sr3', 'console_scripts', 'sr3')())
  File "/home/peter/Sarracenia/sr3/sarracenia/sr.py", line 2610, in main
    gs = sr_GlobalState(cfg, cfg.configurations)
  File "/home/peter/Sarracenia/sr3/sarracenia/sr.py", line 1109, in __init__
    self._resolve()
  File "/home/peter/Sarracenia/sr3/sarracenia/sr.py", line 871, in _resolve
    if len(self.states[c][cfg]['instance_pids']) >= 0:
KeyError: 'instance_pids'

```

petersilva:
PR #725

PR #725, opened 2023-07-23
fix, now it does not do that.

petersilva:
crash #724 of all sr3 (including just status) caused by funny state (nothing run, but configs and state directories present.)


petersilva:
the MQTT test fails are the same as in v03_wip.  fixed by #723  redis... stuff dunno yet..

PR #730, opened 2023-07-26
Sarracenia __init__ unit test

gcglinton:
This test is pretty exhaustive, and probably overkill for the needs, but it's got 90% code coverage so that's pretty good.

There's some odd things that weren't able to provide coverage for though:
- `Message.dumps()`:
  - Couldn't generate anything that wasn't printable, and thus fail the `try` on line 451
- `Message.fromFileInfo()`:
  - It generates a new `msg` at every run, so couldn't get it to do the rename bits on lines 532-536. How sure how those are supposed to work
  - Messages never have the `new_retrievePath` key, so the `if o.rename` block on lines 543-547 throws a key error if it runs (there's a test case for that though)
  - Couldn't get the `try` on line 555 to fail no matter how many directories were striped, so the `except` never runs
- `Message.updatePaths()`:
  - The `new_dir` and `new_file` parameters are optional, but their concatenation on line 686 throws an error if they're not set (there's a test case for this too)
  - With `new_dir` effectively being mandatory, the `if`s on lines 719 and 735 don't really work properly because they can never evaluate to `True`
- "Extras" checking
  - It doesn't work because no way was found to emulate the packages being in a different state than they actually are on the system

Interesting bits/lessons learned when writing these tests:
- Mocked some internal modules/methods/attributes in order to "trick" the code to trigger a certain behaviour
  - Lines 150 and 162 are replacing `random.choice`'s return with `None`
  - Line 409 is faking being Windows for the following lines
  - Line 471 is faking `urllib.request.urlopen`'s return as a static string, bypassing the need to *actually* do an HTTP request
- Some method calls throw errors, and Pytest catches them properly (lines 317, 351, 443)
  - Could be good for doing doing input validation and such

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -55,12 +55,15 @@ If you want to run this in VSCode, and have it do all the things nicely, you'll 
 - [GitLens — Git supercharged](https://marketplace.visualstudio.com/items?itemName=eamodio.gitlens)  
   Not strictly required, but *very* strongly recommended as it makes VS Code's git features fully functional
 
-Beyond that, changing a few options in your settings file will make it all work; thusly:
+Beyond that, changing a few things in your VS Code configs will make it all work.
+
+In `settings.json`, to get all the reports and coverage when running tests, and allow you to run individual tests even if they have dependencies:
 ```json
 {
     "python.testing.pytestArgs": [
         "tests", "-v", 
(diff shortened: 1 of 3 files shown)


petersilva:
more tests always better than no tests... this is great!  if there is something wrong, we can just fix it later...

PR #731, opened 2023-07-26
Update unit testing patterns

gcglinton:
In writing more and more tests, some things have been streamlined, or improved, and this is just back-porting those changes to previous tests.

- Define a `pretty` method that you can use while debugging your rests
  - Use like: `pretty(ThingYouWantToDump)`, then run your tests with `-vv` and `--capture=tee-sys` parameters
- Move `Options` and `Message` creation into methods (versus creating objects manually)
  - Should probably further use `sarracenia.config.default_config()` to generate base config
  - Could also turn them both into a fixture, but then you might have to break each test set into it's own test method so the fixture resets properly
- Use `deepcopy` instead of just straight copy, so tests don't affect eachother
- With the changes above there was some broken tests in `rety_test.py`, so fix those
- Add a note to `retry_teststeps.py` about how it works, and why we're not using it for our tests
  - tl;dr, it doesn't compare outputs from different drivers against a desired output, so we can't ensure drivers are doing/returning the same thing
- In some tests, I was seeing if logs were being generated, but the assertions were only happening inside an `if` block, so they could never fail; Fixed that

### DIFF
--- a/tests/sarracenia/diskqueue_test.py
+++ b/tests/sarracenia/diskqueue_test.py
@@ -11,6 +11,7 @@ def pretty(*things, **named_things):
         pprint.PrettyPrinter(indent=2, width=200).pprint(v)
 
 from sarracenia.diskqueue import DiskQueue
+from sarracenia import Message as SR3Message
 
 class Options:
     def __init__(self):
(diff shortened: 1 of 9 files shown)


petersilva:
will do a squash merge... number of commits not an issue then...

Issue #643, opened 2023-04-11
flowcb cleanups not called when cleanup is run.


petersilva:

petersilva:
This is complicated because generally, the management cli is used for maintenance operations.  The actual configuration is read, but no actual Instance to run it is created, rather the cli uses the global configuration to make the change of state.   So in order to run cleanup parts of flow_callbacks, every time we do a cleanup operation, we need to actually instantiate the configuration? that's really, really expensive... much thinking needed.

petersilva:
PR #733 solves this.

PR #733, opened 2023-07-27
V03 issue643 - on_cleanup now works (needed by redisqueue )

petersilva:
This builds on the previous two pull requests, fixing issue #643 by adding event callbacks
for on_cleanup, on_declare, and on_sanity.  The on_cleanup is needed by redisqueue 
stuff to ensure that resources allocated on a redisqueue can be cleaned out.

Also noticed that the previous work to enable this, had broken cleanup
as there was no call to xxxSetup()


petersilva:
@gcglinton might want to try this branch out, and see if the cleanup works on his redis stuff.
I renamed the routine to match the callback, but haven't tested it. (used flowcb/log to provie callbacks were called.)

Issue #745, opened 2023-08-11
poll does not notice when the filename is always the same but it is getting re-written


petersilva:
if you have a poll with default settings, and it polls a remote resource where the path name is the same at all polling intervals, and (as is usual for polls) one uses cod for the identity method, the sr3 puts the path in the key value, so it will always return the same value:

In v2, it used the mtime by default, so this worked, in v3 the file is always suppressed as a duplicate.

from sarracenia/flowcb/disk.py (and redis.py... re-factor needed ;-) deriveKey method:

```
     elif 'identity' in msg:
            if msg['identity']['method'] in ['cod']:
                # if cod, revert to using the path.
                key = msg['relPath']
            else:
                key = msg['identity']['method'] + ',' + msg['identity']['value'].replace('\n', '')


        if not key:
            if 'mtime' in msg:
                t = msg['mtime']
            else:
                t = msg['pubTime']
            if 'size' in msg:
                key = f"{msg['relPath']},{t},{msg['size']}"
            else:
                key = f"{msg['relPath']},{t}"

```

should likely re-arrange things such that the relpath is only used as the key if no mtime or size is available.

@reidsunderland @andreleblanc11  look right?

reidsunderland:
Yes, that looks right to me.

PR #746, opened 2023-08-12
V03 issue745 Improve poll cod deriveKey & refactor to remove code duplication

petersilva:
This is a fix for #745, just changing the default key derivation.  The main idea was to change one if statement in deriveKey() method to change how key is derived for "cod" case,  But noticed that deriveKey in nodupe.disk.NoDupe and nodupe.redis.NoDupe are the same... there are two identical methods in two weirdly named classes.

Which revealed the need to refactor:

* re-create create sarracenia.flowcb.nodupe.NoDupe as parent class.
* the only method in the above class is deriveKey()
* sarracenia.nodupe.disk.NoDupe, is renamed to sarracenia.nodupe.disk.Disk (as per normal python practice) and it subclasses the parent  NoDupe class
* sarracenia.nodupe.redis.NoDupe, is renamed to sarracenia.nodupe.redis.Redis (as per normal python convention) and it subclasses the parent NoDupe class.

Result: 

* deriveKey is now only exists once, in the parent class, where formerly it was present in both.
* both disk.py and redis.py follow python conventions for module naming.

Then had to update the unit tests to reflect the change. (2nd commit.)

### DIFF
--- a/sarracenia/flow/__init__.py
+++ b/sarracenia/flow/__init__.py
@@ -178,9 +178,9 @@ class Flow:
         # open cache, get masks.
         if self.o.nodupe_ttl > 0:
             if self.o.nodupe_driver.lower() == "redis":
-                self.plugins['load'].append('sarracenia.flowcb.nodupe.redis.NoDupe')
+                self.plugins['load'].append('sarracenia.flowcb.nodupe.redis.Redis')
             else:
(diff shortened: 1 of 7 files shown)


petersilva:
Ideally, would have had separate commits for the re-factor vs. the actual algorithmic change.

petersilva:
discussed and blessed at weekly meeting.

Issue #752, opened 2023-08-29
scheduled plugin doesn't implement please_stop


reidsunderland:
https://github.com/MetPX/sarracenia/blob/main/sarracenia/flowcb/scheduled/__init__.py

The scheduled plugin should implement please stop. Without it, the flow algorithm will try to stop cleanly but fail because the scheduled plugin just keeps sleeping. Then the process has to be killed.

```
sr3d restart flow/scheduled_hydro_quebec

...

ddsr4: stopping: sending SIGTERM  ( 0 ) Done
ddsr4: Waiting 1 sec. to check if 4 processes stopped (try: 0)
ddsr4: Waiting 2 sec. to check if 4 processes stopped (try: 1)
ddsr4: Waiting 4 sec. to check if 4 processes stopped (try: 2)
ddsr4: Waiting 8 sec. to check if 4 processes stopped (try: 3)
ddsr4: Waiting 16 sec. to check if 4 processes stopped (try: 4)
ddsr4: doing SIGKILL this time
ddsr4: Done
ddsr4: Waiting again...
ddsr4: not responding to SIGKILL:
ddsr4:  19102: ['/usr/bin/python3', '/usr/lib/python3/dist-packages/sarracenia/instance.py', '--no', '1', 'start']
ddsr4:  19103: ['/usr/bin/python3', '/usr/lib/python3/dist-packages/sarracenia/instance.py', '--no', '1', 'start']
ddsr4:  24435: ['/usr/bin/python3', '/usr/lib/python3/dist-packages/sarracenia/instance.py', '--no', '1', 'start']
ddsr4:  24437: ['/usr/bin/python3', '/usr/lib/python3/dist-packages/sarracenia/instance.py', '--no', '1', 'start']
ddsr4: starting: .( 1 ) Done
ddsr4: 
```

I think the comment is also out of sync with the code? https://github.com/MetPX/sarracenia/blob/bcf4f4771f9c379722c3f9623f8341a12675ac4c/sarracenia/flowcb/scheduled/__init__.py#L28-L30

Comma separated hours/minutes doesn't work.

reidsunderland:
```
ddsr4: 2023-08-29 17:55:44,453 [INFO] sarracenia.moth.amqp __putSetup exchange declared: 
[...]

PR #754, opened 2023-09-01
V03 issue752 2

petersilva:
fixing #752 
 * it now works for multiple hours in the day.
 * it is based on building a list of appointments, once per day, then waiting until each one.
 * code now uses python datetime module using datetime, time, and timedeltas (like sarracenia durations.) datatypes.
 * notice that housekeeping doesn't run when _waiting_ ... hmm... desirable or not?
 * output much clearer now.

Also: 
* added support to interrupt sleeps for housekeeping intervals, so that happens now.


petersilva:
@gcglinton you might want to try adding unit test to this branch? the new implementation is much easier to test.

tysonkaufmann (approved):
Code looks good - unit testable - Peter has confirmed functionality

PR #756, opened 2023-09-01
Flowcb accept fixes

gcglinton:
While writing unit tests for all the flowcb/accept plugins, I discovered a number of issues.

- Some (most?) of them lacked an `__ini__` function, which is required to set logging level properly
- Many of them had an improper way of getting/setting config options
  - Added `options.add_option` calls as needed
- Finished porting message structure to v3
  - Propertiesare all just dict keys
  - Some functions are gone
  - the 'headers' key no longer exists
- Added better config validation in some cases
- Fixed numerous logic bugs
- Updating formatting
  - Removing many multi-line statements
  - Fixing indentation of comments

I've consulted with @petersilva and team as much as possible when making these changes, but it's possible some new issues (or regressions, I guess you could say) will be found.

Once this PR is merged, I'll submit a separate PR for these plugins' unit tests.

### DIFF
--- a/sarracenia/flowcb/accept/httptohttps.py
+++ b/sarracenia/flowcb/accept/httptohttps.py
@@ -24,8 +24,8 @@ class HttpToHttps(FlowCB):
             if not 'http:' in message['baseUrl']:
                 new_incoming.append(message)
                 continue
-            baseUrl = message['baseUrl'].replace('http:', 'https:')
-            message['set_notice'](baseUrl, message['relPath'])
+            message['baseUrl'] = message['baseUrl'].replace('http:', 'https:')
+            message['set_notice'] = '%s %s %s' % (message['pubTime'], message['baseUrl'], message['relPath'])
             new_incoming.append(message)
 
         worklist.incoming = new_incoming
--- a/sarracenia/flowcb/accept/httptohttps.py
+++ b/sarracenia/flowcb/accept/httptohttps.py
@@ -17,6 +17,8 @@ logger = logging.getLogger(__name__)
 
 
 class HttpToHttps(FlowCB):
+    def __init__(self, options):
+        super().__init__(options,logger)
 
     def after_accept(self, worklist):
         for message in worklist.incoming:
--- a/sarracenia/flowcb/accept/httptohttps.py
+++ b/sarracenia/flowcb/accept/httptohttps.py
@@ -19,13 +19,7 @@ logger = logging.getLogger(__name__)
 class HttpToHttps(FlowCB):
 
     def after_accept(self, worklist):
-        new_incoming = []
         for message in worklist.incoming:
             if not 'http:' in message['baseUrl']:
-                new_incoming.append(message)
                 continue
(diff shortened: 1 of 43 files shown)


petersilva (changes requested):
the fix for *set_notice* is to remove it.  It is not relevant in sr3.

petersilva (commented):
I do not understand the copy.deepcopy(message) in PostOverride... I think we want to override the original have no use for the original, so copying seems pointless.  I could be missing something.

It's a bit strange... not sure it is a blocker though.

sarracenia/flowcb/accept/tohttp.py
petersilva, line 1: this one is a bit terrifying... we don't know why this logic is here... would prefer to just remove the set_notice
and NOT replace it at all... like in here, there is code where it removes slashes... that is superconfusing in v2,
and it might be a little simpler in v3, but it isn't clear that what the v2 code was doing was "the right thing" ...
just remove all msg["notice"] and callse to set_notice() they are strictly to build v2 messages in v2... they have no relevance to sr3.

sarracenia/flowcb/filter/wmo2msc.py
petersilva, line 1: this replacement of set_notice is find... just setting the baseUrl and relPath is reasonable.
I feel bad contradicting myself from before... but this is, I think, the only case (fingers crossed) where set_notice needs some translation to sr3.

sarracenia/flowcb/accept/httptohttps.py
petersilva, line 1: remove msg['set_notice']

gcglinton:
> I do not understand the copy.deepcopy(message) in PostOverride... I think we want to override the original have no use for the original, so copying seems pointless. I could be missing something.
> 
> It's a bit strange... not sure it is a blocker though.

I'm using `copy.deepcopy` because otherwise, messages appended to `new_incoming` were being modified later on (as those appended messages weren't real copies of the original, but references to them). `deepcopy`ing each message into the `new_incoming` fixed that, and produced what I assumed was the desired behaviour.

petersilva:
> 
> I'm using `copy.deepcopy` because otherwise, messages appended to `new_incoming` were being modified later on (as those appended messages weren't real copies of the original, but references to them). `deepcopy`ing each message into the `new_incoming` fixed that, and produced what I assumed was the desired behaviour.

Is it possible that you are filling a list with one message in your tests? It is probablynecessary to copy.deepcopy when putting the messages in the list in the first place. with real data, the situation you describe is a problem that should not occur.

but if that's really happenning, then the copy.deepcopy should happen somewhere far upstream.  Either way it does not make sense to have it where it is.

gcglinton:
Per discussion with Peter, the changes to make are:
- `tohttp` should *just* swap out whatever scheme is in baseURL to be http
- In `postoverride`
  - Don't append to new_incoming, just work on the messages themselves
- In all changes, don't do anything with `message['headers'][<thing>]`, do things directly on `message[<thing>]`
- 
@petersilva, does this seem right?

petersilva (approved):

Issue #758, opened 2023-09-19
adding wiskis query support 


petersilva:
This is a copy of an internal to Government of Canada request... there is more detail there... Briefly:

request from client:

> We would like your assistance in acquiring data from the Ontario Power Energy. They changed their protocol of transfer to the new WISKI system. 
> We would like to do tests with a sample of data before to switch to the new system.

This is a reference to: https://www.kisters.net/wiski/

and there are some python bindings here: https://pypi.org/project/kiwis-pie/

and we also have a VB sample from OPG (omitted because it mixes credentials and code.)

A substantial complication is that one uses a username to obtain a token, and then do the queries using the tokens for authentication.

@Rehouma63 provided an initial proof of concept standalone python script that obtained a sample time series.  I have
re-worked it to fit into sarracenia for eventual deployment.

petersilva:
@Rehouma63 finished his term, so I picked up where he had gotten to and brought it a little further.
Here is a sample file being produced:  

fractal% more /tmp/wiski/ts_23755_2023-09-19h01:19:44.629537+00:00_2023-09-20h01:19:44.629537+00:00.csv
Timestamp,Value
2023-09-19 00:00:00-05:00,390.07
2023-09-19 00:01:00-05:00,390.08
2023-09-19 00:02:00-05:00,390.1
2023-09-19 00:03:00-05:00,390.1
2023-09-19 00:04:00-05:00,390.11
...
(time series covers 24 hours and has a single value in it: Air Temperature)

things that puzzle me:

* do people want more than one value/station? which ones?
* the data seems to only come up for one station, no matter what I do.
* The token code works, but I had it only be invoked when we encountered a 403 (permission denied) and we never do, and it works anyways... so the token auth stuff is pointless?

petersilva:
ok latest commit, I get some data, and write a bunch of timeseries files.

petersilva:
sample released with 3.0.44

PR #759, opened 2023-09-27
758 adding wiskis query support

petersilva:
as per #758 , add a proof of concept WISKI web services polling thing, supporting the token management needed by Ontario Power Generation.

* Generates tokens, supplies them to kiwis_pie (python wrapper module to access WISKI)
* polls the server given by *pollUrl*
* writes timeseries to a local directory given by *directory*
* generates posts using provided baseUrl etc... as per normal post/poll.
* implemented as a flowcb/scheduled plugin.  
* example configuration file provided.
* for testing, can be invoked on command line.
* works on linux and/or windows (different file naming conventions due restrictions on windows.)

generic items:
* noticed a bug in scheduled_intervals (applies to most of the ones deployed) fixed.
* found a bug in scheduled/init as well (missing argument would be found by derived classes.)
* documented use of None (as opposed to NONE) for filename setting... needed when filenames have colons in them.

### DIFF
--- a/docs/source/Explanation/CommandLineGuide.rst
+++ b/docs/source/Explanation/CommandLineGuide.rst
@@ -2501,6 +2501,9 @@ of colon separation in files, the default in Sarracenia is WHATFN
 
 The possible keywords are :
 
+**None**
+ - no Sundew-style filename processing. Leave the filename alone.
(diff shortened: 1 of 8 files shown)


tysonkaufmann (approved):
👍

PR #769, opened 2023-10-11
Add plugins to get data from NASA, with examples

reidsunderland:
A poll plugin to poll NASA's CMR API, and an after_accept plugin that uses a username + password to login to https://urs.earthdata.nasa.gov and generate/retrieve a bearer token for the account and add it to the credentials DB for every baseUrl processed by the plugin.


reidsunderland:
Thanks :)  I thought it was a good example that should be in the public repo.

I really wanted to avoid re-implementing HTTP downloading code to use cookies like I did for the v2 version, so I'm happy the bearer token worked

Issue #770, opened 2023-10-11
Error on Exit on v3.0.44


gc-nrcan-michael:
Follow up to issue #753.

With Sarracenia 3.0.44, the following Dockerfile works, without python-magic:

```.Dockerfile
FROM ubuntu:latest


RUN apt-get update && apt-get install -y --no-install-recommends \
    build-essential \
    libssl-dev \
    libffi-dev \
    python3-dev \
    python3-pip \
    python3-setuptools \
    python3-wheel \
    && rm -rf /var/lib/apt/lists/*


WORKDIR /app

COPY ./app .

RUN pip3 install -r requirements.txt

RUN pip install metpx-sr3
RUN pip3 install metpx-sr3[amqp,mqtt,vip]
RUN pip3 install metpx-sr3


EXPOSE 5671
EXPOSE 5672
EXPOSE 15671
EXPOSE 15672


CMD [ "python3", "./__main__.py" ]

```

Although on exit, there is a new error:

```
********************************************************************************
Docker App ECCC-to-Cloud: Running
********************************************************************************
Sarracenia AMQP URL: "amqps://anonymous:anonymous@hpfx.collab.science.gc.ca"
Sarracenia topic prefix: "v02.post"
Sarracenia binding: "#"
Sarracenia queue name tag: "LOCAL_TEST"
Send One Message Then Quit mode: True
Mock Sarracenia and only receive test data: True
Mock Event Grid send: True
process: iteration 
count = 0, totalMessageCount = 0
listen_for_messages: returning MOCK Sarracenia data
send_message_batch: send_one_message_then_quit is True, truncating message batch
mock_event_grid_send is True, not sending events
events that would have been sent:
event 0:
CloudEvent(source=docker:eccc_to_cloud:amqps://anonymous:anonymous@hpfx.collab.science.gc.ca/v02.post/#, type=eccc-v1, specversion=1.0, id=3b8d1571-fb1e-4620-a577-b2c7d87aff75, time=2023-10-11 22:36:52.236370+00:00)
event.data =       
{'url': 'https://hpfx.collab.science.gc.ca//20230425/WXO-DD/observations/swob-ml/20230425/CNPL/2023-04-25-2100-CNPL-AUTO-swob.xml', 'group': 'SWOBAUTO'}
send_one_message_then_quit is True, exiting    
sarracenia.moth.amqp:deleteing queue q_anonymous_8dea9a90d820_NRCAN_LOCAL_TEST
[...]

PR #777, opened 2023-10-14
770 error on exit on v3044

petersilva:
place holder for #770 ... addresses a harmless error message when a user is running sarracenia in a docker container.  don't merge until a further not confirms that the fix in this branch works.


petersilva:
Client confirmed it worked for him, but I had issues with merging, so I just cherry picked the one commit.
did not use this PR.

Issue #780, opened 2023-10-16
`_find_missing_instances_dir(...` isn't working when disabled configs exist, causing sr3 sanity to not work


reidsunderland:
When a disabled config is encountered, `_find_missing_instances_dir` stops because it gets lost in the directory structure. I added comments showing the directory structure at each step and where the missing `os.chdir('..')` is.

This causes issues with sr3 sanity, and maybe other parts of the code that depend on knowing about missing instances.

https://github.com/MetPX/sarracenia/blob/dedb8d24e2fe600e643911f307648be8482ed411/sarracenia/sr.py#L512-L562

```python
def _find_missing_instances_dir(self, dir):
        """ find processes which are no longer running, based on pidfiles in state, and procs.
        """
        missing = []
        if not os.path.isdir(dir):
            return
        os.chdir(dir)
        # in .../.cache/sr3/
        for c in self.components:
            if os.path.isdir(c):
                if c not in self.configs: continue
                os.chdir(c)
                # in .../.cache/sr3/c
                for cfg in os.listdir():
                    if cfg[0] == '.': continue
                    
                    if cfg not in self.configs[c]: continue

                    if os.path.isdir(cfg):
                        os.chdir(cfg)

                        # in .../.cache/sr3/c/cfg

                        if os.path.exists("disabled"): # double check, if disabled should ignore state.
                            # need to os.chdir('..') here
                            continue

                        for filename in os.listdir():
                            # look at pid files, find ones where process is missing.
                            if filename[-4:] == '.pid':
                                i = int(filename[-6:-4])
                                if i != 0:
                                    p = pathlib.Path(filename)
                                    if sys.version_info[0] > 3 or sys.version_info[
                                            1] > 4:
                                        t = p.read_text().strip()
                                    else:
                                        with p.open() as f:
                                            t = f.read().strip()
                                    if t.isdigit():
                                        pid = int(t)
                                        if pid not in self.procs:
                                            missing.append([c, cfg, i])
                                    else:
                                        missing.append([c, cfg, i])
                        if ( len(self.states[c][cfg]['instance_pids']) > 0 ) or ( len(missing) > 0 ) :
                            # look for instances that should be running, but no pid file exists.
                            for i in range(1, int(self.configs[c][cfg]['instances'])+1 ):
                                if not i in self.states[c][cfg]['instance_pids']:
                                    if i not in self.procs:
                                        if i != 0:
                                            missing.append([c,cfg,i])
                        os.chdir('..')
[...]

PR #781, opened 2023-10-16
Fix for #780, disabled configs prevents finding missing instances

reidsunderland:
#780 describes the problem


petersilva (approved):
clean and neat. and this is an important fix.

Issue #785, opened 2023-10-19
log rollover fails on Windows.


petersilva:
user reported complaint using the self-extracting windows executable, 3.0.44
```
Traceback (most recent call last):
  File "logging\handlers.py", line 74, in emit
  File "C:\Program Files\metpx-sr3\pkgs\sarracenia\instance.py", line 32, in doRollover
    super().doRollover()
  File "logging\handlers.py", line 435, in doRollover
  File "logging\handlers.py", line 115, in rotate
PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\Administrator\\AppData\\Local\\MetPX\\sr3\\Cache\\log\\subscribe_cbc_01.log' -> 'C:\\Users\\Administrator\\AppData\\Local\\MetPX\\sr3\\Cache\\log\\subscribe_cbc_01.log.2023-10-03'
```

. The log file rollover is not working. So the file keeps growing until I manually stop the service and rename it.

petersilva:
I have implemented some work for #784 should help here.
with this patch: https://github.com/MetPX/sarracenia/commit/a6851e719c84e22c7b7454701843bb4e610d166f

I can set up log rollover to be every 30 seconds, and can test.

petersilva:
With the v03_issue784 branch, I set the log to rotate every 30 seconds.  It works perfectly on linux.  On Windows:
* it doesn't rotate, but doesn't crash immediately... after like a few minutes, it fails with the error message reported about threading.

So... it's easy to reproduce... but... why does it think two processes are accessing the file? they aren't... fixing it is not immediately obvious.  I tried asking ChatGPT... it said: don't use multiple processes... ok, but I'm not.  other suggestion: use another python module called loguru, which apparently does a better job.  Bot loguru is semantically different, and won't do what the current logging does.

so...  no obvious solution so far...

petersilva:
Here is this is what we do on linux:
* we use the logging.handlers.TimedRotatingFileHandler to get log messages re-directed to a file.
* we use os.dup2() to re-direct stdout and stderr onto the log file.  This is done using dup2 because we don't only want log messages re-directed, or even print() messages redirected, we want the stdout/stderr of sub-processes we launch to be re-directed into the log as well.  Just re-directing python's idea of stdout does not do that.
* we subclassed the handler so that we move the re-direction to the new file after each rotation.
[...]

Issue #784, opened 2023-10-18
Bug file error traceback - on startup and loading a plugin


iMacadan:
```py
2023-10-18 15:52:01,877 [INFO] sarracenia.flow close flow/close completed cleanly pid: 34672 sender/wxo-dev-be03-wxapp-minify instance: 1
--- Logging error ---
Traceback (most recent call last):
  File "/usr/lib/python3.6/logging/handlers.py", line 72, in emit
    self.doRollover()
  File "/usr/lib/python3/dist-packages/sarracenia/instance.py", line 32, in doRollover
    super().doRollover()
  File "/usr/lib/python3.6/logging/handlers.py", line 398, in doRollover
    for s in self.getFilesToDelete():
  File "/usr/lib/python3.6/logging/handlers.py", line 363, in getFilesToDelete
    result = result[:len(result) - self.backupCount]
TypeError: slice indices must be integers or None or have an __index__ method
Call stack:
  File "/usr/lib/python3/dist-packages/sarracenia/instance.py", line 242, in <module>
    i.start()
  File "/usr/lib/python3/dist-packages/sarracenia/instance.py", line 233, in start
    self.running_instance.run()
  File "/usr/lib/python3/dist-packages/sarracenia/flow/__init__.py", line 381, in run
    if not self.loadCallbacks(self.plugins['load']):
  File "/usr/lib/python3/dist-packages/sarracenia/flow/__init__.py", line 228, in loadCallbacks
    logger.info( f'flowCallback plugins to load: {plugins_to_load}' )
Message: "flowCallback plugins to load: ['sarracenia.flowcb.gather.message.Message', 'sarracenia.flowcb.retry.Retry', 'sarracenia.flowcb.housekeeping.resources.Resources', 'rename.msg_rm_sundew_ext_from_tilde', 'rename.msg_rm_minify_from_filename', 'log']"
Arguments: ()
2023-10-18 15:52:09,922 [INFO] sarracenia.flow loadCallbacks flowCallback plugins to load: ['sarracenia.flowcb.gather.message.Message', 'sarracenia.flowcb.retry.Retry', 'sarracenia.flowcb.housekeeping.resources.Resources', 'rename
```

petersilva:
It looks like a problem with config parsing ofr backupCount... being a string... when it should be a number.

petersilva:
logRotateCount is not declared as a count... config.py problem.

petersilva:
this is the result of a startup:

```
[...]

PR #786, opened 2023-10-19
V03 issue784 (refactored/simplified logBackup parameter parsing.) also now can rollover every 30s

petersilva:
Have not been able to reproduce the problem. #784

The parsing of options for logBackup controlling rollover date to before there was option parsing.  It looks like there was some kind of typed parsing, and somehow a string was being passed to an option that expected an integer.

Revamped out the parsing is done, to just use the existing standard option parsing.  That involved changing the units of logBackupInterval, from days to being an ordinary interval (defaults to seconds, but can put 1d to get days.)

The commits here have the side effect of letting the client do something like:

```
logBackupInterval 30s
logBackupCount 3
```

to have the logs rollover every 30 seconds.  I need That to work on #785 

Before accepting the PR, should probably test on the system where it failed.

### DIFF
--- a/docs/source/Explanation/CommandLineGuide.rst
+++ b/docs/source/Explanation/CommandLineGuide.rst
@@ -321,7 +321,7 @@ View all configuration settings (the result of all parsing... what the flow comp
      'logLevel': 'info',
      'logReject': False,
      'logRotateCount': 5,
-     'logRotateInterval': 1,
+     'logRotateInterval': 86400.0,
      'logStdout': True,
      'log_flowcb_needed': False,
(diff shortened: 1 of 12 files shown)


petersilva:
It turned out that the original failure had nothing to do with logBackup, but was actually about command line argument processing with foreground (see #789 ) the logBackup missing thing was a red herring.

But, the change here is good in that it removes old parameter parsing, and gets rid of the only exception to how time intervals are expressed to the logging module, and gives me a way to easiily test #785.   The other fixes are the *friends we met along the way* resulting from having incorrect broker settings, and seeing the error handling was sub-optimal.

So yeah, this should likely go in.

Issue #783, opened 2023-10-18
sr3 status reports bogus data when no traffic has occurred...


petersilva:
```
ddsr-cmc-dev02:                                          State   Run Retry  msg data   LagMax  LagAvg  %rej     pubsub   messages     RxData     TxData
ddsr-cmc-dev02:                                          -----   --- -----  --- ----   ------  ------  ----   --------       ----     ------     ------
ddsr-cmc-dev02: shovel/copy-ddsr-stage-MSC-CMC           run     1/1     0 -52360077112%   0%    0.00s    0.00s  0.0%  0 Bytes/s   0 msgs/s  0 Bytes/s  0 Bytes/s
ddsr-cmc-dev02: shovel/copy-ddsr-stage-MSC-DMS-DEV       run     1/1     0  99%   0%  181.14s    1.86s  0.0% 15.9 KiB/s  26 msgs/s  0 Bytes/s  0 Bytes/s

```

the bogus numbers are because no traffic has passed.

PR #787, opened 2023-10-19
Correct metrics miscalculations. #783

petersilva:
fixes the display issues, there was an error calculating disconnection time making it a really large negative number if no connection had occurred.

also... -1 values were meant to be defaults for %ages when no data is available, but they look odd. replacing with 0%.


reidsunderland (approved):
That looks good to me

PR #788, opened 2023-10-21
posts with too short topics (just v02.post)

petersilva:
When Daniel was working on a plugin to to minify input json files, it was publishing files
with a topic of just *v02.post* (stripping out the rest of the topic) so the consumers, being bound to a much longer topic, were not receiving the plugin's products. 

Last winter, there was code done to introduce postformat, so that networks other than native sarracenia messages could be supported (e.g. WMO WIS, ICAO SWIM) those protocols have different topic hierarchies, so need to make sarracenia able to flexibly deal with different postformats' topic schemes.

One way of doing that would be to have plugins set topics.  There is logic in postformat for to accept a topic header if it is provided.

It turned out that, when we have plugins from v2, we need to convert the message to v2, and that required changing the message to a v2 one (with topic v02.post) ... when the message is reconverted back to internal format, there was missing logic to convert the topic back to original.

I'm suspicious that there root cause of having the wrong topic is actually that the messages don't have a relPath in them (where relPath is normally used to calculate the topic.) but this masks the problem and is correct (in the sense that I want to be able to override the topic by specifying the header) so while this patch is a good thing, it might not be the end of the story.

### DIFF
--- a/sarracenia/flowcb/log.py
+++ b/sarracenia/flowcb/log.py
@@ -29,7 +29,7 @@ class Log(FlowCB):
         self.o.add_option('logEvents', 'set',
                           ['after_accept', 'on_housekeeping'])
         self.o.add_option('logMessageDump', 'flag', False)
-        logger.info(f'{self.o.component} initialized with: {self.o.logEvents}')
+        logger.info(f'{self.o.component} initialized with: logEvents: {self.o.logEvents},  logMessageDump: {self.o.logMessageDump}')
         if self.o.component in ['sender']:
             self.action_verb = 'sent'
         elif self.o.component in ['post', 'poll', 'watch']:
@@ -41,6 +41,10 @@ class Log(FlowCB):
         else:
             self.action_verb = 'downloaded'
         self.started = nowflt()
+ 
+        self.rxTopicSeparator='.'
+        if hasattr(options,'broker') and options.broker and options.broker.url.scheme.startswith('mqtt'):
+            self.rxTopicSeparator='/'
 
         self.__reset()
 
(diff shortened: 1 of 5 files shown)


petersilva:
wait... it isn't right...

petersilva:
OK... with the latest patches, the normal log messages in after_accept and after_post are more customized:
```

2023-10-21 15:13:47,921 [INFO] sarracenia.flowcb.log after_accept accepted: (lag: 9.46 )  exchange: xsarra subtopic: 20231021.WXO-DD.bulletins.alphanumeric.20231021.SN.KWNB.19 a file with baseUrl: https://hpfx.collab.science.gc.ca relPath: 20231021/WXO-DD/bulletins/alphanumeric/20231021/SN/KWNB/19/SNVD20_KWNB_211700_RRX__31979
   .
   .
   .

2023-10-21 15:13:51,070 [INFO] sarracenia.flowcb.log after_post posted to exchange: x topic: v03.post.20231021.WXO-DD.bulletins.alphanumeric.20231021.SR.KWAL.19 a file with baseUrl: http://localhost:8001 relPath: 20231021/WXO-DD/bulletins/alphanumeric/20231021/SR/KWAL/19/SRCN40_KWAL_211913___54517

```

petersilva:
These new log messages are more focused: less data is shown, but relevant for different phases (after_accept, after_post).  So it should be easier to understand what is happenning without turning on debugging.

Also, because of the log message changes, this branch will not pass the flow tests until the corresponding branch from sr_insects is merged...  The different log formats are conflicting so have to merge one with errors for the other to pass.

https://github.com/MetPX/sarracenia/pull/788

iMacadan:
The new focused log messages is great improvement for newbies like me !

sarracenia/flowcb/log.py
```diff
@@ -81,7 +138,7 @@ def after_accept(self, worklist):
                         "%s rejected: %d %s " %
                         (msg['relPath'], msg['report']['code'], msg['report']['message']))
                 else:
-                    logger.info("rejected: %s " % self._messageStr(msg))
```
reidsunderland, line 84: can we keep the original full message output as a debug-level log?

petersilva: already have logMessageDump  to dump the entire message instead of the default subset selection.

PR #792, opened 2023-10-23
V03 issue785 workaround

petersilva:
Get windows log rotation working for the time being.  Give the client a way of dealing with it.


petersilva:
hang on... this one needed re-basing... will re-open.

PR #799, opened 2023-10-30
fixes for mirror off and relPath settings from it.

### LINKED ISSUE #791: sr3 relative path on modified downloads, wrong reposts in some cases.
given subscriber settings like:
* prefix v03
* postBaseDir /A/B/C
* directory /A/B/C/D/E
* mirror off.

a file  mine.txt obtained will be written to:  /A/B/C/D/E/mine.txt   

* topic: v03, and a relPath of "mine.txt"  

This is fine for downloading, as it gets it written to the right place (and topic does not matter) but incorrect when re-posting for subscribers.  The correct values would be:

* topic:v03.D.E   relPath: D/E/mine.txt

Need to confirm that what the code does, and correct it.   basically, mirror should affect download path, but the message should be correct relative to baseDir... not affected by relPath in download.

### LINKED ISSUE #796: v3.0.45 mirror off doesn't
the mirror setting is always on in v3.0.45.

petersilva:
There mirror processing in sarracenia/flow/__init__.py ... updateFieldsAccepted(...) routine had two problems with the mirror processing, that broke it quite thoroughly.
The real tragedy was that none of the flow tests caught it.

* closes #796 
* closes #791 

With the changes here, It will pass the tests in https://github.com/MetPX/sr_insects/pull/27

Things that were done:
* there was one line that didn't set relPath correctly for the mirror case.
* there was a second line that incorrectly override new_relPath.

Other stuff that had to be done to get working test cases:
* support of the self.o.fileEvents to choose which fileEvents to act one and forward was missing.
* had to fix up gaps in fileEvents processing in the built in downloading, the sender, and the poll.

### DIFF
--- a/.github/workflows/flow.yml
+++ b/.github/workflows/flow.yml
@@ -26,7 +26,7 @@ jobs:
       # Don't cancel the entire matrix when one job fails
       fail-fast: false
       matrix:
-       which_test: [ static_flow, flakey_broker, dynamic_flow, restart_server ]
+       which_test: [ static_flow, no_mirror, flakey_broker, dynamic_flow, restart_server ]
        osver: [ "ubuntu-20.04", "ubuntu-22.04" ]
 
     runs-on: ${{ matrix.osver }}
--- a/.github/workflows/flow_mqtt.yml
+++ b/.github/workflows/flow_mqtt.yml
@@ -23,7 +23,7 @@ jobs:
       # Don't cancel the entire matrix when one job fails
       fail-fast: false
       matrix:
-       which_test: [ static_flow, flakey_broker, dynamic_flow ]
+       which_test: [ static_flow, no_mirror, flakey_broker, dynamic_flow ]
        osver: [ "ubuntu-22.04" ]
 
     runs-on: ${{ matrix.osver }}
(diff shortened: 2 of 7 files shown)


petersilva:
after both pulls are merged, I will do a separate direct change on the development branch to add the no_mirror test to the tests to run in the flow.yml matrix.

petersilva:
Question for people:  
* the code currently set up so that, if a subscriber is ignoring some events (symbolic link, mkdir , rmdir) it doesn't do them... but it forwards the messages in case someone downstream wants them.  I'm not sure if that's the right thing to do.

petersilva:
OK, so this works... one thing I'm puzzled about...

```

* mirror off
* fileEvents -mkdir,rmdir

```

so like... should mirror off affect fileEvents (because no-one wants *dir operations when mirror is off.) should the mkdir & rmdir be automatically skipped, so one doesn't have to exclude them? 

do people want to send symbolic links? file removes? renames? so ... hard to say... and in any event a watch where people are not expected to build trees, just exclude the events on the watch, so the senders don't need to do anything.

mirror is an option that is within directory ... accept scope ... not global, but fileEvents is a global thing...

PR #801, opened 2023-11-02
#793 - Move metrics to sub-directory (like logs)

### LINKED ISSUE #793: move metrics to sub-directory for cloud compatibility.
@gcglinton is playing with containers and auto-scaling.  He is reporting difficulties reading the metrics with the tools he is using... so it was suggested to change how they are written... if they are placed in a sub-directory, it is easier to map them for use by the container tools.

petersilva:
* closes #793 

co-requisites:

* https://github.com/MetPX/sarrac/pull/117  - corresponding changes in C implementation.
* https://github.com/MetPX/sr_insects/pull/28 - corresponding changes in flow tests.

moves the metrics from being in the state file with the pid, queue, etc... files to it's
own directory parallel to the log directory.   This is more readily harvested by other tools and so is more integration friendly.

### DIFF
--- a/sarracenia/config.py
+++ b/sarracenia/config.py
@@ -360,6 +360,21 @@ def get_log_filename(hostdir, component, configuration, no):
 
     return logdir + os.sep + component + configuration + '_%02d' % no + '.log'
 
+def get_metrics_filename(hostdir, component, configuration, no):
+    """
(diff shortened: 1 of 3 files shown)


gcglinton (approved):
This all looks good to me!

PR #812, opened 2023-11-05
error in docs from #811 acceptUnmatched default value.

petersilva:
acceptUnmatched default is now true everywhere, but some documentation shows it as false.
it's an error.


andreleblanc11:
https://github.com/MetPX/sarracenia/blob/development/docs/source/Explanation/CommandLineGuide.rst?plain=1#L815
https://github.com/MetPX/sarracenia/blob/development/docs/source/fr/Explication/GuideLigneDeCommande.rst?plain=1#L815
Also have acceptUnmatched False

PR #814, opened 2023-11-07
Issue808 setup on start

### LINKED ISSUE #808: `sr3 start poll/*` does not create exchanges when they are missing
During the poll migration of the ddsr nodes, it was noticed that exchanges don't get auto declared when they're not yet created.

It would be nice if sr3 would be capable of declaring exchanges for polls (which are usually the components to declare the exchanges?) when they are missing.

This makes `exchangeDeclare` config option not work as intended.

Steps to recreate problem:

1. Create a basic poll config (with declareExchange to True and with a post_exchange that doesn't exist and a post_broker)
2. Run `sr3 enable $YOURPOLL && sr3 start $YOURPOLL`
3. Do all of this without running `sr3 declare $YOURPOLL`

We see that `getSetup` stalls in incremental backoffs. `putSetup` should likely get called prior, which would properly declare the exchange.

petersilva:
hey @andreleblanc11 ... maybe try this branch... I put the putSetup() in a on_Start in sarracenia/flowcb/post/message.py ... so it should solve the
bind before declare problem in any component.

closes #808

### DIFF
--- a/sarracenia/config.py
+++ b/sarracenia/config.py
@@ -1745,7 +1745,10 @@ class Config:
 
             if (component in ['poll' ]) and (hasattr(self,'vip') and self.vip):
                 if (not hasattr(self,'exchange') or not self.exchange):
-                    self.exchange = self.post_exchange[0]
+                    if type(self.post_exchange) is list:
+                        self.exchange = self.post_exchange[0]
+                    else:
+                        self.exchange = self.post_exchange
                 if (not hasattr(self,'broker') or not self.broker):
                     self.broker = self.post_broker
 
--- a/sarracenia/flowcb/post/message.py
+++ b/sarracenia/flowcb/post/message.py
@@ -63,6 +63,11 @@ class Message(FlowCB):
         else:
             logger.info( "no metrics available" )
 
+    def on_start(self):
+        if hasattr(self,'poster') and self.poster:
+            self.poster.putSetup()
+        logger.info('starting')
+
     def on_stop(self):
         if hasattr(self,'poster') and self.poster:
             self.poster.close()


andreleblanc11:
The changes are working. `sr3 cleanup` is working as it's supposed too. 

A comment not really related to this issue, I'm not sure why we have `putSetup` placed in the `cleanup` routine. `putSetup` tries to connect to the broker (which is already done by `getSetup`) and then declares the exchange. Everytime I run cleanup, the exchange gets declared before getting deleted. 

```
leblanca@edcm-dirt22-3:~/.config/sr3$ sr3 cleanup poll/test_issue808.conf
cleanup: 2023-11-08 13:27:39,329 590239 [INFO] sarracenia.moth.amqp getCleanUp deleting queue q_tfeed.poll.test_issue808.edcm-dirt22-3.edc-mtl.ec.gc.ca
 remove q_tfeed.poll.test_issue808.edcm-dirt22-3.edc-mtl.ec.gc.ca from xs_tfeed_test subscribers: ['q_tfeed.poll.test_issue808.edcm-dirt22-3.edc-mtl.ec.gc.ca']
No local queues found for exchange xs_tfeed_test, attemping to remove it...
2023-11-08 13:27:39,340 590239 [INFO] sarracenia.moth.amqp putSetup exchange declared: xs_tfeed_test (as: amqp://bunnymaster@localhost/)
2023-11-08 13:27:39,353 590239 [INFO] sarracenia.moth.amqp putCleanUp deleted exchange: xs_tfeed_test
```

Declaring the exchange agasin is harmless but just doesn't seem to make sens in the logic of things.

andreleblanc11:
The last 4 dynamic flow tests are failing, I'm not sure if this can be ignored

petersilva:
for the two questions:
* the only entry point to connect to the broker is *Setup (either put or get)... so in order to delete the queue, you need to connect to it, that's why the setup is called.  the declare is a side effect.  One could alter settings on the call to remove the declareQueue setting... before calling putSetup.
* the last four are related to C components... I have ideas, but I don't know what's wrong... yes, hi on the list to investigate.

petersilva:
also note: the "normal" flow tests are passing ... here... it's the redis and MQTT ones that are broken... they are known broken... getting them to pass again is a project...

andreleblanc11:
I think we're good to merge

PR #820, opened 2023-11-11
Issue816 - poll improve how permCopy/timeCopy work

### LINKED ISSUE #816: v2 poll does not produce mode field, sr3 does unless permCopy is off.
Need to add code to sr3 convert to add:

```

timeCopy off

```
in sr3.

petersilva:
closes #816 

Before this patch, if **timeCopy** was off, then a message would be created without *ctime* and *mtime* headers.  Now it is created, but the two headers are added to the *_deleteOnPost* header so they will not be forwarded.  The same thing was done for **permCopy**

So the code can validate that permissions in the poll are readable (because mode bits will be present), even though permCopy is off.   Code can also use the mtime/ctime as part of duplicate suppression or other time based filtering, even though timeCopy is off.

### DIFF
--- a/sarracenia/__init__.py
+++ b/sarracenia/__init__.py
@@ -658,8 +658,9 @@ class Message(dict):
         if lstat is None: return msg
 
         if (lstat.st_mode is not None) :
-            if (o.permCopy and lstat.st_mode):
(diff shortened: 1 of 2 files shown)


reidsunderland (approved):
Looks good to me

PR #822, opened 2023-11-11
adding class so octal values print intuitively.

petersilva:
this came up... with this code, sr3 show displays
```
'permDefault': o0,
 'permDirDefault': o775,
 'permLog': o600,


```

### DIFF
--- a/sarracenia/__init__.py
+++ b/sarracenia/__init__.py
@@ -688,7 +688,7 @@ class Message(dict):
             fh.write(data)
 
         if hasattr(o, 'chmod') and o.chmod:
-            os.chmod(path, o.chmod)
+            os.chmod(path, int(o.chmod))
 
         return sarracenia.Message.fromFileData(path, o, stat(path))
 
--- a/sarracenia/__init__.py
+++ b/sarracenia/__init__.py
@@ -688,7 +688,7 @@ class Message(dict):
             fh.write(data)
 
         if hasattr(o, 'chmod') and o.chmod:
-            os.chmod(path, int(o.chmod))
+            os.chmod(path, o.chmod)
 
         return sarracenia.Message.fromFileData(path, o, stat(path))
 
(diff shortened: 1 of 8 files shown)


petersilva:
OK original version was buggy... I thought it was too simple to bother testing... but it was me that was too simple... all the flow tests fail, it turns out python doesn't convert things to int without you telling it to so have to add conversions at all makedirs and chmod calls in the code.

patches coming...

reidsunderland:
What if you make class octal_number inherit from the built in int, and just override the __str__ and __repr__ methods?

reidsunderland:
```python
class octal_number(int):
    def __repr__(self) -> str:
        return oct(self)

    def __str__(self) -> str:
        return oct(self)

test = octal_number(0o755)
print("octal", test)
print("decimal", int(test))
```

Output:

```
octal 0o755
decimal 493
```

petersilva:
With @reidsunderland's comment reverted the previous commit that was converting everything.

petersilva:
Sigh...  for documentation purposes, after the merge, I don't understand why things were passing, but for what it's worth, tests started failing later, so I had to add a __new__ routine, and that fixed things...  It was doing this:

```

  File "/home/peter/Sarracenia/sr3/sarracenia/instance.py", line 248, in <module>
    i.start()
  File "/home/peter/Sarracenia/sr3/sarracenia/instance.py", line 239, in start
    self.running_instance.run()
  File "/home/peter/Sarracenia/sr3/sarracenia/flow/__init__.py", line 491, in run
    self.do()
  File "/home/peter/Sarracenia/sr3/sarracenia/flow/subscribe.py", line 27, in do
    self.do_download()
  File "/home/peter/Sarracenia/sr3/sarracenia/flow/__init__.py", line 1549, in do_download
    if self.mkdir(msg):
  File "/home/peter/Sarracenia/sr3/sarracenia/flow/__init__.py", line 1395, in mkdir
    mode=int(mode,base=8)
TypeError: int() can't convert non-string with explicit base
fractal%

```

fixed with 89a18c5427f153cbb6caa330b26685636e1efc79

PR #823, opened 2023-11-14
First attempt at switching to AMQP consumers instead of using basic_get

reidsunderland:
Adds a new Moth implementation that uses consumers instead of basic_get.

Enabled by setting `amqp_consumer True` in the config.


petersilva:
Looks great!

PR #830, opened 2023-11-17
Issue828 - log message printing the wrong exchange.

### LINKED ISSUE #828: log message after_post prints reception exchange instead of transmission.
log messages about after_post are misleadingly printing the exchange on which a message was received, rather than sent  in cases where messages are initially received and then forwarded.

petersilva:
* closes #828 

@andreleblanc11 ran into a case where it became obvious that when a component is subscribing and republishing, the post message prints the name of the exchange the message was received from, rather than the one it was posted to.

### DIFF
--- a/sarracenia/flowcb/log.py
+++ b/sarracenia/flowcb/log.py
@@ -103,12 +103,9 @@ class Log(FlowCB):
             return msg.dumps()
 
         s = "to "
-        if 'exchange' in msg: 
-            if ('post_topic' in msg) and not msg['post_topic'].startswith(msg['exchange'][0]):
-                if type(msg['exchange']) is list:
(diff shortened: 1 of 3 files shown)


andreleblanc11:
Weird that so many of the flow tests are failing.. 

The changes look good. I'm not sure why you've added https://github.com/MetPX/sarracenia/pull/830/commits/09c769fb634752f56cc460c2c374db164a3708f0#diff-b63c7bafd4f6cea2e9dab58f12ac8113a0b60da6f12077c3680506c56cb89d1eR684

andreleblanc11:
Yes you're right. I've skimmed by the changes too quickly.. oops

Issue #827, opened 2023-11-15
what should sr3 --dangerWillRobinson remove do?


petersilva:
This is the sr3 version of "rm -rf /" ... with help, can be recovered, but it's painful... There are no interactive prompts in the entire application... the --dangerWillRobinson is the only thing between a user and oblivion... is this too dangerous should we do something else?

petersilva:
@junhu3  @reidsunderland

reidsunderland:
`sr3 --dangerWillRobinson remove` will delete all configs. It's consistent with `sr3 start` and `sr3 stop`. But also dangerous and kind of easy to use by accident (if you start typing the command, and accidentally hit enter before specifying the configs).

One option is forcing a user to explicitly specify `sr3 --dangerWillRobinson remove */*` if they really want to remove every config. This would make it inconsistent with other actions like stop and start, but makes it harder to accidentally delete everything. It also avoids having to add a warning or "are you sure you want to delete everything??" prompt.

petersilva:
what if  --dangerWillRobinson had an argument... and the argument had to equal the number of configurations being affected.
so... if you do 
```

sr3  --dangerWillRobinson 2  remove poll/a*

```
then it will only proceed if number of configurations that result from expanding poll/a* is exactly 2.

petersilva:
it would perhaps default to two... since for 1 you don't need the option...

andreleblanc11:
I like Reid's option of using `sr3 --dangerWillRobinson remove */*`

Should we just just outright ban the use of `--dangerWillRobinson` without any arguments?

petersilva:
Making --dangerWillRobinson take an argument that has to match has a lot more value... it will stop all mismatch mistakes...  I
* */* is easy to specify also, it doesn't solve the problem of someone using --dangerWillRobinson and it still being a mistake, making them provide a second value that must match the first (like MFA) to me is a lot more added security.

petersilva:
example ... someone says sr3 --dangerWillRobinson remove poll \*/\*

they left a space between poll and the \* ... with just the specification, it will go through, but if --dangerWillRobinson 5 is there, indicating it should match exactly 5 configurations, then the \*/\* will match 20, and it will refuse.

petersilva:
have a branch doing it, here is an example:

fractal% **sr3 status**

```
status: 
Component/Config                         Processes   Connection        Lag                              Rates                                        
[...]

PR #831, opened 2023-11-17
one way of addressing #827, making dangerWillRobinson a number that m…

petersilva:
This is a first shot at #827... I think getting a count that matches what is being done is a good thing.  I don't think y/n prompts help, because the analyst will then ask for a -y option, and in any event if the analyst didn't notice the problem immediately, they could answer yes anyways.  having to supply a number that matches the number of configs affected is almost like a 2nd factor.  Correlation is a good thing.

So this patch does that, but more is needed prior to merging this:
* documenting the change in behaviour  (cleanup and remove want the number, other ops just require --dangerWillRobinson=1 ) 
* when doing sr3 stop and you want to stop even foreground tasks, currently that requires --dangerWillRobinson.  That's not super dangerous, so just ask for --dangerWillRobinson=1 for that.

### DIFF
--- a/sarracenia/config.py
+++ b/sarracenia/config.py
@@ -2339,8 +2339,8 @@ class Config:
                             nargs='?',
                             help=' specifical configuration to select ')
         parser.add_argument('--dangerWillRobinson',
-                            action='store_true',
-                            default=False,
+                            type=int,
+                            default=0,
                             help='Confirm you want to do something dangerous')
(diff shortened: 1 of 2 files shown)


petersilva:
If this is good, then merge the sr_insects stuff first:
https://github.com/MetPX/sr_insects/pull/29

petersilva:
merged manually

PR #836, opened 2023-11-18
Issue834 change default filename setting.

### LINKED ISSUE #834: default value of *filename* is *WHATFN*... I think it should be None.
If you have a filename that has colons (:) in it, then the Sundew specific processing has everything after the first colon stripped off, and put into a sundew_extension message field.

If you aren't dealing with Sundew (or worse, have never heard of Sundew) then this is going to be very surprising and hard to figure out.  so the default for the filename setting should be:
```

filename None

```

so that colons are ignored (just treated as part of an ordinary file name.) If people want Sundew specific processing, they should set filename to a non default value.

petersilva:
* closes #834 

The *filename* setting in sr3 is set to 'WHATFN' at the moment. This causes names to be mangled when they contain a colon, for compatibility with Sundew.  When one is not working with Sundew, this split is surprising, and confusing. Users are befuddled and confused by this behaviour.

I looked at v2 source code and the default there is None already, so, this should provide better compatibility with sarra v2. It is unclear why the default is set the way it is.  Should there be a need for further discussion, can always re-open this issue for further discussion.


petersilva:
discussed and merged manually.

Issue #639, opened 2023-03-28
SFTP url's with absolute and relative paths, a discussion.


petersilva:
sftp is a particular protocol in that it path specified at the end is an actual path, and not something relative from a document root.

sr3 treats all sftp paths as absolute currently. In some other software, there is a convention for using an absolute path in an SFTP url is to use the double slash,

e.g.:

```
     sftp://user@host/my_favourite_dir   -- relative path.
    sftp://user@host//home/user/my_favourite_dir  -- absolute path.

```

The relative path is expected to be relative to the user's home directory.
*Scp* has a different convention  *scp user@host:path* use of the colon separator allows for easy disambiguation.

sources:
* https://www.arubanetworks.com/techdocs/AOS-CX/10.07/HTML/5200-7851/Content/Chp_Cfg_FW_mgt/FW_mgt_cmds/sft-url.htm

It's just a convention, and sometimes it works, but sometimes it doesn't.  there are popular tools that don't support it, like curl:
```
curl sftp://peter@localhost/dirA.sums

```
fails even though the file is in peter's home directory... Need to use:

```

curl sftp://peter@localhost/~/dirA.sums

```

So there is a second possible notation.  Note that one must be implemented on the client.  If you send ~/dirA.sums to
an SFTP server, it will not interpret it correctly, so need to remove the ~/ in the sftp protocol driver... hmm... that actually sounds relatively easy.

Should we support relative paths in sftp URLs?  If so, which convention to follow?

I think ~ is the easier and clearer one.

petersilva:
ok, so manually built a paramiko session to open an sftp connection to localhost.
* when you login, sftp.getcwd() returns None.
* when you then sftp.listdir() it lists the home directory.
* if you sftp.chdir('/tmp'), then sftp.getcwd() -> "/tmp", and listdir() lists that directory... ok.
* sftp.chdir(None) , then sftp.getcwd() -> None, listdir() gives you the home directory.

So seeing "/~" in the path part of a URL, could be interpreted as None... but if you want ~/hoho?
you have to do make it relative.... issue an sftp.chdir("hoho") (assuming you just established the connection.
if you are somewhere else, then it looks like it must be done in two steps:  chdir(None) to get to home, then
a second, relative one.

petersilva:
I tested another server (polled by @andreleblanc11 originally)

``` 

sftp user@host
sftp> pwd
Remote working directory: /HOME
sftp> chdir .
sftp> pwd
[...]

PR #837, opened 2023-11-24
Fix one problem related to #639 is relative paths being absoluted...

petersilva:
fixing one problem related to #639 is relative paths being treated as absolute ones when baseUrl.path=''"  

This solves one problem with relative SFTP paths. I don't think it addresses any problem that anybody ran into, it's just I found a different problem while investigating #639.

prior to PR, tested with:
* flows: static, flakey, dynamic, no_mirror.
* C all the make shim_test (6 different mirroring config tests.)

There are two other PR's that should be merged before this one, matching ones:
* sarrac to change configuration to properly use absolute paths.
* sr_insects to put in a use case for relative paths so (test fails if this PR isn't applied.)

### DIFF
--- a/sarracenia/flow/__init__.py
+++ b/sarracenia/flow/__init__.py
@@ -1694,17 +1694,13 @@ class Flow:
             u = sarracenia.baseUrlParse(msg['baseUrl']) 
             logger.debug( f"baseUrl.path= {u.path} ")
             if remote_path:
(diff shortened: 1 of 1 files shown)


petersilva:
the other PRs:
*  https://github.com/MetPX/sr_insects/pull/30
* https://github.com/MetPX/sarrac/pull/131

PR #838, opened 2023-11-27
V2 issue639

petersilva:
these are fixes for v2 to:
* ignore unimplemented mkdir and rmdir events, rather than having failures and putting them on a retry list.
* understand sftp messages produced by v03.   There is a bug in v2 where all *relPaths* are absolute instead of relative, and there is no connection to the baseUrl.  this branch looks at sftp urls and grabs the path from the baseUrl to interpret it correctly.

The v2 dynamic flow tests are messed up.  I confirmed they were messed up before I changed anything.  they are still badly broken.

### DIFF
--- a/sarra/sr_config.py
+++ b/sarra/sr_config.py
@@ -2807,10 +2807,15 @@ class sr_config:
 
         if flgs == self.lastflg : return
 
-        if flgs in [ '0', 'L', 'R' ]:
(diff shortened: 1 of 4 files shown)


petersilva:
will produce a similar on_message plugin to be able to handle relPaths without a / on the current version as well.

Issue #839, opened 2023-11-27
sr3 does not parse positional parameters in regex's correctly.


petersilva:
sr3 does not interpret grouping in accept regexes for use in substitution patterns in other settings, which works in v2.

reidsunderland:
Config:

```
post_baseDir /apps/sarra/public_data
directory  ${PBD}/${YYYYMMDD}/${SOURCE}/loggernet/${0}/
accept .*.data.(.*)/.*
```

Log:

```
2023-11-24 20:40:24,588 [INFO] sarracenia.flow do_download attempt 3 failed to download sftp://sarra@server//var/opt/CampbellSci/LoggerNet/data/WQL/WQL_final_storage_1.dat to %/apps/sarra/public_data/20231124/MSC-OBS/loggernet/{0/WQL_final_storage_1.dat
```

petersilva:
ok, so ... ${0} in python re stuff refers to the entire input string. so positional parameters are kind of off by one. ${1} would be the first field... so the sarra v2 (and Sundew) are using ${0} as the first field...  suggest adding:
```
sundew_compat_regex_first_match_is_zero   flag:  default False.

```
and put a True in default.conf on internal clusters?

(UPDATED: changed name of bad flag)

reidsunderland:
i think it would be better to leave it False in our internal clusters and have the v2 -> sr3 converter add it to each config

petersilva:
hey @reidsunderland ? If you can dig up an example with multiple positional parameters, that would make it easier to build a second test case for that...

reidsunderland:
This isn't one we actually use, just an example I came up with using public data. The config should be compatible with v2 or sr3, 

```
broker amqps://anonymous:anonymous@hpfx.collab.science.gc.ca
exchange xpublic

subtopic *.WXO-DD.meteocode.#

# /tmp/meteocode/TIME/REGION/OTHER_REGION
directory /tmp/meteocode/${2}/${0}/${1}
#  starting counting at 0            ${0}                         ${1}                ${2}
accept .*.WXO-DD/meteocode/(atl|ont|pnr|pyr|que)/.*/TRANSMIT\.FP([A-Z][A-Z]).*([0-2][1-9][0-6][0-9]Z).*
reject .*
```

v2 logs:

```
2023-11-27 22:40:29,889 [INFO] file_log downloaded to: /tmp/meteocode/2300Z/pnr/NT/TRANSMIT.FPNT11.11.27.2300Z.xml
[...]

PR #840, opened 2023-11-28
fixing Issue #839 - fixing positional grouping in sr3

petersilva:
This branch fixes an incompatibility with v2, where position parameter matching using regexes works.
 
* added some unit tests to config_test.py with single and multiple positional parameters.
* capture the match resturned by re.match when processing an accept line.  store it in a message field.
* add a loop to variableExpansion() to to substitutions using the matching field.
* implement a new setting: *self.sundew_compat_regex_first_match_is_zero*
  setting is needed, because the natural thing, is to have regexes work like python match & groups.
 
unrelated fix also in the same branch:
* stop printing queue declare if passive==True, as it's spamming the logs.

### DIFF
--- a/docs/source/Reference/sr3_options.7.rst
+++ b/docs/source/Reference/sr3_options.7.rst
@@ -1713,6 +1713,30 @@ One can turn off queue binding as follows::
 
 (False, or off will also work.)
 
+sundew_compat_regex_first_match_is_zero (default: off)
+------------------------------------------------------
+
+When numbering groups in match patterns, Sundew groups start from 0.
+Python regular expressions use the zeroth group to represent the entire string, and each match
+group starts from 1. It is considered less surprising to conform to Python conventions,
+but doing so unilaterally would break compatbility.  So here is a switch to use
+when bridging between Sundew, sarra v2 and sr3.  Eventually, this should always be off.
+Examples:
+
+* sundew_compat_regex_first_match_is_zero: True
+* input url:  https://hpfx.collab.science.gc.ca/20231127/WXO-DD/meteocode/que/cmml/TRANSMIT.FPCN71.11.27.1000Z.xml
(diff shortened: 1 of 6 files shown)


petersilva:
discussed... agreed.

gcglinton:
Not that anyone asked, but my $0.02 is that regression tests, or tests that are meant to validate a specific fix, should be pulled out into their own tests.

The name of the test set could have the PR/issue number referenced in them. `test_variableExpansion_pr839`, or `test_variableExpansion_issue839`

At the very least, I suggest to mention it in the comments for a given set of asserts, otherwise it might seem that some tests overlap, and could lead to removal of those specific conditions down the road.

petersilva:
wait... tests have names? We could add issue references... that would be cool!  how does one name tests?

gcglinton:
Yeah, each function/class inside a `*_test.py` file is it's own "test set". So in this case you can just create a new function called `test_variableExpansion_issue839`, and put the tests for just that issue in there, and voila, a well-identified regression test.

The other option, if a regression test covers a number of functions, would be to create a class for them, and group them that way.

With that pattern, tests that are wholly contained within a single module, but span various functions, I'd suggest putting the the test class in the module test file.
However tests that touch a number of modules (and would therefore normally be split across a number of test files to ensure consistency with the idea of "1 file, one module") could live in their own `Test_Regressions` class in a `regression_test.py` file at the root of the `test/sarracenia` directory.

For naming, the classes must start with `Test_` for PyTest to pick them up properly, so in this case, you could create a class called `Test_Regressions`, which would then contain a function called `test_issue839`. Because test names are "fully-qualified" so to speak, the full name of the test contains the path to the test file, the class name, and function name. In this particular case, since the regression test just affects `config.py`, it'd be `sarracenia/config_test.py::Test_Regressions::test_issue839`, or something to that effect.

Personally, I'd go with the class way, as I think it makes the most sense, keeps regressions grouped, well identified, and the pattern can work across multiple files if needed.

The other way of doing this is with [markers](https://docs.pytest.org/en/7.1.x/example/markers.html), but those aren't as obvious to report on, or notice.

PR #842, opened 2023-11-28
Issue841 sr3 convert add / do end of post_baseUrl.

### LINKED ISSUE #841: sr3 all relPaths are relative, so post_baseUrl, ported from v2 likely needs a / at the end.
in v2... the relPaths are often start with / ... which is strange for a "relative" path.  in sr3 this is corrected so that all relative paths are, in fact, relative.  However, this means that if someone wants an absolute path, then the post_baseUrl needs to include that.... so often in v2:
```
post_base_url   http://server
```
should be:
```
post_baseUrl http://server/
```

This modification should be added to "convert" functionality.

petersilva:
closes #841


petersilva:
People want the sundew_compat thing added to convert before merging...

PR #849, opened 2023-11-30
plugin fixes from ddsr dev work...

petersilva:
@andreleblanc11 and @petersilva  deploying stuff to ddsr-dev ran into these issues.


petersilva:
I closed the original PR as the development tree wasn't upto date at that time, this one is current, so it is only 1 commit now.

reidsunderland (approved):

PR #852, opened 2023-12-01
result of andre and peter working on ddsr-dev  transition.

petersilva:
miscellaneous fixes found working on ddsr-dev.


petersilva:
did some local testing and found other issues with local directory/date templates in wiski plugin.
should work now.

PR #859, opened 2023-12-06
V03 release process revision

petersilva:
documenting what was discussed.   
* Read it, and feel free to comment.
* once people basically agree with the text...  don't approve it... just put a comment in this thread.
* will build french version after comments are incorporated, and add to this branch,


andreleblanc11:
I think it looks good.

andreleblanc11:
It would be nice though to have a way of automating those VM builds.

PR #870, opened 2023-12-15
Issue860 fixing sourceFromExchange

petersilva:
I think this fixes it, but I have no tests... The easiest way to test is to include this in the next pre-release, then there are lots of places where it will fail or pass. (people can remove the source put in as a work-around.)


petersilva:
some concerns about inability to reproduce the issue, but have cases on internal dev
machines that can be tested once this is in a pre-release.

PR #871, opened 2023-12-16
Issue824 - last transfer (for files) last message (for message flows) in metrics and status.

### LINKED ISSUE #824: Metrics and settings for last sent/received files
One of the useful features in Columbo is that it shows when each config either last received a file or last transmitted a file. Adding this information to the metrics could make it easier to develop a monitoring tool for sr3.

Columbo uses its [own config file](https://github.com/MetPX/Columbo/blob/master/etc/maxSettings.conf.sample) to set limits on how long is too long since last sending/receiving a file. I think it would be better to put these limits directly in the sr3 config for at least two reasons:
1) it's more intuitive (to me) to set these limits directly in the config file. It's easy to forget about Columbo's maxSettings.conf when setting up or modifying a Sundew config.
2) sr3 can know when something is wrong and report it. 

For example:   
- `metrics_maxDelayRecv`:  when more time than this option's value has elapsed since the last accepted message, this config is "late".
- `metrics_maxDelaySend`: when more time than this option's value has elapsed since the last message was posted or file was sent, this config is "late".

sr3 status already takes up a lot of space on the screen, I'm not sure if it's worth trying to squeeze in detailed information about the last send/receive time, but maybe it could show "late" if a file hasn't been sent or received within the defined time limits.

I think the last reception time is always the time of the last accepted message, but the last transmitted time is a little less obvious. For configs that post, using the last post time should work but for senders that don't post, we would have to use the last successful file send time.

petersilva:
closes #824

### DIFF
--- a/sarracenia/__init__.py
+++ b/sarracenia/__init__.py
@@ -30,6 +30,7 @@ from ._version import __version__
 from base64 import b64decode, b64encode
 import calendar
 import datetime
+import humanize
 import importlib.util
 import logging
 import os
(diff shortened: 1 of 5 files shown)


petersilva:
approved in discussion.

PR #877, opened 2023-12-23
Issue860 4 suppress the default source guess if some other option is specified. And more!

petersilva:
in rc4 the logic uses "source" if it was set, the idea being that if the option, the expectation being that if it wasn't set, it would not be defined.

unfortunately, the finalize logic in config.py establishes default values for source based on the broker username, or failing that, the post_broker username.   since a default value was guessed, "source" is *always* specified.

This change prevents guessing of source values unless no overriding option is specified.


petersilva:
I'm a bit confused.  What I intended to do is:
* do one commit to development to increment version to rc5 directly to development.
* do a second branch called track_mqtt_PR_753, which had a single commit for a name change because a reviewer asked for it, and I knew our mqtt uses it.  ( https://github.com/eclipse/paho.mqtt.python/pull/753 )
* do a third branch issue860_4 with hopefully a final fix for sourceFromExchange.

all three somehow ended up in this PR, and 2/3 are already in development.
I messed up, but not sure how.  anyways, they should all be good.

petersilva:
the PR for mqtt just got merged (took three years!) so that patch is now actually useful at some point in the future.

reidsunderland (approved):

PR #883, opened 2024-01-01
close #880 implements list of vips, ORed.

### LINKED ISSUE #880: Support Multiple VIPs?
I'm wondering if we should support multiple VIP options? The config would run on a particular node as long as it has one of the defined VIPs.

My use case is downloading data from a source that only allows 4 concurrent connections. If the sarra config runs on ops, it would be possible to exceed the 4 concurrent connections. I assume 4 of the 8 nodes would get errors and eventually retry, but there could be contention. Supporting multiple VIPs would prevent this by only allowing 4 nodes to download at a time. There might be a better way.

Another use case I thought of is config portability. You could have a 3 VIP settings in a config, one VIP each for dev, stage and ops, and then you can run the config on any cluster without needing to edit the VIP in the config. That might not be a good/tidy thing to do, but it's a possibility.

petersilva:
close #880  support multiple vips.

as described in the bug report.

### DIFF
--- a/docs/source/Explanation/CommandLineGuide.rst
+++ b/docs/source/Explanation/CommandLineGuide.rst
@@ -377,7 +377,7 @@ View all configuration settings (the result of all parsing... what the flow comp
      'v2plugin_options': [],
      'v2plugins': {'plugin': ['msg_total_save', 'file_total_save']},
      'vhost': '/',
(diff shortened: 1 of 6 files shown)


reidsunderland (approved):
That was fast, thanks :)

PR #885, opened 2024-01-04
reverting quoting in one place, to see if it helps

### LINKED ISSUE #884: sr3 HTTPS polling issue, doesn't happen on v2
I found a case where a standard HTTPS poll that works on v2 doesn't work on sr3.

Here's the log where it works on v2:

```
2024-01-01 19:29:40,965 [DEBUG] sr_http init
2024-01-01 19:29:40,965 [DEBUG] sr_http connect https://example.com
2024-01-01 19:29:40,965 [DEBUG] sr_http credentials https://example.com
2024-01-01 19:29:40,965 [DEBUG] sr_credentials has https://example.com
2024-01-01 19:29:40,965 [DEBUG] sr_credentials get in cache https://example.com https://USERNAME:PASSWORD_REDACTED@example.com None True True False False None
2024-01-01 19:29:40,965 [DEBUG] poll_directory /wifs/data/GRIB2/EGRR/T+06 /local/home/sarra/.cache/sarra/poll/wifs_grib/ls_wifs_data_GRIB2_EGRR_T+06
2024-01-01 19:29:40,965 [DEBUG]  cd /wifs/data/GRIB2/EGRR/T+06
2024-01-01 19:29:40,966 [DEBUG] sr_http cd /wifs/data/GRIB2/EGRR/T+06
2024-01-01 19:29:40,966 [DEBUG] sr_http ls
2024-01-01 19:29:40,966 [DEBUG] sr_http open
2024-01-01 19:29:41,725 [DEBUG] Html_parser parse
```

Doesn't work on sr3:

```
2024-01-01 19:27:34,257 [DEBUG] 36629 sarracenia.transfer.https connect sr_http connect https://example.com
2024-01-01 19:27:34,257 [DEBUG] 36629 sarracenia.transfer.https credentials sr_http credentials https://example.com
2024-01-01 19:27:34,257 [DEBUG] 36629 sarracenia.transfer.https cd sr_http cd /wifs/data/GRIB2/EGRR/T+06
2024-01-01 19:27:34,257 [DEBUG] 36629 sarracenia.transfer.https ls sr_http ls
2024-01-01 19:27:34,257 [DEBUG] 36629 sarracenia.transfer.https __open__ https://example.com//wifs/data/GRIB2/EGRR/T%2B06
2024-01-01 19:27:34,593 [ERROR] 36629 sarracenia.transfer.https __open__ Download failed 4 https://example.com/wifs/data/GRIB2/EGRR/T%2B06
2024-01-01 19:27:34,593 [ERROR] 36629 sarracenia.transfer.https __open__ Server couldn't fulfill the request. Error code: 401, Unauthorized
2024-01-01 19:27:34,593 [WARNING] 36629 sarracenia.flowcb.poll lsdir dest.lsdir: Could not ls directory
```

I compared the v2 and sr3 code and found the difference. In sr3, the path is quoted. The HTTPS server being polled in this case doesn't seem to work correctly in this case.

v2:

https://github.com/MetPX/sarracenia/blob/5b063af501cc843e9623a40da1fc782134c79f0e/sarra/sr_http.py#L166

sr3:

https://github.com/MetPX/sarracenia/blob/62cfca0591b70aaac1585e01ecc8799bbe2fc2b7/sarracenia/transfer/https.py#L222

I tested by removing the `...quote` and it works fine. 

I'm assuming this might have been done because there was a case where the non-quoted path didn't work. But here the quoted path doesn't work...

I'm just not sure how to implement the fix. Possible options: 

- add an option that allows us to change what characters are "safe" for quoting?  Default is `/`, in this case, changing safe to `/+` would fixes the issue. https://docs.python.org/3/library/urllib.parse.html#urllib.parse.quote. I think this is my preferred fix.
- add an option to turn off the quoting?
- just revert back to the v2 code without quoting?
- something else?

petersilva:
close #884  ... 

Looking at why the quoting was added in the first place, not sure there was a good reason.  All the flow tests pass without it.  The description of the original change does not make sense either: It refers to utf8 support, but actual change is about url encoding... I'm not confident this @petersilva guy knew what he was on about in 2021.

Might be worth trying this branch out (issue882) if this solves the problem Also, I gave the branch the wrong name...

### DIFF
--- a/sarracenia/transfer/https.py
+++ b/sarracenia/transfer/https.py
@@ -161,8 +161,7 @@ class Https(Transfer):
             url = self.sendTo + '/' + msg['retrievePath']
         else:
             u = urllib.parse.urlparse( self.sendTo )
-            url = u.scheme + '://' + u.netloc + '/' + urllib.parse.quote(self.path + '/' +
-                                                              remote_file)
+            url = u.scheme + '://' + u.netloc + '/' + self.path + '/' + remote_file
 
         ok = self.__open__(url, remote_offset, length)
 
@@ -219,7 +218,7 @@ class Https(Transfer):
 
         self.entries = {}
 
-        url = self.sendTo + '/' + urllib.parse.quote(self.path)
+        url = self.sendTo + '/' + self.path
 
         ok = self.__open__(url)
 


reidsunderland:
This does fix my problem

tysonkaufmann (approved):

petersilva:
OK, so I will merge it to solve the problem we have... but that guy in 2021 was trying to fix something... It would be better if we had tests so that we don't end up oscillating between getting two use cases that we don't have tests for working.

Issue #403, opened 2021-08-25
sr_watch polling version re-factor to incremental operation.


petersilva:
when polling a tree (not using inotify.) sr_watch does not batch things. It just dumps an entire tree.
This will need to be re-wored before release into an iterator style approach, so if you start in a tree with a million files, it will scan the entire million and present them as a single in memory worklist.  This will have performance problems.  want to incrementally proceed though lists one 'prefetch' batch at a time.

There is an interim fix to pretend it does batching properly, but the memory impact and delay to producing the first file is still there, but at least returns one batch at a time.

an entire directory is read into memory, as one message per file, a better way would be for a stateful traversal to only scan one batch at a time, and work through the entire tree by repeated calls. This has horrible performance on bigger trees.

sr_watch works, so it is not an actual bug, but the performance is really sucks compared to what it could be.
but not fatal problem, since polling is a fallback method anyways.

petersilva:
https://www.youtube.com/watch?v=WR7mO_jYN9g&ab_channel=Socratica

https://www.youtube.com/watch?v=gMompY5MyPg&ab_channel=Socratica

likely a generator is a good approach for this.

petersilva:
note that for sr_poll, because of this issue, you have to do a complete poll of all directories and return one massive list every time so:

* _batch_ makes no sense and is not implemented.
* all notifications published only after the entire poll is completed.

petersilva:
There is another way of looking at this, which is unifying sr_watch polling with sr_poll of file: url's... maybe they should become the same thing?  or sr_watch with a remote URL is a poll... maybe poll goes away? potential for useful simplification by merging these two components.

PR #886, opened 2024-01-04
Add plugin for rate limiting polls

reidsunderland:


petersilva:
*  consistency thing.... The other rate limiting things use _RateMax_ .  ( Limit could be a maximum or a mininum... when used as a mininum, it generates warning messages in the log that we aren't able to poll fast enough.)  the minimum is also sort of a place holder for monitoring tools to alert on.

I looked at config.py to see examples:

```
fractal% grep -i rateMax *.py
config.py:    'post_exchangeSplit', 'prefetch', 'messageCountMax', 'messageRateMax', 
config.py:size_options = ['accelThreshold', 'blocksize', 'bufsize', 'byteRateMax', 'inlineByteMax']
config.py:        'post_rate_limit': 'messageRateMax',
config.py:        self.byteRateMax = 0
config.py:            setattr(self, 'byteRateMax', bytes_ps)
fractal% grep RateMin *.py
config.py:    'messageRateMin'

```
so pollRateMax would be more consistent.  ototh... a rate is about how many items are polled... not how often it is polled. Looking at the code, what is actually being done is to limit how often one is polling.  so you are limiting the interval between polls... so an more accurate description would be _pollIntervalMin_  (specifying a mininum interval between polls, 

Isn't that just _sleep_ in the built-in poll?  
https://metpx.github.io/sarracenia/Reference/sr3_options.7.html#sleep-time

I think you are re-implenting sleep as a plugin... can you try out the sleep setting and let me know what is missing/wrong with it.

petersilva:
um... obviously, if it was so hard to find out what sleep does that you wrote a plugin, there is clearly a documentation/discovery issue...

reidsunderland:
Yeah, it's a bit confusing and it took me a while to wrap my head around why the existing features weren't enough. I could rework this plugin to use better option names or to use a "max directory listing per second" rate to be consistent with the other options. 

The source I'm polling allows a maximum of 120 requests within 1 minute, so I wrote it to ensure that we wouldn't exceed that.

The poll config I was working with has ~8 directories to poll, and I initially thought that it was recursively polling subdirectories with multiple subdirectories for each of the 8 directories (more than 120 directories total). I thought the poll would be making >120 requests (poll_directory/lsdir calls) in a short time. 

This actually wasn't true, there are no subdirectories so it only polls 8 directories (8 requests to the source), but I thought this plugin could still be useful for other cases in the future.

To fix the source rate limit problem I'm running into, the sarra that downloads also needs to be rate limited, and `messageRateMax` should work there. 

---

How is this different than sleep?

Assuming we have 150 directories that need to be polled and we can make a maximum of 120 requests per minute:

- `sleep 5m` would list all 150 directories within a short time period, then sleep for 5 minutes before polling all 150 directories again. This would trigger the source's rate limiting and our IP would get blocked.

- This plugin, with ``pollRateLimit_count 60`` and ``pollRateLimit_period 60s`` would:
    - list 60 directories in a short time (however long it takes the source server to respond)
    - wait for <= 60 seconds
    - list 60 more directories
    - wait for <= 60 seconds
    - list the remaining 30 directories
    - sleep for 5 minutes before starting the process again

Using a "max directory listing per second" rate would also work, but would behave slightly differently. The equivalent of ``pollRateLimit_count 60`` and ``pollRateLimit_period 60s`` would be a rate of 1.0 directory listing per second. This would:
- list 1 directory
- wait until <= 1 second has elapsed
- list 1 more directory
- repeat until all 150 directories have been polled
 - sleep for 5 minutes before starting again

One very small drawback to the rate method is that the poll could be a bit slower when the number of directories to be polled is < 60. If you only need to list 50 directories, the rate method will always take ~50 seconds to do it. The way I implemented it will take <50 seconds, assuming the server responds in <1 second per directory listing.

petersilva:
ok... now I understand better what the PR does.... this goes back to:  #403 If that were done, you might (?) not need this, but without it for sure this is a good thing to have.

back to consistency/naming/settings... so the limit is the number of directory listings per minute... ok... pollDirectoryRateMax then?   All rates are expressed in per-second amounts... (like all intervals are in seconds... hard lesson learned from v2 when some things were in milliseconds ;-)  so to get 120, you would just set 2.   and not have the *period* setting.  

would lsdirRateMax be too obscure... I feel like the thing you are limiting is the rate of directory listings per unit time... so there should be a RateMax in the setting for consistency purposes... but pollDirectory? poll_ls_direcotry? directory_list_rate_max? hmm...

reidsunderland:
I think #403 is similar, but I'm not quite sure it would rate limit the polls. You could poll 1000 directories and only get 5 files (1000 lsdirs), or poll 5 directories and find 1000 files (lots of batches, but only a few lsdirs).

I'll rework the code to use a rate. I'll call it `pollLsdirRateMax` for now but would be happy to change it if we can come up with something better :)

petersilva:
The current poll implementation always goes through the entire tree, and only returns some potentially very large number of file messages after a full traversal.  Your PR slows down that process to respect remote server constraints.  It's the best/simplest/most practical approach to the current problem.

Ideally, as per #403 rather than sleeping, we would return the things we have seen, and continue later from where we left off... in very large trees, that would make a big difference, in smaller trees, or in the case of sleeping, it would also be livelier/faster to post.  I don't know know that it makes a practical difference, but would certainly be a lot more elegant.  Also the generator pattern is very pythonic.

petersilva:
the lack of #403 also makes it, so I don't think poll implements *batch* .  I think the *batch* setting is completely ignored by that component... again, likely not a practical problem, as the thing doing the download does implement it, but it remains an inconsistency between components.

reidsunderland:
I kept the code mostly the same and used `pollLsdirRateMax` to set the period/count variables internally. I also added a log to show what the actual vs desired rate is

petersilva:
The old settings are still there, the add_option should go away and become instance variables instead.  _self.o.pollRateLimit_count_ should become something like: _self.pollRateLimit_count_  as they are calculated from the new setting.

or document all three and how they relate to eachother.

### rationale:

Previous analyst complaints were that, when they found options in the v2 parent structure, they didn't know if it was a user setting or some kind of working variable. For v3, intend to keep working variables and settings apart.  

* Settings are in self.o.   
* variables are instance variables.

petersilva (commented):
I would have preferred one option... but three is fine if they are all documented... merged.

PR #887, opened 2024-01-07
Issue740 3 Adding pyproject.toml support for better installation on python > 3.6

### LINKED ISSUE #740: replace setup.py with pyproject.toml ?
in many environments, I read that setup.py is deprecated, and it looks like we should migrated to pyproject.toml.   I'm reading in here:

* https://tttapa.github.io/py-build-cmake/FAQ.html#:~:text=The%20setup.py%20script%20and,declarative%20configuration%20files%20like%20setup.

it quotes the python setup guide, which says: "The setup.py file should be used only when custom scripting during the build is necessary. "  ...

the current setup.py also copies other files existing in the directory as fields for the metadata, which again is something pyproject.toml doesn't look like it can do.  I mean I would expect any/every project to require the version to be set. and the version will be different every time, so ... the .ini format of .toml has no was to run code in it, so you cannot interrogate the file system to determine the version number... so I guess that counts as "custom scripting" ?
I can't see any package requiring manually setting the version every release...

anyways the pip tools threaten that setup.py support may go away in the future... and that one "should" use pyproject.toml, but since it's completely static, we need to create a script to write the pyproject.toml.

I think I am missing something.

### LINKED ISSUE #863: RPMs for redhat 8 don't include systemd unit file.
The RPMS are built with  python setup.py bdist_rpm. 
* python3.6 is deprecated unsupported.
* setup.py is deprecated/unsupported.
* googling "python setup.py bdist_rpm systemd unit file" resulted in 1 semi-relevant hit:
  https://stackoverflow.com/questions/35448758/using-setup-py-to-install-python-project-as-a-systemd-service
  * otoh, it's from 7 years ago, for python 2, and not for bdist_rpm but for install.

super unclear.

On newer OS, need to use different method to generate RPMs.

petersilva:
closes #740 
closes #863 

was working on #740, and the idea was to replace setup.py, and ended up using hatchling as a replacement build tool, and getting it work quite well on fedora 39 and redhat 9. Had to build a .spec file for rpmbuild as well.

But after a bunch of tests, it became clear that older OS's cannot use the replacements.  The python 3.6 in use on redhat 8 and ubuntu 18 doesn't work with hatchling, and the RPM macros the spec files depend on, are not present in the older eco-system. So the choice was to either maintain the new method on a branch for newer systems, or forget about it until we drop the older OS versions.  

I have been trying to do that with the issue740_hatchling branch in parallel with development, but failing.  It is now different from development in ways I don't understand.  I kept trying to rebase, and we now have "diverged" trees.  So started over from development and cherry-picked what was relevant but remain consistent with development.  Given those difficulties...

A third choice was, since the new methods use different files (pyproject.toml and metpx-sr3.spec) just have both methods present, and use old setup.py on older versions, and the replacement pyproject.toml on newer ones.  That avoids the parallel maintenance issue, and is what this PR is about. There may be some dependency editing required to generate rpms on older versions, but it is straight-forward and documented.

so this PR:

* adds pyproject.toml to the project, using [Hatchling](https://hatch.pypa.io/latest/) as the build tool.
  * for python > 3.6 setup.py is gradually being deprecated in favour of this file.
  * there are many build tools that work with the new format.  chose hatchling after some reading.
  * supports all the same features as the old one. 
* adds metpx-sr3.spec to the project to support use of the native rpmbuild  (replacing python3 setup.py bdist_rpm )
   * using the spec file allows us to add the systemd service files.
   * in redhat >= 9 (fedora 39+, perhaps a few versions earlier) all the dependencies which cause such misery for building on 
   * redhat 8 seem to be present, so the dependencies on RPM packages are there (rather than having to rely on pip)
* ran the flow tests on fedora 39
   * the rpm tools do linting of the python, which complains about escape codes which were fixed.
   * Every test except on result on flakey_broker passes. on flakey_broker, one test fails:  

Ran the flow tests on fedora39, and now they all pass except one test where there.
 
 ```
     test 26 FAILURE: static tree	 (307) directoiries should be posted twice: for 1st copy and linked_dir by shim_f63	 (307)
 ```

on ubuntu 22, all tests pass:
```
test 26 success: static tree	 (307) directoiries should be posted twice: for 1st copy and linked_dir by shim_f63	 (614)
```

I know there are some versions of redhat where output re-direction doesn't work the same... that could be what is wrong.  I Have not figured that out yet, but would rather figure it out later, as all the regression tests pass (  redhat version 9 / fedora39 has never been tested before.)  Being able to generate much better RPM files for modern redhat's is a big improvement.
 
I think that the benefits for newer OS's are such that it is worthwhile, and easier overall to bring the new methods into mainline, rather than having to maintain a parallel branch.


petersilva:
discussed... blessed in meeting.

PR #889, opened 2024-01-09
Close gzip target file before renaming it (to ensure it's flushed)

endlisnis:
After years of chasing this, I found that my previous plugin was causing subtle bugs.

The existing code would let the python garbage collection destroy (and thus flush/close) the gzip file.  This would happen *after* the file was renamed, sometimes many seconds after it was renamed.  This created a race condition in any application that was trying to access these files.

By explicitly closing the file, it causes the contents to be flushed BEFORE the rename.  Closing the race condition window.


petersilva:
looks good.

Issue #565, opened 2022-10-06
Dynamic Bearer Tokens


reidsunderland:
sr3 has bearer tokens built in, but some APIs need a dynamic bearer token, e.g. https://eumetsatspace.atlassian.net/wiki/spaces/DSDS/pages/316014673/API+Authentication

In the EUMETSAT case, you post the username/password to a URL and get a token back.

Since static bearer tokens are built in, it would be nice to have the ability to use dynamic tokens as well.

reidsunderland:
I've written a few plugins that use APIs to generate bearer tokens and add them to the credentials DB, the `auth*.py` ones here: https://github.com/MetPX/sarracenia/tree/development/sarracenia/flowcb/accept

There's a small problem with the implementation of the https transfer class. When the bearer token changes, the https still tries to use the old one, because the bearer_token is cached internally.

https://github.com/MetPX/sarracenia/blob/f466d4b4586b2684ea1f5149dd492d6d45c95c9b/sarracenia/transfer/https.py#L134

Usually this causes an error during the download, then the instance of Https gets closed and re-created, and everything works again. It's not a big problem. 

This could be fixed by calling `self.credentials()` more frequently than just in `connect()`. Maybe in `__open__` or in `check_is_connected`?

reidsunderland:
## Short term fix

Implement re-calling credentials() in HTTPS to prevent the errors. 

@petersilva recommends only calling credentials if there is already a bearer_token.

## Long term
- Store a bearer token expiry in the credentials class.
- Look into implementing a better API for this. In my `auth_....py` plugins, the code in after_accept is mostly repeated. It would be nice to have an entry point (`get_bearer_token`) that can be called directly. That method could be called when there is no token, or the token is expired, and should return a token.

PR #891, opened 2024-01-09
New EUMETSAT plugins; updates for copernicus

reidsunderland:
Improvements/fixes for auth_copernicus. The issue described [here](https://github.com/MetPX/sarracenia/issues/565#issuecomment-1881321496) still applies when the token expires, but after failing once, it starts working again. 

Add EUMETSAT poll/auth plugins.


petersilva:
the test failures are my issue, fixed in a different PR, could always do a cherry-pick if the other PR doesn't doesn't get merged.

PR #897, opened 2024-01-16
Add missing AM headers in amserver

andreleblanc11:
The changes look like a lot but most of them are just indent reformatting's that were done automatically when executing `=G` on vim.

The actual changes are at the end of the `gather`. I recreate the bulletin with the added `CN00 CWAO` on the first line. There's also more verbosity in the code and in the comments

### DIFF
--- a/sarracenia/flowcb/amserver.py
+++ b/sarracenia/flowcb/amserver.py
@@ -284,14 +284,15 @@ class Amserver(FlowCB):
 					##									 Station
 					##										  Random Integer
 
-					
-
 					# Complete specific bulletins, add default header (derived from sundew -> https://github.com/MetPX/Sundew/blob/main/lib/bulletinAm.py)
 					missing_ahl = b'CACN00 CWAO'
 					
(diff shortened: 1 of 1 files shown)


petersilva:
* In python, an indent is four spaces. period.  That's a hard rule in python style guides. It looks like 8 ? 
* re-formatting is fine. code changes are fine... both in the same patches are verboten.

Please start over with:
* patches that ONLY change the code.
* patches that ONLY re-format.

if they are mixed, it is really hard to understand the changes.
It might be fun to add a setting so that the CN00 default is a settable string.

andreleblanc11:
I forgot to make the bulletin contents be accessible by the message `contents` field. Putting this PR on hold for now

andreleblanc11:
I'll close this PR and create a new one

petersilva (changes requested):
an indent in python is 4 spaces, not 8.
CN00 CWAO should be a settable option.
please have patches with formatting changes and code changes completely separate.

PR #900, opened 2024-01-19
Add AM AHL headers and put bulletin contents in message

andreleblanc11:
The changes include:

- Add bulletin contents to the sarracenia message (v03 only) via the contents field.
- Add logic for the built-in AHL header auto-corrections

### DIFF
--- a/sarracenia/flowcb/amserver.py
+++ b/sarracenia/flowcb/amserver.py
@@ -16,16 +16,22 @@ Description:
         To start, when on_start is called, the socket binds, listens and afterwards accepts a connection when one is received from a remote host.
         The receiver acts like an amtcp server. Instances are split into child/parent once an outbound connection is accepted.
         The child proceeds to the receiver service while the parent stays and keeps accepting connections.   
-        In the service, bulletins are received and are stored locally inside of a given filepath.
+        In the service, bulletins are received and are stored in a newly built sarracenia message from the contents field.
+        With the `download` option set to true, the file contents can then be ingested through the sarracenia main flow.
 
     Options:
-        AllowIPs (list): Filters outbound connections with provided IPs. All other IPs won't be accepted. 
-        If unselected, accept all IPs. 
+        AllowIPs (list): 
+            Filters outbound connections with provided IPs. All other IPs won't be accepted. 
+            If unselected, accept all IPs. 
 
-        directory (string): Specifies the directory where the bulletin files are to be stored. 
+        MissingAMHeaders (string):
+            Specify headers to be added inside of the file contents.
+
+        directory (string): 
+            Specifies the directory where the bulletin files are to be stored. 
 
     NOTE: AM protocol cannot correct data corruption and cannot be stopped without data loss.
-    Precautions should be taken during maintenance interventions.
+    Precautions should be taken during maintenance interventions    
 
 Platform:
     Linux/Mac: because of process forking, this will not work on Windows. 
@@ -275,7 +281,7 @@ class Amserver(FlowCB):
                 bulletinHeader = parse[0].decode('iso-8859-1').replace(' ', '_')
                 firstchars = bulletinHeader[0:2]
 
-                # Insert AHL headers in bulletin contents
+                # Treat bulletin contents and compose file name
                 try:
                     ## NOTE: Bulletin filenames have the following naming scheme
                     ##   1. Bulletin header (composed of bulletin type, Issuing office, timestamp)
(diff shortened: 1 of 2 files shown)


petersilva:
the decode should be from 'iso-8859-1' not utf-8 as the AM message bodies are in the older character set.

petersilva:
Also, first attempt encoding as characters, but if that fails, fall back to base64 encoding... there is similar sample code in lines 213+10 in flowcb/post/file.py

sarracenia/flowcb/amserver.py
petersilva, line 1: ident_update(decoded_bulletin)... what is the value of decoded_bulletin in the binary / base64 case

andreleblanc11: Yeah I overlooked that.. I'll add the variable in the try/except to match each case

andreleblanc11: For the binary bulletin, I imagine we can leave the `utf-8` decoding as there is no character set associated with binary files?

petersilva: not all binary is valid utf-8, you will get decode errors... the content is supposed to be either utf-8 or base64... should encode with base64 for binary data.
flowcb/gathe/file.py already has logic to do that.

andreleblanc11:
I've tested the `try/except` with the binary data and it wasn't behaving as expected, binary data was being encoded in `utf-8` in the message.

I substituted that with a case statement where the binary data is determined the same way as Sundew : https://github.com/MetPX/Sundew/blob/main/lib/bulletin.py#L271-L273

andreleblanc11:
Another note,

There are some specific actions that Sundew executes for certain types of bulletins. How would we want to handle these?
We could:
1. Keep these anomalies in the `amserver`, adding them as they are encountered on our feeds
2. Substitute these anomalies in a separate flowcb and adding them as they are encountered on our feeds
3. Try to port the bulletin classes responsible for this on Sundew. @petersilva makes note that this would likely be time costly and would require lots of testing.

An example, https://github.com/MetPX/Sundew/blob/main/lib/bulletinAm.py#L114-L115, `SI*` and `SM*` bulletins require additional specific treatment to complete its contents.

petersilva:
> I've tested the `try/except` with the binary data and it wasn't behaving as expected, binary data was being encoded in `utf-8` in the message.
> 
> I substituted that with a case statement where the binary data is determined the same way as Sundew : https://github.com/MetPX/Sundew/blob/main/lib/bulletin.py#L271-L273

after I thought about it, I realized that all sequences are valid in iso8859-1, so it will never fail to decode....  utf is very specific about the 8 bit characters that are valid, but the older character set... it just represents a different character, so they are all ok.  The method you found makes a lot of sense, but might fail if there are other data types in the stream.  it would be better to make it a a list type option, and just check for membership in the list 'if line[0:4] in self.o.binaryInitialCharacters    where self.o.binaryInitiualCharactes = [ 'BUFR', 'GRIB', '\211PNG' ]

andreleblanc11:
That's a good idea. 

Also, I just noticed that the sarra receiver on development (receiving the data from AM) is failing with

```
2024-01-24 19:04:02,245 [WARNING] sarracenia.flow write_inline_file decoded data size (188 bytes) does not have expected size: (252 bytes
```

This is because in the code, here,
https://github.com/MetPX/sarracenia/blob/d13f5c7f8169fd95298af4550d567f6ef0c3f941/sarracenia/flow/__init__.py#L1136-L1137
https://github.com/MetPX/sarracenia/blob/d13f5c7f8169fd95298af4550d567f6ef0c3f941/sarracenia/flow/__init__.py#L1164-L1168

The length of the **binary** data is taken instead of the decoded `utf-8` string

We should mimic this behaviour with amserver

andreleblanc11:
I guess it doesn't matter as we'll have the downloaded data with https://github.com/MetPX/sarracenia/pull/903

petersilva:
fwiw... in a python interpreter...
```
>>> l="téléphone"
>>> len(l)
9
>>> m="telephone"
>>> len(m)
9
>>> l.encode('utf-8')
b't\xc3\xa9l\xc3\xa9phone'
>>> len(l.encode('utf-8'))
11
>>> 


```
a utf encoded string with accented characters will have more raw bytes than an iso8859-1 one.

petersilva:
for completeness:
```
>>> len(l.encode('iso-8859-1'))
9
>>> l.encode('iso-8859-1')
b't\xe9l\xe9phone
```

petersilva:
This is an interesting question... should the content of the file be in utf-8 or iso-8859-1 ? 
The code is currently reading in iso8859-1 and writing utf-8... downstream apps will expect iso8859-1 (if they were receiving the old data.) so we probably want to write the data to disk in iso-8859-1 ... but if we do that... then the content of the 'inline' stuff is different too...
so maybe all alphanumeric stuff should be processed as iso8859-1...

I'm not sure... I'm thinking/wondering...

petersilva:
make an 'outputCharset' option? let it be either utf-8 or iso8859-1... similarly an 'inputCharset' ? I wonder if these things will vary over time?

andreleblanc11:
Yes, and we can default the values to `utf-8` unless otherwise specified.

andreleblanc11:
I've defaulted the bulletin decoding to `iso-8859-1`. We can always add the input/output charset options in a separate PR.

Next steps:

   Merge this PR 
   Create new PR for new plugin that handles all of Sundew's quirks

petersilva:
the binary messages are NOT iso8859-1, they are binary... the last commit is not doing anything helpful... it is only text messages that are iso8859-1.

petersilva:
ok... the iso8859-1 won't do any harm... but It should really be 'ascii'.

andreleblanc11:
Running the flow tests on U18.04

petersilva:
It's wrong but, I'm going to merge it anyways. and we'll fix it when we get a bulleting with something exotique like an accented French character.

PR #909, opened 2024-01-29
Issue905 Queue Names Getting Clobberred.

### LINKED ISSUE #905: Queue getting changed in spite of queueName setting.
When restarting 3 configs on operations, we noticed that, at random, some configurations had taken up a new queueName that had been defined in its cache.

i.e. the new queue value would look something like this in the cache

    q_$USER_$COMPONENT.$CONFIG.51219975.02233416

instead of 

    q_$USER_$COMPONENT.$CONFIG.$HOSTNAME

When performing an `sr3 show`

    'queueName': 'q_$USER_$COMPONENT.$CONFIG.51219975.02233416',

After restarting, the rogue queue names would get defined and the old correct queue names would be left abandoned.

petersilva:
fixes #905  (I hope...)

* see the issue for detailed description of what each commit does.
overall:

* only write the file if action is start, foreground, or declare.
* only write the file if it doesn't exist.
* when you do write the file write it to a .tmp file and then rename to make it atomic ... avoid race condition.

Stuff I left out:
* don't use flufl.lock... feels like overkill.
* if .tmp file exists when checking... just skip it, don't try again...


petersilva:
basic declare/cleanup test is failing because there is an API change.  The purpose of that test failing is to act as a canary to detect api changes... so yay! works as designed.  The fix is an update to the example, which is already in the PR, as well as documentation in UPGRADING.rst to indicate the user-level change.

sarracenia/config.py
```diff
@@ -1750,10 +1747,21 @@ def _resolveQueueName(self,component,cfg):
...
 
-            if (self.queueName is not None) and (not hasattr(self,'no') or (self.no < 2)):
-                f = open(queuefile, 'w')
-                f.write(self.queueName)
-                f.close()
+            # first make sure directory exists.
+            if not os.path.isdir(os.path.dirname(queuefile)):
+                pathlib.Path(os.path.dirname(queuefile)).mkdir(parents=True, exist_ok=True)
+
+            if not os.path.isfile(queuefile) and (self.queueName is not None):
```
andreleblanc11, line 1754: Is this what we want? Only write when the `queuefile` doesn't exist?
What if the `queuefile` is corrupt? It wouldn't get rewritten?

petersilva: if the queuefile is corrupt... something weird is going on... you get rid of it with 'cleanup'.   or you can stop the flow and manually delete just the file, but if the relationship with the queue is gone... state is kind of pooched... so cleanup makes sense.

andreleblanc11 (approved):

PR #911, opened 2024-01-31
fix #910 add a further heuristic for detecting v02 messages

### LINKED ISSUE #910: sr3 can't determine message format when content-type is not set
We receive v02 format messages from non-Sarracenia sources, and they don't set the content-type in the message. This is incorrect, but we want to try to deal with it.

We expect [content-type to be `text/plain` for v02 messages](https://github.com/MetPX/sarracenia/blob/df51bdf85d6ab562aa2a857953870e301a598ed4/sarracenia/postformat/v02.py#L23), so messages with no content type don't get decoded when this code executes:

https://github.com/MetPX/sarracenia/blob/df51bdf85d6ab562aa2a857953870e301a598ed4/sarracenia/moth/amqp.py#L124-L132

In an email, @petersilva said:

> I can look for ‘{‘ not being in the first few characters as a definite positive heuristic of v02.
> Another thing might be to establish a “default” if none of the guesses work…
> We could debate some other heuristics.

petersilva:
fix for #910.

I think this is similar to the logic in v2... but I worry that other future message formats might be caught by this.  

v2 sr_message.py :

```
          if msg.body[0] == '[' :
               self.logger.debug("from_amqplib transitional v03" )
               self.pubtime, self.baseurl, self.relpath, self.headers = json.loads(msg.body)
               self.notice = "%s %s %s" % ( self.pubtime, self.baseurl, self.relpath )
          elif msg.body[0] == '{' :
               self.logger.debug("from_amqplib v03 body: %s" % msg.body)
           ...
          else: # v2.
```

* can't use the methods from v2  in sr3 because there is *wis* format which geo+json and *v03* which is a simpler json, so json is not enough to identify the format.  so the decision is... if it ISN't JSON, then it's v02.

* a future format (for aviation/AMHS/SWIM) will likely be XML based, and thus some adjustments to the detection might be necessary if that post format is added.

I think the check for the brace is helpful, and the body of v02 messages is very simple so it should work for all known current cases.  but it feels like a heuristic.

### DIFF
--- a/sarracenia/postformat/v02.py
+++ b/sarracenia/postformat/v02.py
@@ -31,7 +31,7 @@ class V02(PostFormat):
             return True
 
         # all the other formats are JSON based. only v02 has plain-text body.
-        if not b'{' in payload[0:5]:
+        if not '{' in payload[0:5]:
             return True
 
         # in the v02, we used topic to identify message format. (not reliable for other formats.)
--- a/sarracenia/postformat/v02.py
+++ b/sarracenia/postformat/v02.py
@@ -29,6 +29,11 @@ class V02(PostFormat):
        """
         if content_type == V02.content_type() :
             return True
+
+        # all the other formats are JSON based. only v02 has plain-text body.
+        if not '{' in payload[0:5]:
+            return True
+
         return False
 
     @staticmethod
--- a/sarracenia/postformat/v02.py
+++ b/sarracenia/postformat/v02.py
@@ -31,9 +31,13 @@ class V02(PostFormat):
             return True
 
         # all the other formats are JSON based. only v02 has plain-text body.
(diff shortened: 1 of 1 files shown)


petersilva:
I'm puzzled as I think the test should be b'{' (bytes) but the testing doesn't fail... so it's working?
There is no unit testing of this aspect.

petersilva:
@reidsunderland tested it with DMS messages and it works.

petersilva:
So I tested it by removing the content-type check so it would use the new methods on every v02 message, and it passes flow tests with that.  (the flow tests include v2 messages as well as v03 ones) I think this is good to go.

andreleblanc11 (approved):

PR #921, opened 2024-02-02
correct work description for different running components

petersilva:
@reidsunderland reported seing "downloaded" in winnow output.
it's a bug in the log plugin.


reidsunderland (approved):
looks good to me, assuming it doesn't break the tests :)

PR #924, opened 2024-02-05
changing defaults and documenting the consideration

### LINKED ISSUE #904: Quirk with nodupe_fileAgeMax higher then nodupe_ttl on polls
When using nodupe, there is a problem that comes up when one of the nodes gets removed from a cluster.

The nodupe settings were
`nodupe_ttl 25200` -> 7 hours
`nodupe_fileAgeMax 2592000` -> 30 days

Those are default values being set at 
https://github.com/MetPX/sarracenia/blob/d13f5c7f8169fd95298af4550d567f6ef0c3f941/sarracenia/flow/poll.py#L35-L36

The nodupe cache on node with vip is getting updated everytime the directory is polled, so files are not expiring from the cache there. But they're also not being re-posted which is correct for sarra download purposes (we don't want to re-download) but maybe not correct for cache synchronization purposes?
 
Because there's no messages being posted, the other node's nodupe cache is not being refreshed, and the entries are expiring from those caches after 7 hours. 

The likely solution is either:
1. Adjust `nodupe_fileAgeMax` to a value below `nodupe_ttl` in the defaults
2. Adjust `nodupe_ttl` to have a higher value then `nodupe_fileAgeMax` in the defaults

Having `nodupe_fileAgeMax` be slightly lower (6 hours?) then `nodupe_ttl` would be great. There are not a lot of use cases where we would need to keep duplicates for so long.

petersilva:
closes #904 

documented in the issue itself... just changing poll default and explaining the issue in the documentation.

### DIFF
--- a/docs/source/Explanation/CommandLineGuide.rst
+++ b/docs/source/Explanation/CommandLineGuide.rst
@@ -954,8 +954,10 @@ These options set what files the user wants to be notified for and where
 - **accept    <regexp pattern> [rename=] (must be set)**
 - **reject    <regexp pattern> (optional)**
 - **permDefault     <integer>        (default: 0o400)**
-- **nodupe_fileAgeMax <duration>   (default 30d)**
+- **nodupe_fileAgeMax <duration>   (default: 7h)**
 
+nodupe_fileAgeMax should be less than nodupe_ttl when using duplicate suppression,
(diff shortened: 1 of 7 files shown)


andreleblanc11 (approved):
Looks well documented. The warning message will be good too.

PR #926, opened 2024-02-06
Issue920 fix inflight mtime not working.

### LINKED ISSUE #920: Make watch work with fileAgeMin and inotify.
@reidsunderland is trying to use watch with loggernet data collection software:

* when using the most efficient file noticing method (Inotify) events for files are generated when they are written.
* some programs scribble on files for a while before the files settle down to final content.
* if we "notice" and post the files when the modifications happen, subsequent scribbling to invalidate the posts we would create and confuse things.

There is a setting fileAgeMin ... to reject files which are too new... but with the inotify stuff, that just means they are ignored forever.   We should likely instead place them on worklist.deferred. and at the end of each cycle, the deferred files get copied resolved with the newly gathered at each pass.

this would make it work properly.

petersilva:
closes #920 

there was a regression related to numeric inflight settings (minimum age before processing a file.) I think this fixes it... but the testing so far only confirms that the branch doesn't break anything.  Folks should try the branch in a use case,  where it will be be obvious if it's working properly or not.


petersilva:
the branch passes test... but I'm seeing weirdness in the way watch works... more testing needed.

petersilva:
ok... yeah... confirmed it works properly now. looks good to me.

petersilva:
so... things I am wondering about... if we get an event... and the file is too new.... so the event stays in the event queue...
so reviewed the source code. and I see that , in gather/file.py an OrderedDict() gets new events added to it with the update() method.  The key seems to be the event, and the paths affected.  so this means that if a file gets written again, the update() will have the new event replace the old one.  So it looks like that's OK.

petersilva:
oh yeah... also... when merging... the interim patches break the flow, so it should be squashed into a single commit to avoid messing up future git bisections.

petersilva:
discussed and blessed.

PR #928, opened 2024-02-07
Issue912 retry refilter flag

### LINKED ISSUE #912: Should retry'd transfers reflect updated configuration?
Config before:

```
# oops, there's a typo:
directory /my_datta/
accept .*
```

If the problem in the config is fixed and the process is restarted:

```
directory /my_data/
accept .*
```

When sr3 starts up, failed messages will be recovered from the retry queue and their `new_...` fields will still have the original directory. This isn't wrong, but v2 works differently (I haven't looked at the code to prove it, but that's how I've seen it work). 

We're considering deleting the `new_` fields when recovering messages from retry queues.

If we don't change this, if a directory needs to be changed, the user would need to edit the retry queue file on disk. I assume it would be even harder to edit the contents of a retry queue stored in redis.

petersilva:
closes #912 

builds on the initial issue912 branch by adding a new option:
retry_refilter ... switches between original sr3 processing (just resend using the same message) and v2 method (reprocess the message from scratch.)

I think the vast majority of the time, the client wants the sr3 behaviour... but I'm not sure.  

I have tested this to make sure the flow tests pass, but I don't understand what the interaction with nodupe_ will be.   These retry_refilter messages which were added to a duplicate_suppression cache are likely to be considered duplicates when reprocessed.

I haven't done anything about that yet.  Not sure what should be done (a free pass by defining a "isRetry" field as was done in v2?)

### DIFF
--- a/docs/source/Reference/sr3_options.7.rst
+++ b/docs/source/Reference/sr3_options.7.rst
@@ -1528,6 +1528,19 @@ retryEmptyBeforeExit: <boolean> (default: False)
 Used for sr_insects flow tests. Prevents Sarracenia from exiting while there are messages remaining in the retry queue(s). By default, a post will cleanly exit once it has created and attempted to publish messages for all files in the specified directory. If any messages are not successfully published, they will be saved to disk to retry later. If a post is only run once, as in the flow tests, these messages will never be retried unless retryEmptyBeforeExit is set to True.
 
 
+retry_refilter <boolean> (default: False)
+-----------------------------------------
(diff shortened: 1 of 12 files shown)


petersilva:
This PR is safer than the other two, because, by default, there is no change in behaviour.
To get a change, you need to turn on --retry_refilter ... so regressions of previous sr3 deployments are unlikely.

I'm worried that changing how retry works back to v2 is actively worse in most cases.

PR #933, opened 2024-02-13
Issue918 - Add plugin to rename AM bulletins properly

andreleblanc11:
Couple of things to note for this new plugin.

- The plugin looks for bulletins with 2 characters as its header in the filename. If that is not found, the bulletin remains untouched. If it is found, we add all the necessary data to the filename.
- I added the new directory `flowcb/rename` to accomodate for the function of the plugin.
- I'll be looking at delegating Sundew derived methods, i.e. `getStation`, `getBBB`, `buildHeader`, `getTime`, to a parent class as they might be useful for future plugins. Will do this in a separate PR

I've been testing the plugin for the last couple of hours on dev and it's been working well.

One thing to note is that the AM receiver still doesn't append the timestamp inside the bulletin. I'm waiting to add this with the parent class PR.

### DIFF
--- a/sarracenia/examples/flow/amserver.conf
+++ b/sarracenia/examples/flow/amserver.conf
@@ -1,13 +1,19 @@
+# For message content purposes
+post_topicPrefix v03.post
+post_format v03
 
 callback gather.am
 callback post.message
+
+# Add appropriate data to filename
+callback rename.raw2bulletin
 
 post_broker amqp://tsource@localhost
 post_exchange xs_tsource_am
 post_baseUrl file://
 
 download on
-directory /tmp/andre_am_receiver
+directory /tmp/am_receiver
 accept .*
 sum sha512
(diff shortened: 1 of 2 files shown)


petersilva (approved):
Looks good to go. since no tests will fail, and you are the only one running AM to be able to test it, there is no reason to refuse the PR.  I especially like that you have sample inputs and outputs in comments in the code.  It makes it much more intelligible for whoever has to modify this in a decade.

andreleblanc11:
I added another commit so that erroneous bulletins get better handled.

Previously, sr3 would just post the erroneous bulletins downstream. Sundew will add the file to a PROBLEM directory so that the files don't get ingested by senders.

What I did is I added the message to `worklist.rejected` and I write out the file with `_PROBLEM` at the suffix and with a fresh timestamp. In this case, we'll have clear records of bulletins that have failed.

PR #944, opened 2024-02-16
Issue942 - sarra consumer writing wrong files because 'rename' field should have been deleted.

### LINKED ISSUE #942: When posting a file after downloading, the rename header should be removed from the message
I encountered a problem where a v2 sr_sarra was subscribed to the output of an sr3 sarra.

The sr3 sarra receives messages with a rename header, downloads the file to the correct location (the directory set in the config) with the correct name (from the rename header). The message posted by the sr3 sarra has the correct relPath, *but it still includes the rename header. *

e.g. `relPath = path/to/file.txt` and `rename = file.txt`

The downstream v2 sr_sarra has `mirror True`, so it should download the file into `some_directory/path/to/file.txt` but instead the rename field overrides the relPath and the file gets downloaded to `some_directory/file.txt`.

v2 would delete the rename header after downloading the file:
https://github.com/MetPX/sarracenia/blob/5b063af501cc843e9623a40da1fc782134c79f0e/sarra/sr_subscribe.py#L1409-L1415

(the answer to the FIXME in that comment is *yes*, v2 updates relPath correctly)

Slightly related to this other rename change: https://github.com/MetPX/sarracenia/pull/936

That changed the behaviour of rename when download is False. We need to make a change so that when download is True, the rename field gets deleted from the message posted by the sarra.

### LINKED ISSUE #943: processing of rename header is wrong...
before 3.0.51 winnow & shovel would corrupt messages going through them with "rename" fields.
had to add this:  b1c9ee9454619251be1b6670af4b3986a1e21532

which made rename processing in the winnow and shovel work correctly, but later on, when the rename is processed, the header needs to be deleted.

petersilva:
closes #942 
closes #943 

This adds deletion of the 'rename' message field after it is applied.
For similar reasons, retrievePath needs to be deleted once the download is done.

### DIFF
--- a/sarracenia/flow/__init__.py
+++ b/sarracenia/flow/__init__.py
@@ -805,10 +805,16 @@ class Flow:
             u = sarracenia.baseUrlParse(msg['baseUrl'])
             relPath = u.path[1:] + '/' + relPath
 
-        # FIXME... why the % ? why not just assign it to copy the value?
         if self.o.download and 'rename' in msg: 
+            # FIXME... why the % ? why not just assign it to copy the value?
             relPath = '%s' % msg['rename']
+
+            # after download we dont propagate renaming... once used, get rid of it
             del msg['rename']
+            # FIXME: worry about publishing after a rename.
+            # the relpath should be replaced by rename value for downstream
+            # because the file was written to rename.
+            # not sure if this happens or not.
 
 
         token = relPath.split('/')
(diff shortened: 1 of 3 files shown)


petersilva:
@reidsunderland if you can confirm the fix... that would be good!

mshak2 (approved):
Reviewed with Reid and confirmed to be working.
Logged renames in posted messages.

PR #952, opened 2024-02-29
Move AM renamer to gather. Allows us to properly use sundew variable expansions.

andreleblanc11:
This should've probably been tested before v3.00.52rc2 but oh well.

I started toying around with [Sundew compatible substitutions](https://metpx.github.io/sarracenia/Reference/sr3_options.7.html#sundew-compatible-substituions) and ran into some problems.

The renamer was getting called via `after_accept` which is wrong since we're doing Sundew substitutions BEFORE calling `after_accept`. With this PR, by moving the renamer before the filtering, we essentially avoid all of the limbo that is required to call the sundew substitutions after the filtering. It makes the whole process easier and makes more sens.

### DIFF
--- a/sarracenia/flowcb/gather/am.py
+++ b/sarracenia/flowcb/gather/am.py
@@ -65,6 +65,7 @@ from base64 import b64encode
 import urllib.parse
 import sarracenia
 from sarracenia.bulletin import Bulletin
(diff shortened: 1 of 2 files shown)


petersilva (approved):
looks good!

Issue #649, opened 2023-04-21
sr3 status display size of queue.


petersilva:
when people do an sr3 status, they want to see the size of the upstream queue.  AMQP has no API to be able to do this.  I think you can do something with MQTT, but in any event, is not clear how to implement.

option 1: Query/Response API with messages.

publish a message to the upstream broker that formulates a query of some kind, and it publishes a message in response.

option 2: inline metrics messages:
Mixed in with the normal publication messages would be a message from the broker that includes the size of the upstream queue.  subscriber could note it down for local use.

could be in a separate topic tree, so binding could be optional.

both options require flow on the broker side that builds custom messages.

petersilva:
users have requested multiple times to see the queue size that, on rabbitmq isn't available to users without giving them additional permissions to use the management GUI.

petersilva:
It turns out the current implementation breaks things on high traffic queues. See #953

reidsunderland:
This is fixed and working for AMQP. I guess it can be left open to eventually try to implement it for other protocols?

PR #954, opened 2024-03-04
Issue953 PRECONDITION_FAILED unknown delivery tag

petersilva:
The main patch is to comment out the queue declare invocation that implemented the fix for #649.  I do not undestand why declaring a queue on one channel confuses message identifiers on another channel, but so I do not know how to address it.  So I re-opened the old issue.

related stuff I found:

* when some corrupt messages are received, they were being dropped without being acknowledged, added acknowledgement.

* as one attempt to address it, tried specifying channel numbers when they are created.  Made no perceptible difference.

unrelated stuff I found:

* if you do *sr3 declare componenet/config* and there is no admin.conf it bombs with an error about admin.conf url existing.  patch for that.

### DIFF
--- a/sarracenia/moth/amqp.py
+++ b/sarracenia/moth/amqp.py
@@ -119,6 +119,7 @@ class AMQP(Moth):
                     logger.error(
                         'ignoring message. UTF8 encoding expected. raw message received: %s' % ex)
                     logger.debug('Exception details: ', exc_info=True)
+                    self.channel.basic_ack( raw_msg.delivery_info['delivery_tag'])
                     return None
 
             if 'content_type' in raw_msg.properties:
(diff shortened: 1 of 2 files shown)


andreleblanc11 (approved):
I'm approving. Not sure if we want to keep the channel numbers long term, but it seems harmless as of now (and likely should be long term too).

reidsunderland:
Is it possible for a message to get corrupted by network issues between the broker and client? If yes, shouldn't we nack or reject the message in that case?

If this is only handling cases where the message really contains incorrect content then acking it makes sense.

petersilva:
I was reading a bit about ack. and nack... Either way they don't resend... To resend we just leave the code as it was before.
The option nack adds is to send the message to a dead letter exchange if we define it.  But I didn't even see how to do that....  I read that nack is a rabbitmq extension and the amqp library doesn't even have the call.  But I think the likelihood of corruption is nil because of tcpip...

PR #957, opened 2024-03-06
Change raise_signal to os.kill for #948

### LINKED ISSUE #948: raise_signal requires Python >= 3.8
We support Python >= 3.6. On the older versions, this causes an exception.

```
./moth/amqp.py:407:            signal.raise_signal(signal.SIGINT)
./moth/amqp.py:481:            signal.raise_signal(signal.SIGINT)
./moth/mqtt.py:420:            signal.raise_signal(signal.SIGINT)
./moth/mqtt.py:501:            signal.raise_signal(signal.SIGINT)
```

```
2024-02-23 14:46:56,771 [WARNING] sarracenia.moth.amqp getNewMessage lost connection to broker
2024-02-23 14:46:57,781 [WARNING] sarracenia.moth.amqp getNewMessage failed q_feeder_sarra.get-cmems-oceanobs.88019679.07467901: module 'signal' has no attribute 'raise_signal'
```

reidsunderland:
Fixes #948, `raise_signal` method doesn't exist on older versions of Python that we still support.

### DIFF
--- a/sarracenia/moth/amqp.py
+++ b/sarracenia/moth/amqp.py
@@ -34,6 +34,7 @@ import sarracenia
 from sarracenia.postformat import PostFormat
 from sarracenia.moth import Moth
 import signal
+import os
 
 import time
 from urllib.parse import unquote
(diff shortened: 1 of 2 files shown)


petersilva (approved):
after some reading about os.kill and signal.raise calls in various python versions, it looks like using os.kill as proposed in the PR is at least a reasonable approach... it might even be the best approach. ;-)

Issue #951, opened 2024-02-28
Implement processing of `post_baseUrl` field in a message


reidsunderland:
I ran into a case where setting `msg['baseUrl']` and `msg['new_baseUrl']` from inside a poll plugin doesn't do anything. It will always use the configured `post_baseUrl` option.

I discussed with @petersilva and @andreleblanc11 and we thought the best way to force set a baseUrl is to set `msg['post_baseUrl']`. A message field that matches the name of a config option will override other ways of setting the `baseUrl`.

I'm creating this issue so I don't forget to implement it.

petersilva:
This aligns with a  similar thing with msg['topic'] if the 'topic' header is set, it acts as an override for what would be calculated from the options.

reidsunderland:
In my poll plugin, I wasn't able to set the `baseUrl` in the messages I was creating.

I used [`sarracenia.Message.fromFileInfo`](https://github.com/MetPX/sarracenia/blob/47dd8dc3325f96d49befc2609556e0d1615cf69d/sarracenia/__init__.py#L572), which would set baseUrl using updatePaths, which sets baseUrl from options.post_baseUrl. My plugin needs to be able to post multiple different baseUrls, so I couldn't just use a pre-defined post_baseUrl. 

I assumed that I could just change `msg['baseUrl']` and/or `msg['new_baseUrl']` after creating the message, but this didn't work. 

The reason it doesn't work is because the flow algorithm calls updatePaths again, which would replace whatever I had set `msg['new_baseUrl']` to back to options.post_baseUrl. Then this code in the flow algorithm:

https://github.com/MetPX/sarracenia/blob/47dd8dc3325f96d49befc2609556e0d1615cf69d/sarracenia/flow/__init__.py#L517-L522

would override my correct `msg['baseUrl']` with the incorrect `msg['new_baseUrl']` (set to options.post_baseUrl by updatePaths).

I worked around this by adding an after_accept method to my poll plugin that would delete the `new_baseUrl` field before the above code. 

```python
def after_accept(self, worklist):
    for msg in worklist.incoming:
        del msg['new_baseUrl']
```

But I discussed this with @petersilva  and @andreleblanc11 and we decided that it would be better that if someone sets a field in a message that matches the name of an option (like post_baseUrl), the option from the message would override other possible ways of setting the field.

I implemented this in updatePaths:
https://github.com/MetPX/sarracenia/commit/640a1f1f7eab59c06862945a20d0b3bc80ba7a31

PR #959, opened 2024-03-07
Add plugin to poll the Copernicus Marine Data Store service

reidsunderland:
This is the plugin I was writing that lead to #951.

It posts https URLs that can be downloaded without any special code or authentication needed.

The S3 part is based on https://github.com/MetPX/sarracenia/blob/development/sarracenia/flowcb/poll/s3bucket.py, but list_objects only returns a maximum of 1000 items so I had to use a "paginator". Should that also be implemented in s3bucket.py?

```python
                    # s3.list_objects returns a maximum of 1000 items, need to use paginator instead
                    operation = 'list_objects_v2' if 'list_objects_v2' in dir(s3) else 'list_objects'
                    paginator = s3.get_paginator(operation)
                    page_iterator = paginator.paginate(Bucket=bucket_prefix['bucket'], Prefix=bucket_prefix['prefix'])
                    for page in page_iterator:
                        if 'Contents' in page:
                            for obj in page['Contents']:
                                if bucket_prefix['bucket'] not in objects_by_endpoint_bucket[endpoint]:
                                    objects_by_endpoint_bucket[endpoint][bucket_prefix['bucket']] = []
                                objects_by_endpoint_bucket[endpoint][bucket_prefix['bucket']].append(obj)
```

### DIFF
--- a/sarracenia/flowcb/poll/copernicus_marine_s3.py
+++ b/sarracenia/flowcb/poll/copernicus_marine_s3.py
@@ -1,0 +1,236 @@
+"""
+Polls the Copernicus Marine Data Store STAC API and S3 buckets.
+----------------------------------------------------------------
+
+Based on https://github.com/MetPX/sarracenia/blob/development/sarracenia/flowcb/poll/s3bucket.py
+
+This was developed because the software provided by Copernicus, the Copernicus Marine Toolbox requires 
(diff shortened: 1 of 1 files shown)


petersilva (approved):
great stuff!  it cannot beak anything existing, since it is brand new.
good demo for others...

PR #965, opened 2024-03-10
Remove threading from AMQP Consumer code

reidsunderland:
I finally found some time to remove the threading code and call drain_events from getNewMessage instead.


petersilva (approved):
very cool!

PR #979, opened 2024-03-18
Reject float types for Julian day value in getTime

andreleblanc11:
When Sundew code was ported to sarracenia V2, some analysts added some logic to enable float values for Julian days.

Sundew doesn't allow float values for its Julian days and simply rejects the wrong bulletins.

This will mimic what Sundew already does.

### DIFF
--- a/sarracenia/bulletin.py
+++ b/sarracenia/bulletin.py
@@ -224,11 +224,10 @@ class Bulletin:
 
             if len(parts) < 4: return None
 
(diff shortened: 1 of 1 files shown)


petersilva:
Discussed and agreed:

* occurred for one bulletin in several weeks of auditing vs. sundew.
* code in v2 was inexplicable.
* floats as julian days should never be correct.
* Want to ensure such a message is reported as an error for folllowup.

current fix seems to do that.

PR #984, opened 2024-03-19
Scheduled poll refactor done.

### LINKED ISSUE #974: Redesign poll to work like scheduled flow? (Because polls do not consume from queues until after sleeping for a long time)

3. Sleep
4. Repeat

image

--- 

petersilva:
closes #974 

So this implements the refactoring for #974 .

It worries me, in that I'm not sure what side effects the change will have.  @reidsunderland do you have an easy way to test if the queueing thing is fixed by the branch?  The branch is: https://github.com/MetPX/sarracenia/tree/scheduled_poll
The testing in the flow tests is pretty limited, just a straight vanilla SFTP with a short sleep, a VIP and failover.

Things I am wondering about:

* It doesn't do a poll on startup, currently.  it waits for one interval before the first poll.  easy to fix... should we?

* So now in poll, the loop towards the end of the run() routine is only used to sleep for 1 second at a time... and the real waiting 
  is done in the scheduled.poll plugin instead.

  In watch... still using long sleeps in that loop at the end...  Is this an opportunity to simplify sr_watch also... so it also uses the 
  same methods?  Is there an advantage to doing it that way?
 
* I'm wondering if there is a way to have watch use a shared duplicate suppression cache the same way as poll does, populating itself with the output exchange binding queue.  Same benefit as poll, run on two machines with a vip and failover.

* wondering if this change might be an opportunity to simplify sarracenia.flow/__init__.py run() routine.  It feels too complicated.

### DIFF
--- a/.github/workflows/flow.yml
+++ b/.github/workflows/flow.yml
@@ -1,4 +1,4 @@
-name: sr_insects Flow Tests.
+name: Flow Tests / AMQP
 
 on:
   pull_request:
@@ -36,7 +36,7 @@ jobs:
 
     runs-on: ${{ matrix.osver }}
   
-    name: ${{ matrix.which_test }} test on ${{ matrix.osver }}
+    name: ${{ matrix.which_test }} on ${{ matrix.osver }}
     timeout-minutes: 40
     
     steps:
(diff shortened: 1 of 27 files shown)


petersilva:
replaced with new PR because the rebases clutter things up.
want a squash for this because the individual commits don't pass flow tests.

PR #987, opened 2024-03-20
Issue986 - stop converting *directory* to *path* in poll.

### LINKED ISSUE #986: compatibility conversion of directory to path for polls is misleading
in a v2 poll (and only a poll), when a *directory* argument was provided, it referred to a pattern to match on the remote. Similarly, *get*, and *accept* are applied on the server, as globbing patterns.   

This is different from all other components, where *directory* refers to where your are downloading to, the destination to write to, and the accept/reject would use regular expressions, rather than globbing.

To improve consistency, we convert:  

* v2: directory -> v3: path

* v2: get -> v3: path

* v2: accept ... does not really match v3 accept.

v3 accept is the same in all components.

When converting from v2 to sr3, it is important/useful to convert *directory* verbs to *path*.

However, once at sr3... one can set *download True* and there may be polls that download files, so getting "directory" back to it's meaning in the other components would be a good thing.

petersilva:
closes #986 

A year or two ago, in one of the re-factors of poll, *directory* stopped being used,
replaced by path.  but then since v2 uses directory, an automated conversion of *directory* to *path* was added.

That made *directory* essentially the same as a path, but only in the *poll* component.  Path is an upstream thing you are querying in all components.
directory (everywhere except poll) is a directive about what directory to put
files in when transferring... the destination directory.

This PR removes automated conversion of directory to path in poll component,
so that:

* *directory* means the same thing in all components now (and not something different in poll vs. other components.)
* *path* means the same thing in all components (refers to the thing being queried or listed as a source of files.)
* if we start having polls that download, *directory* will be available.
* Add a message notifying when a directory message does not accomplish anything.
* Also... the poll documentation still referred to directory, rather than path...  This change in behaviour is about a year or two old?  I guess it wasn't changed because they were synonyms... but now that they aren't, it is important to align.

Note that sr3 convert still converts v2 *directory* statements into v3 *path* statements... because ... umm that's correct... v2 directories were ... different from v3 ones.

### DIFF
--- a/docs/source/Explanation/CommandLineGuide.rst
+++ b/docs/source/Explanation/CommandLineGuide.rst
@@ -934,8 +934,8 @@ POLLING
 
 Polling is doing the same job as a post, except for files on a remote server.
 In the case of a poll, the post will have its url built from the *pollUrl* 
-option, with the product's path (*directory*/"matched file").  There is one 
-post per file.  The file's size is taken from the directory "ls"... but its 
+option, with the product's path (*path*/"matched file").  There is one 
+post per file. The file's size is taken from the directory "ls"... but its 
 checksum cannot be determined, so the default identity method is "cod", asking
 clients to calculate the identity Checksum On Download.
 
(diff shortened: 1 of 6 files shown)


petersilva:
There is also poll specific rename feature as part of *filename* ... but filename works completely differently from other instances of filename.  This should be removed.

petersilva:
hold off... want to remove *filename rename* support from sr3.

petersilva:
ok, it's removed... good to go.

petersilva:
btw... I run all the flow tests before submitting a PR.  the tests failing is disconcerting... but

andreleblanc11 (approved):

PR #988, opened 2024-03-20
NASA PODAAC checksum change

reidsunderland:
fix a possible problem found by @andreleblanc11, where a PODAAC sourc…e doesn't have an md5sum when first posted, then the md5sum becomes available later, it will re-post and re-download


petersilva:
I don't know why all checks have failed, but it has nothing to do with this change.

reidsunderland:
http://[ppa.launchpad.net](http://ppa.launchpad.net/ssc-hpc-chp-spc/metpx/ubuntu/dists/focal/InRelease)/ssc-hpc-chp-spc/metpx/ubuntu/dists/focal/InRelease 

seems to be down, causing the failed checks

PR #990, opened 2024-03-21
Issue984 flow run refactor (branches from the other polling PR)

petersilva:
poll can now run download like all the other components... kind of a happy side effect of just refactoring the code to make it easier to understand.

builds on https://github.com/MetPX/sarracenia/pull/985

The run() routine is 288 lines ... it is too long to think about.  Refactored code chunks into separate routines
just to shorten the code in run()  The code didn't change much at all, it is just re-arranged.
Moved:
* took all the message reformatting, callback invocation and worklist acknowledgements around the do() routing and created
  a work() function as a wrapper for it.
* similarly moved the callback invocation and ack and worklist changes and from run() to the end of the filter() function.
* did similar thing for the post() routine moving stuff from run() into post.
* made a _run_vip_update routine outside of run() but called from it.
  * now could move *on_start* callback invocation so that run_vip_update() is called from a single place.

run() is now 159 lines.  Changes in behaviour:

* when poll has the vip it calls do(), so similar to all other components it can do a download in a single component if desired.
* now novipfile management done in one place, also calls to housekeeping processing reduced from 2 to 1.

This PR includes the PR it is based on since that one isn't merged yet... will see how it goes if the other one gets merged first.

### DIFF
--- a/docs/source/Explanation/SarraPluginDev.rst
+++ b/docs/source/Explanation/SarraPluginDev.rst
@@ -585,12 +585,17 @@ for detailed information about call signatures and return values, etc...
 |                     | permanent name.                                    |
 |                     |                                                    |
 |                     | return the new name for the downloaded/sent file.  |
+|                     |                                                    |
 +---------------------+----------------------------------------------------+
 | download(self,msg)  | replace built-in downloader return true on success |
 |                     | takes message as argument.                         |
 +---------------------+----------------------------------------------------+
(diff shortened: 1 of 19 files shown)


petersilva:
Probably best to merge the other one, then start this over... because I want a single squash merge for the other one, but in this one, every change is ok (none break the tests.) so it's ok to keep them separate.

petersilva:
will close for now.

PR #993, opened 2024-03-22
Clarify a bit more how each component works in the documentation

andreleblanc11:
I added a bit more in Table 2. in the Concepts section of the documentation.

French docs were also modified accordingly.


petersilva (approved):
approved.  I added some more stuff in the other columns.
maybe a third person should review.

PR #995, opened 2024-03-23
Issue994 unit tests broken and failing alot (but weirdly not always.)

### LINKED ISSUE #994: unit tests broken.
They seem to work sometimes,and fail sometimes, which is very confusing so had a look, and there are a number of failure modes.  first mode:

```

FAILED sarracenia/flowcb/nodupe/disk_test.py::test_after_accept__InFlight - AttributeError: 'Options' object has no attribute 'fileAgeMax'
=============================== 4 failed, 161 passed, 6 skipped, 1 warning in 0.99s ================================


```

why is fileAgeMax missing and/or needed? Why would this ever pass? hard to understand.

petersilva:
fix #994 

I'm puzzled about the unit tests.  I thought they were working occasionally. and looking
at the history of results they seem to succeed once in a while.  I started running them manually
to see why they were failing often, and found a bunch of problems.  Fixed what I could figure out easily.

The most elaborate assertions were failing, and I think it is because they are asserting things that just aren't true... the failed tests don't reflect anything wrong with the classes...

### DIFF
--- a/tests/sarracenia/flowcb/nodupe/disk_test.py
+++ b/tests/sarracenia/flowcb/nodupe/disk_test.py
@@ -23,6 +23,9 @@ class Options:
         self.config = "foobar.conf"
         self.pid_filename = "/tmp/sarracenia/diskqueue_test/pid_filename"
         self.housekeeping = float(39)
+        self.fileAgeMin = 0
+        self.fileAgeMax = 0
+
     def add_option(self, option, type, default = None):
         if not hasattr(self, option):
             setattr(self, option, default)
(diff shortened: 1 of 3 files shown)


petersilva:
@gcglinton I thought I'd invite you to review in case the tests I had to disable are obvious to you. If they aren't feel free to ignore. Thanks!

gcglinton:
The change to the message generation for these tests is a good one

I wrote the asserts/tests based on the code I could decipher, and assumptions I had made (like apparently using v2 message formats?). If those assumptions were wrong, then obviously the tests are also wrong.

Coverage doesn't seem to be affected, so I guess all is well.

Also, you've already merged it!

PR #998, opened 2024-03-24
No download message when not downloading

### LINKED ISSUE #976: flowcb/log prints downloaded message even when download is off

petersilva:
close #976


petersilva:
discussed and blessed.

PR #999, opened 2024-03-24
Issue992 - re-parse values using correct type for add_options from flow_callbacks.

### LINKED ISSUE #992: type of options added with add_option wrong...
if you have an option that isn't built into the main clases, then the option will be parse as a list
of strings (default type) rather than the type declared in the plugin via add_option.

example:
```

scheduled_interval 10

```

in a poll,  results in a display in sr3 show poll/xxx

```

'scheduled_interval': '10',

```

note the quotes around 10, so it is a string, if supplied twice it will be a list with the two values.
in the flowcb/scheduled callback, this option is declared as a duration, it should show up as
a number of seconds, in the sr3 display.

Similar thing happens with anything that isn't a list.

petersilva:
close #992 

repackaging of stuff from work with @mshak2 on wis2... there were enough patches that it made sense to deal with this separately. The problem is explained in #992.   when options were declared in callbacks, and their type was not list (the default) they would be parse as lists anyways, because parsing encountered the type declarations after already having parsed the values.

At the end of option processing... there is already logic to re-examine "undeclared" variables. The existing logic removes variables that were declared in callbacks from the undeclared lists so that we don't generate bogus "undeclared" messages. These patches add, that when you find a declared option... re-parse the value to have it match the type declared for the option.

The string parsing for set values, was done weirdly with code in two places, Since we now needed it in a third place, the original two were consolidated into a set value parsing routine, and call it from all three places.

### DIFF
--- a/sarracenia/config.py
+++ b/sarracenia/config.py
@@ -322,9 +322,9 @@ r"""
 
 
 def isTrue(S):
-    s = S.lower()
-    if s == 'true' or s == 'yes' or s == 'on' or s == '1': return True
-    return False
+    if type(S) is list:
+        S = S[-1]
+    return S.lower() in ['true', 'yes', 'on', '1']
 
 
 def get_package_lib_dir():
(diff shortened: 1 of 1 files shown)


petersilva:
should be done with a squash, since the first patch breaks things...

petersilva:
discussed and blessed.

Issue #982, opened 2024-03-18
Cleanup doesn't remove state files when `statehost` is enabled


reidsunderland:
When statehost is off, cleanup will delete the qname file from `~/.cache/sr3/$component/$config/*qname` and other state files.

```
2024-03-18 21:49:29,306 1924819 [INFO] sarracenia.flowcb.log on_cleanup hello
removing state file: /net/local/home/sunderlandr/.cache/sr3/sarra/get_copernicus_marine/sarra.get_copernicus_marine.root.qname
```

With statehost on, it doesn't work.

```
2024-03-18 21:53:31,395 1925091 [INFO] sarracenia.flowcb.log on_cleanup hello

```

I added a bunch of prints to the relevant section of sr.py and it looks like it's not using the correct directory:

```
2024-03-18 22:00:41,480 1926129 [INFO] sarracenia.flowcb.log on_cleanup hello
f = sarra/get_copernicus_marine
cache_dir = /net/local/home/sunderlandr/.cache/sr3/sarra/get_copernicus_marine
cache_dir is a dir

```

```python
        for f in self.filtered_configurations:
            if self.please_stop:
                break
            print(f"f = {f}")
            cache_dir = self.user_cache_dir + os.sep + f.replace('/', os.sep)
            print(f"cache_dir = {cache_dir}")
            if os.path.isdir(cache_dir):
                print(f"cache_dir is a dir")
                for state_file in os.listdir(cache_dir):
                    print(f"working on {state_file}")
                    if self.please_stop:
                        break
                    if state_file[0] == '.':
                        continue

                    if state_file in [ 'disabled' ]:
                        continue

                    asf = cache_dir + os.sep + state_file
                    print(f"RS {asf} {cache_dir} {os.sep} {state_file}")
                    if self.options.dry_run:
                        print('removing state file (dry run): %s' % asf)
                    else:
                        print('removing state file: %s' % asf)
                        os.unlink(asf)
```

petersilva:
the fix in March was no good.... the fix for this will only come out in 3.0.54.

PR #1003, opened 2024-03-27
#982 statehost config option should be used to see pick state directory

petersilva:
Prior to this change, when a statehost directory existed for a given configuration,
the values in the statehost directory would override those in the non-statehost one.
e.g.:

The ~/.cache/sr3/sarra/download_f20/  directory contained correct .pid files for the
three instances that were running, but sr status indicated it that the instances were
orphans because, after reading that state directory, it then read ~/.cache/sr3/*hostname*/sarra/download_f20, which had no .pid files.

When reading the second directory, it was resetting the known instances to empry,
and then, because the directory was empy, never adding any values back.

This change causes the config settings to be used to select the state directory, based
on the statehost setting.

### DIFF
--- a/docs/source/Explanation/CommandLineGuide.rst
+++ b/docs/source/Explanation/CommandLineGuide.rst
@@ -939,6 +939,18 @@ post per file. The file's size is taken from the directory "ls"... but its
 checksum cannot be determined, so the default identity method is "cod", asking
 clients to calculate the identity Checksum On Download.
 
+To set when to poll, use the *scheduled_interval* or *scheduled_hour* and *scheduled_minute*
+settings. for example::
+
+   scheduled_interval 30m
+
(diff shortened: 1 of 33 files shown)


andreleblanc11 (approved):
Changes look good based on explanation provided. Approving.

petersilva:
This branch forked from development several weeks ago, and there have been several breaking changes since then. I'm running some tests to see if rebasing helps. will see.

petersilva:
so rebased from development trying to get all the tests to pass, and noticed that python3.6 cannot handle the type hints that I had provided on a later version.  So removed them entirely. We should see more tests pass as a result.

petersilva:
have merged these changes into #1010 to get it's tests to pass... this pull is perhaps no longer needed?

Issue #997, opened 2024-03-24
 Possible race condition when reading retry file during housekeeping intervals


andreleblanc11:
I saw over the weekend that the AM server crashed due to the following error.

```
Traceback (most recent call last):
  File "/home/sarra/.local/lib/python3.10/site-packages/sarracenia/instance.py", line 248, in <module>
    i.start()
  File "/home/sarra/.local/lib/python3.10/site-packages/sarracenia/instance.py", line 239, in start
    self.running_instance.run()
  File "/home/sarra/.local/lib/python3.10/site-packages/sarracenia/flow/__init__.py", line 586, in run
    self._runCallbackMetrics()
  File "/home/sarra/.local/lib/python3.10/site-packages/sarracenia/flow/__init__.py", line 319, in _runCallbackMetrics
    self.metrics[module_name] = p()
  File "/home/sarra/.local/lib/python3.10/site-packages/sarracenia/flowcb/retry.py", line 166, in metricsReport
    return {'msgs_in_download_retry': len(self.download_retry), 'msgs_in_post_retry': len(self.post_retry)}
ValueError: __len__() should return >= 0
```

What I was able to find in the logs is that the retry file will get opened during a flow algorithm passage by the process that will soon die.

```
# Process 1
2024-03-24 00:57:12,029 [DEBUG] sarracenia.diskqueue msg_get_from_file DEBUG /home/sarra/.cache/sr3/flow/amserver/diskqueue
_post_retry_001 open read
2024-03-24 00:57:12,030 [INFO] sarracenia.flowcb.log after_work flowed ok: /datalocal/am/bulletins/20240324/IS/CWVH/01/ISAB
64_CWVH_240100___00001
```

And just before, another process will open the retry file for read also

```
# Process 2
2024-03-24 00:57:11,470 [DEBUG] sarracenia.diskqueue msg_get_from_file DEBUG /home/sarra/.cache/sr3/flow/amserver/diskqueue_post_retry_001.new open read
```

What I'm thinking is that there is some kind a race condition happening where the process that dies accesses the retry file just after the first one, which causes the crash.

I think this is normally avoided with multiple instances by denominating a retry file for each instance #. But with AM server, we fork from the same process so they all use the same retry file in consequence.

FYI, this only crashed the instance affected and it was able recover one second after. However, we got unlucky and lost 7/8 bulletins due to that crash as there were lots of files being transferred on the socket at the time of the crash.

petersilva:
[...]

PR #1005, opened 2024-03-28
Issue997 - Eliminate retry file race condition and make AM work more like a sarracenia flow

andreleblanc11:
Solves #997

### DIFF
--- a/sarracenia/flowcb/gather/am.py
+++ b/sarracenia/flowcb/gather/am.py
@@ -99,6 +99,7 @@ class Am(FlowCB):
         self.minnum = 00000
         self.maxnum = 99999 
         self.remoteHost = None
+        self.timeout = 0.1
 
         # Initialise socket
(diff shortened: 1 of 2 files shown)


petersilva (approved):
Looks good!

gcglinton:
This broke the retry unit tests, because the driver selection got moved from `__init__` into `on_start`, so now none of the test logic works.

In all the failing tests of `retry_test.py`, it should be enough to call `retry_disk.on_start()` or `retry_redis.on_start()` right after those retry objects (`retry_disk`, `retry_redis`) are instantiated. Like right after lines 75, 82, 109, 115, etc..

petersilva:
yes the I found the same issue, there is a a commit 25df9852aa8603c0725d4afdd90a9c79debd3b76 
It is in the other PR ( https://github.com/MetPX/sarracenia/pull/1003 )

PR #1009, opened 2024-04-04
Transfer documentation and a fix for all tests...

petersilva:
after working with @gcglinton, @reidsunderland, @andreleblanc11 on the new s3 transfer class, I added some documentation about how to write transfer classes.  The principle unfathomable part is what the ls() entry point returns, which is quite odd.  It is the right thing to do, but it requires explanation, included in the commits.

Also noticed that after AM protocol work with @andreleblanc11 the cleanup() function of the retry class was crashing many tests.  Have to add some initialization to get the cleanup to work.

### DIFF
--- a/sarracenia/flowcb/poll/__init__.py
+++ b/sarracenia/flowcb/poll/__init__.py
@@ -379,6 +379,7 @@ class Poll(FlowCB):
 
 
         # assert at this point we have an sftp_obj...
+        # filter out files we don't have the necessary permissions for.
         if 'sftp_obj' in locals() and ((sftp_obj.st_mode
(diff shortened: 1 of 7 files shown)


reidsunderland (approved):
Looks good to me. 

There's a typo in `transfer/__init__.py`: `a seqeence of bytes... will be parsed as an html page.`

gcglinton:
Seem like adding the type hints broke unit testing on Ubuntu 20.04, as it hasn't worked there since this was merged.

petersilva:
yeah... trying to figure out why the tests broke because we added type hints.  that routine is weird... it's polymorphic... returning different types depending on the protocol... but that is really what is needed... it's not wrong.

PR #1010, opened 2024-04-05
Native S3 transfer driver

gcglinton:
Per many discussions, voila, a native S3-compatible transfer driver.

Built on the backs of the good work done by @reidsunderland, and @tysonkaufmann, and a collaborative programming session with @petersilva, @reidsunderland, and @andreleblanc11.

It does all the things it's supposed to do, at least as far as I can tell. 

I ran a very simple set of configs, and there was no errors:
 - A `poll` to find files in an S3 bucket
 - A `subscribe` to download and save them, and then post a new message for..
 - A `sender` to pick up those local files and re-upload them to a different bucket

It's only been tried against AWS S3 proper, so it may or may not work with MinIO, or others, but theoretically, everything is in place to do that.

More testing will likely be required to ensure it's working exactly as intended, but this might be a good start.

There's also some unit tests, because they're tasty, and fun, and I like writing them (what does that say about me?). It also meant figuring how to mock `boto3` in order to keep from having to spin up a dependent service, which turned out to be super easy using [moto](https://github.com/getmoto/moto/tree/master).

The only issue left is that these new tests fail on Ubuntu 20.04, but they work fine on 22.04. It seems like the Ubuntu provided `python3-openssl` is broken, or incompatible with boto3. Following a combination of answers from [this](https://stackoverflow.com/questions/73830524) SO question led me to fixing it by forcing pip to upgrade the pyOpenSSL package: `sudo pip3 install pyopenssl --upgrade`

Fixing the packages in the runner before doing the unit tests is the best way to go about it, but that would also mean documenting that fix somewhere so users can actually make it work on Ubuntu 20.04, and potentially other distributions/versions as well.

It might be also be possible to conditionally [skip](https://docs.pytest.org/en/stable/how-to/skipping.html) the s3 tests, based on Python version, or possibly even OS, but from what I've been able to find, that's not very reliable, and very much doesn't solve the issue.

### DIFF
--- a/docs/source/Explanation/CommandLineGuide.rst
+++ b/docs/source/Explanation/CommandLineGuide.rst
@@ -939,6 +939,18 @@ post per file. The file's size is taken from the directory "ls"... but its
 checksum cannot be determined, so the default identity method is "cod", asking
 clients to calculate the identity Checksum On Download.
 
+To set when to poll, use the *scheduled_interval* or *scheduled_hour* and *scheduled_minute*
+settings. for example::
+
+   scheduled_interval 30m
+
+to poll the remote resources every thirty minutes. Alternatively::
+
+   scheduled_hour 1,13,19
+   scheduled_minute 27
+
+specifies that poll be run at 1:27, 13:27, and 19:27 each day.
+ 
 By default, sr_poll sends its post notification message to the broker with default exchange
 (the prefix *xs_* followed by the broker username). The *post_broker* is mandatory.
 It can be given incomplete if it is well defined in the credentials.conf file.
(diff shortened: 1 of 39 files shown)


petersilva:
The PR looks really good.  I read over the commits, and they seem fine.  I'm a bit puzzled about what to do about the unit tests.
I'm looking at unit tests in other contexts, and don't understand how to fix these other things either... some thought needed, but I don't know if the unit test issues show slow/hamper this PR.  more thoughts welcome.

petersilva (approved):
Looks great.  I'm not sure what to do about the unit tests... they have other issues also.
perhaps merge first and fix after?

petersilva:
OK... well I tried to get the unit tests to pass. I'm pretty sure they will AFTER this gets merged...  but I can't figure out why it isn't passing already.

petersilva:
sigh... I merged it, but the unit tests still have a failure on ubuntu 20.04 as @gcglinton had mentioned.
have to look at that some more, but deal with it post merge.

PR #1012, opened 2024-04-05
Make persistent a configurable option (v2)

reidsunderland:
As discussed, add persistent as an option so we can easily do A-B testing to compare performance. I also fixed some defaults in the French documentation that looked wrong.

Thanks @andreleblanc11 for the French translation of the persistent option.


petersilva (approved):
looks reasonable... we have to set both durable and persistent independently.

reidsunderland:
Yeah. I was trying to figure out how to make sense of that. At first glace it feels like they should be tied together sometimes.

But the consumer needs to set `durable` and the publisher needs to set `persistent`. So I decided to leave them completely independent. Obviously some components consume and publish, but the options still apply to different parts.

PR #1019, opened 2024-04-17
Azure transfer driver

gcglinton:
**Draft**

This implements an Azure transfer driver, to complement the S3 one from #1010.


petersilva (approved):

gcglinton:
Huh.. I didn't mean to merge this so quickly, but... great!

petersilva:
commit early-commit often... the gethttps thing... will it provide the proper authentication?

gcglinton:
> commit early-commit often...

❤️

>the gethttps thing... will it provide the proper authentication?

No. It's just giving the URL to the file in question. The assumption would be that the publisher would have anonymous access turned on for the account/container in question, so SAS tokens wouldn't be required. If auth was needed, the consumer could just use the Azure transfer driver, which does accept credentials.

Happy to change that if you want, but the creds returned will be those of the publisher, so... that might not be ideal. Unless we add a config option for "publish" creds that the driver uses when it returns the HTTPs URL.

petersilva:
fair point... forget it for now.

PR #1021, opened 2024-04-17
Issue1020 adds sourceFromMessage option.

### LINKED ISSUE #1020: add source field interpretation to sr3 (some v2 deployments use this.)
There are v2 flows we found in operation that rely on one component setting the *source* field for downstream components to use.  But that is only safe for messages whose origin you are confident in.
sr3 updo now was dropping the *source* field for incoming messages for this reason, but there are times when, if you know the source of the messages, it is kind of handy to trust the source.

so create a new flag option *sourceFromMessage* , and when it is set, accept the *source* field from inbound messages.

petersilva:
fix #1020 in the most obvious way.


petersilva:
closing in favour of #1022 later...

PR #1026, opened 2024-04-24
Wis2 integration - MQTT publishing using sr3 stack

mshak2:
We worked with Toms stuff (https://github.com/ECCC-MSC/msc-wis2node)
- Rearranged to publish to our mqtt, using his library to derive the topic

### DIFF
--- a/sarracenia/config.py
+++ b/sarracenia/config.py
@@ -322,9 +322,9 @@ r"""
 
 
 def isTrue(S):
-    s = S.lower()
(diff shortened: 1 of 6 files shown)


petersilva (approved):
looks good... we'll see whether we can get matching stuff accepted by Tom.
I was working with @mshak2 so my review isn't really independent... someone else should also review.

andreleblanc11 (approved):

PR #1027, opened 2024-04-25
EUMETSAT poll plugin update

reidsunderland:
un-URL-encode posted paths so files are downloaded into a nice folder structure


petersilva (approved):
prettier indeed.

PR #1030, opened 2024-04-28
Issue1029 wmo00 accumulated file format support

petersilva:
WMO-00 is a format for data exchange with WMO GTS nodes.

One needs to put a header and possibly a trailer on the content of inbound individual files. Once the headers are added the data for multiple messages are catenated together.  The resulting accumulation file is then named CCCC00000000.a  where the CCCC is the WMO origin code, the 8 digits are a numbers to make the files unique, and the a suffix is a type specifier.

There are two  plugins introduced to support:  

* filter.wmo00_accumulate  -- given a flow of individual WMO messages create the accumulated files.
* filter.wmo00_split -- given a flow of accumulated files, create files each containing one WMO message.

Reference: https://library.wmo.int/viewer/35800/download?file=386_en_1.pdf&type=pdf&navigator=1
( Page 137... the accumulated message  format.)

Describing the patches:

So 
* sarracenia.message.getContent() has been improved so that it works with files that are on the local machine.
  You don't have to calculate the file name and open it. just call getContent() it will lookup the baseDir setting,
  and check if such a file exists locally.
* message.getContent() was also improved to work with file:/ urls (only worked with http/https before.
* the two plugins.
* robustification: some fixes to mainline code resulting from messages with different and options fields being omiteed.
* was getting annoyed by having to run stop after foreground... turned out it was not checking for missing.
* the rest is just the result of attempted rebasing from development... apparently starting point just was not uptodate.

How it was tested:

set up subscription to WMO-BULLETINS on a test marchine, then have the two modules called 
```

    callback filter.wmo00_accumulate
    callback filter.wmo00_split

```

This provides a round-tripping test. During development ran with messageCountMax 50 to see short runs.  Looking at the logs, could determine if the files were coming out of the grouped file intact.  Also, since the file
naming cannot match (using randomized sequence numbers) applied md5 checksums to make
correlation of input and output easier.

Once it was working, ran it for a few hours on a full international WMO GTS feed.
looking for errors.  So the two are at least consistent with each other.

Used international WMO bulletins present in the subscription as examples to better understand the formats.

next step would be interop tests, but probably need it merged first.

**EDIT**: _changed names of modules from wmo00_write -> wmo00_accumulate,  wmo00_read -> wmo00_split._

### DIFF
--- a/sarracenia/flowcb/log.py
+++ b/sarracenia/flowcb/log.py
@@ -70,7 +70,16 @@ class Log(FlowCB):
         if self.o.logMessageDump:
             return msg.dumps()
         else:
-            return msg['baseUrl'] + ' ' + msg['relPath']
+            s = ''
+            if 'baseUrl' in msg:
+                s+= msg['baseUrl'] + ' '
+            if 'relPath' in msg:
+                s+= msg['relPath']
+            elif 'retrievePath' in msg:
+                s+= msg['retrievePath']
+            else:
+                s+= 'badMessage'
+            return s
 
     def _messageAcceptStr(self,msg):
         if self.o.logMessageDump:
@@ -115,13 +124,17 @@ class Log(FlowCB):
             op=','.join(msg['fileOp'].keys())
 
             if op in ['link']:
-                s+= f"a link to {msg['fileOp']['link']} with baseUrl: {msg['baseUrl']} "
+                s+= f"a link to {msg['fileOp']['link']} "
             elif op in ['rename']:
-                s+= f"a rename {msg['fileOp']['rename']} with baseUrl: {msg['baseUrl']} "
+                s+= f"a rename {msg['fileOp']['rename']} "
             else:
-                s+= f"a {op} with baseUrl: {msg['baseUrl']} "
+                s+= f"a {op} "
         else:
-            s+= f"a file with baseUrl: {msg['baseUrl']} "
+            s+= f"a file "
+
+        if 'baseURl' in msg:
+            s+= f"with baseUrl: {msg['baseUrl']} "
+
         if 'relPath' in msg:
             s+= f"relPath: {msg['relPath']} "
         if 'retrievePath' in msg:
--- a/sarracenia/flowcb/log.py
+++ b/sarracenia/flowcb/log.py
@@ -132,7 +132,7 @@ class Log(FlowCB):
         else:
             s+= f"a file "
 
-        if 'baseURl' in msg:
+        if 'baseUrl' in msg:
             s+= f"with baseUrl: {msg['baseUrl']} "
 
         if 'relPath' in msg:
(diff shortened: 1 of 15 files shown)


petersilva:
config settings:
```

subtopic *.SSC-DATAINTERCHANGE.WMO-BULLETINS.#

baseDir  /apps/sarra/public_data

wmo00_work_directory /tmp/wmo00
wmo00_origin_CCCC CWAO
wmo00_encapsulate on

callback filter.wmo00_accumulate
callback filter.wmo00_split

expire 300
messageCountMax 10
batch 5
#batch 100
post_format v02

accept .*
```

sample run:
[wmo00_run.log](https://github.com/MetPX/sarracenia/files/15140798/wmo00_run.log)

**UPDATE**:  _module names changed... the attachment is a run from before the change was made so modules have the old names._

petersilva:
fwiw... I call it WMO-00 format because I heard that somewhere (possibly in wmo meetings.) I think it refers to bytes 9 and 10 in the header... which are 0's. (aka nulls.)   There is another type described on the same page, where the type specifier is 01 ... but that file format is deprecated according to the source, and our French colleagues.

petersilva:
questions:

*  there is wmo00_work_directory  which is where the output files are written.
   Should it just use self.o.baseDirectory when writing accumulated files? 
   I think the accumulated files should not be kept around their names are not useful, and people likely won't want them, they are just temporary files to feed to another server.  The individual files are much more useful.

* Similarly, when extracting (with wmo00_read) from the accumulated file, we write into wmo00_work_directory... is that good? Those are files that we would expect people to subscribe to , so perhaps "directory" is maybe a good root? 

I'm scarred by amtcpwriter... ;-) where nobody knows what _read and _write are referring to.  perhaps different names for the modules.... suggestions?
wmo-00 definitely refers to the accumulated files, not the individual messages.

* wmo00_write becomes wmo00_accumulate (which builds accumulations of individual files.)
* wmo00_read becomes wmo00_separate   (parse accumulations into individual files.)
  or wmo00_extract ?

Would names like that be clearer?

petersilva:
I went with wmo00_accumulate and wmo00_split ... which seem a bit more descriptive.
other suggestions. welcome.

petersilva:
The other thing missing... is how will this be tested... no integration in flow_tests or unit tests...

sarracenia/flowcb/log.py
```diff
@@ -115,13 +124,17 @@ def _messagePostStr(self,msg):
...
-                s+= f"a rename {msg['fileOp']['rename']} with baseUrl: {msg['baseUrl']} "
+                s+= f"a rename {msg['fileOp']['rename']} "
             else:
-                s+= f"a {op} with baseUrl: {msg['baseUrl']} "
+                s+= f"a {op} "
         else:
-            s+= f"a file with baseUrl: {msg['baseUrl']} "
+            s+= f"a file "
+
+        if 'baseURl' in msg:
```
reidsunderland: There's a typo here:

```
#        v
if 'baseURl' in msg:

petersilva: good eye!

Issue #908, opened 2024-01-26
AM socket forking connections despite remaining ESTABLISHED


andreleblanc11:
It was noticed through the logs overnight that the AM socket will occasionally fork connections for some IPs, despite nothing wrong showing in the logs. Looking at `netstat`, it shows that the connections remain `ESTABLISHED` despite still forking.

It's also worth noting that there isn't any data loss and the forked connections will take up the workload from its predecessor.

## Stats
Last log lines of an unexpected forked connection (No indications of errors)

```
2024-01-26 13:40:26,888 [DEBUG] sarracenia.flow run  not throttling: limit: 0
2024-01-26 13:40:26,888 [DEBUG] sarracenia.flow run sleeping for 0.08
```

At the time, there were 20 total connections forked from the parent. Looking at the IPs in netstat, they match with the forking order shown inside the parent log.
```
netstat -anp | grep :5005 | grep ESTABLISHED | wc -l
21
```

## Initial analysis
My initial guess is that we aren't losing connections from the connected hosts. However, they are still trying to connect on the same port (for whatever reason?) and we need to find a way to deal with those connections.

If it's the case, a simple solution for this could be to introduce a list of `alreadyAcceptedIPs` and have it checked before forking. Would need to test to confirm theory.

petersilva:
This could be correct behaviour.... are all of the connections live, or only some?
Someone could configure multiple connections from the same host.  It's not necessarily an error.

petersilva:
"continuously" ... that means forking a lot... but you only report 20 over a night ... 12  hours?  need a better handle on frequency...
to me continuously would mean forking every minute or so...  constantly would be every second...   forking a couple of times a night?  "occasionally" ... ?

andreleblanc11:
The connection on the forked orphan gets dropped once the new fork gets created.

Comparing with the current Sundew AM sockets, it looks like the connections get broken every so often and sundew auto-restarts them. This might correlate with what is happening with the sr3 AM sockets.

For the same IP,

    On sundew, 

    `[ERROR] connection broken` gets stated in the logs at : 3:23 , 6:26 and 9:13 Z last night

    On sr3,

    Connections got forked at : 2:11 , 3:11 , 5:11, 8:11 and 10:11 Z last night

petersilva:
[...]

PR #1031, opened 2024-04-29
Issue908 - Improve general am/gather code and properly kill children zombies.

andreleblanc11:
The orphan killing is working, but needs to wait for another fork to happen before killing off the latest zombies.

This is because I have the orphaned process kill script running every ten minutes while the AM server will fork a new connection just after the previous one terminates. This leaves a gap for child zombies to still live, but this should be fixed once the core issue of #908 is corrected.


petersilva (approved):
I find it weird to wait after spawning... I mean it's not wrong... it should work...
I just find it odd.. never mind... it's fine.

andreleblanc11:
I tried adding it under the `except TimeoutError` so that it could check at least every 64 seconds of exponential backoff. However I was getting `[Errno 10] No child processes` as there wasn't any children yet (on startup). 

It felt weird adding another try/except inside of an except already

andreleblanc11:
The unit tests keep failing, but it's not related

https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/

PR #1032, opened 2024-04-30
flakey flow test failing when polls killed with -9 cannot remove novip state file

petersilva:
so... set had_vip the previous state based on the presence/absence of the novip file.
so it it does have the vip... it triggers the transition to remove it.

This is the reason flakey flow test suite are failing with two polls in wvip sometimes...

this is the test that fails, usually with both sftp polls being in wVip state.
```

test 36 success: there should be 1 process in wVip state

```

The real cause is that the vip flipped, but since it already had the vip (default true..)
it doesn't remove the novip file.

### DIFF
--- a/sarracenia/flow/__init__.py
+++ b/sarracenia/flow/__init__.py
@@ -212,7 +212,8 @@ class Flow:
 
         # metrics - dictionary with names of plugins as the keys
         self.metricsFlowReset()
-        self.had_vip = True
+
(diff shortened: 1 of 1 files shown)


petersilva:
unit tests always fail because coverage is so low that the list of what is not covered is too big.
I've talked to Greg... he is supposed to fix that.

andreleblanc11 (approved):

PR #1034, opened 2024-05-01
updatePaths work without being supplied new_dir and new_path

petersilva:
For @reidsunderland and @petersilva , this routine in the sarracenia.Message class does not satisfy the Princple of Least Astonishment.  If you have a python method that has =None for a value... you should be able to omit it.

looking through the code, it is clear that the code will crash if either one or both of new_dir and new_file arguments are omitted. in all places in the existing code, both args are supplied.

However to make it nicer to use for plugin writers, added logic to infer the correct values from existing message fields whenever possible.

### DIFF
--- a/sarracenia/__init__.py
+++ b/sarracenia/__init__.py
@@ -769,10 +769,22 @@ class Message(dict):
         ])
         if new_dir:
             msg['new_dir'] = new_dir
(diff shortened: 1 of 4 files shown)


reidsunderland (approved):
Looks good to me, thanks!

PR #1037, opened 2024-05-02
Better sender io errors (download also...)

petersilva:
Robbie asked for better error messages... yeah...

* wrap all interactions with transfer protocols in try/except so instead of gettting stack traces we get a succinct summary:  

I ran the flow tests and they pass.

### DIFF
--- a/sarracenia/flow/__init__.py
+++ b/sarracenia/flow/__init__.py
@@ -2230,13 +2230,21 @@ class Flow:
             cwd = None
             if hasattr(self.proto[self.scheme], 'getcwd'):
                 if not self.o.dry_run:
-                    cwd = self.proto[self.scheme].getcwd()
+                    try:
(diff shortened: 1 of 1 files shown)


reidsunderland (approved):

reidsunderland:
We lost some useful information with this change. Now we only see which file failed to send, before we got a more generic error, but we saw the full path.

This is what we got before:

https://github.com/MetPX/sarracenia/blob/6ad476d5d1438bfa101f5754ed11fbc0cccd5b00/sarracenia/flow/__init__.py#L2545-L2547

The new message doesn't include the path:

https://github.com/MetPX/sarracenia/blob/6ad476d5d1438bfa101f5754ed11fbc0cccd5b00/sarracenia/flow/__init__.py#L2403-L2404

PR #1041, opened 2024-05-05
Exchange split override

### LINKED ISSUE #1040: more flexible load-distribution with exchangeSplitOverride (accelerate DMS processing of RCM data.)
The exchangeSplit function allows a single flow to send outputs to different exchanges, numbered 1...n to provide load distribution. The built-in processing does this in a fixed way based on the hash of the identify field.  

The purpose of exchangeSplit is to allow a common set of downstream paths to receive a subset of the total flow, and for products with similar "routing" to land on the same downstream node. For example, a file with a given checksum, for winnowing to work, has to land on the same downstream node.

Now in the case of processing for the RADARSAT Constellation Mission (RCM), the passes/products are delivered in directories.  The downstream consumer of RCM is DMS, which is, itself, an array of processing nodes.  DMS needs to receive the complete pass (the entire directory content) on a single node in order to process the input correctly.

To accomplish that today, we wait until the directory is complete on the data pump, tar up the directory into a single file and then announce the tar file for DMS processing.   As these files are in the gigabyte range, this introduces delay on the data pump to:

* wait until all the data is available.
* tar it all up, 
* only then start the transfer to DMS
* DMS will then un-tar it.
* only after it is tarred and untarred will processing begin.  This likely means minutes of delay on every pass.

If we can change postExchangeSplit to use a different method of deciding which node gets data, instead
of only the checksum, then we could arrange to have all products in the same directory end up on the same node.

* each file would be forwarded as received (long before the rest of the files are received.)
* all files from the same directory would go to the same downstream node.
* having to wait for the directory to be complete, and the subsequent tar, transfer, and untar are all avoided.

This should save substantial processing delay time.

Mechanism:

Add processing for  *postSuffixOverride* field. When present, it sets which exchange in the array is used to process that message.  A plugin that processes the message can decide on any desired load distribution method, by adding that field to the message.

petersilva:
fixes #1040 

the purpose is explained there.  The implementation is like... two lines.
there is a lot more documentation than implementation.

Could use accept/distbydir plugin to potentially make DMS processing faster.

### DIFF
--- a/docs/source/How2Guides/FlowCallbacks.rst
+++ b/docs/source/How2Guides/FlowCallbacks.rst
@@ -286,6 +286,92 @@ of the one from the message, as the original is necessary for successful upstrea
 * msg['new_subtopic'] ... the subtopic hierarchy that will be encoded in the notification message for downstream consumers.
 
 
+Override Fields
+---------------
+
+To change processing of messages, one can set overrides to change how built-in algorithms work.
+For example:
+
+* msg['nodupe_override'] = { 'key': ..., 'path': ... }   changes how the duplicate detection operates.
+* msg['topic'] ... defines the topic of a published message (instead of being calculated from other fields.)
+* msg['exchangeSplitOverride'] = int ... changes how post_ExchangeSplit chooses among multiple postExchanges.
+
+
+
+Customizing Duplicate Suppression
+---------------------------------
+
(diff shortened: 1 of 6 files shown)


andreleblanc11 (approved):
Reviewed in dev meeting

Issue #980, opened 2024-03-18
AM / Bulletin renaming Miscellaneous fixes to mimic Sundew behaviour


andreleblanc11:
Place holder for all small fixes made to `flowcb/gather/am.py` or `bulletin.py`. These were mostly all picked up by my auditing script, comparing with the old NCP nodes.

Fixes made so far in PRs

Reject float types for Julian day value in getTime : https://github.com/MetPX/sarracenia/pull/979
Add missing tokIsYear method from Sundew. Prevents some erronous bulletins not being filtered. : https://github.com/MetPX/sarracenia/pull/973
Add fix to handle improperly parsed station names from getStation : https://github.com/MetPX/sarracenia/pull/963
Properly handle trailing spaces at end of bulletin headers : https://github.com/MetPX/sarracenia/pull/956
Move AM renamer to gather. Allows us to properly use sundew variable expansions. : https://github.com/MetPX/sarracenia/pull/952
Lots of AM corrections (Renamer + AHL corrections in gather) : https://github.com/MetPX/sarracenia/pull/946

andreleblanc11:
With https://github.com/MetPX/sarracenia/issues/997 's fix that got pushed last week, I'm hoping that those were the final changes to AM before NCP move. 

There hasn't been any Sundew compatible bugs found for about two weeks now.

andreleblanc11:
Found another bug. A `BBB` entry in the header was wrong and sr3 didn't filter it out like Sundew.

Sundew has a method that does this and more to verify the header. This should be likely added to sr3 -> https://github.com/MetPX/Sundew/blob/main/lib/bulletin.py#L601-L671

Not urgent. The sundew receiver downstream was able to filter the problem out so clients won't notice.

petersilva:
BBB is OK, and should make it through... it depends on which header... I would only filter out invalid revision codes "AMENDMENT" is clearly wrong, but AMD, AAA, BBB, COR, are all OK.

UPDATE: nvmd... I understand you were referring to the header in general now, my bad.

andreleblanc11:
Need a patch for https://github.com/MetPX/Sundew/issues/20

andreleblanc11:
Take note of https://github.com/MetPX/sr_insects/pull/43#issue-2333937950

- Some american bulletins `*KWAL*` and canadian `SACN[34]*` bulletins  are failing

andreleblanc11:
There hasn't been any bugs found in the last couple of months. A good sign.

[...]

PR #1044, opened 2024-05-13
Issue1036 - Add unit tests for AM code

andreleblanc11:
These unit tests are mostly based off of the errors that were found under #980, adding also testing for regular bulletins and errors from personal observations noted when monitoring logs. 

Suggestions are welcome for adding more tests to the repertoire.

### DIFF
--- a/tests/sarracenia/flowcb/gather/am_test.py
+++ b/tests/sarracenia/flowcb/gather/am_test.py
@@ -1,0 +1,178 @@
+import pytest
+import os, types, copy
+
+#useful for debugging tests
+import pprint
+def pretty(*things, **named_things):
+    for t in things:
+        pprint.PrettyPrinter(indent=2, width=200).pprint(t)
+    for k,v in named_things.items():
+        print(str(k) + ":")
+        pprint.PrettyPrinter(indent=2, width=200).pprint(v)
+
+from sarracenia.flowcb.gather.am import Am
+import sarracenia.config 
+
+from sarracenia import Message as SR3Message
+from sarracenia.flowcb.rename.raw2bulletin import Raw2bulletin
+
+class Options:
+    def __init__(self):
+        # self.o = sarracenia.config.default_config()
+        self.logLevel = "DEBUG"
+        self.logFormat = ""
+        self.queueName = "TEST_QUEUE_NAME"
+        self.component = "flow"
+        self.config = "foobar_am.conf"
+        self.sendTo = "am://127.0.0.1:5005"
+        self.pid_filename = "/tmp/sarracenia/am_test/pid_filename"
+        self.directory = "/tmp/test/directory"
+        self.housekeeping = float(39)
+        self.fileAgeMin = 0
+        self.fileAgeMax = 0
+        self.post_baseUrl = "http://localhost/"
+        self.post_format = "v02"
+
+    def add_option(self, option, type, default = None):
+        if not hasattr(self, option):
+            setattr(self, option, default)
+    pass
+
+def make_message():
+    m = SR3Message()
+    m["pubTime"] = "20180118151049.356378078"
+    m["topic"] = "v02.post.sent_by_tsource2send"
+    m["mtime"] = "20180118151048"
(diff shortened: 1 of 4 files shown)


gcglinton:
This looks great! Lots of good testing here for individual cases you've seen; not something I've been able to do thus far.

Only observations I'd make are...
1 - Perhaps the test names should reflect the method being evaluated. These, at their core, all seem to be testing `correctContents`, so maybe prefix them all with `correctContents__` (`test_bulletin_wrong_station` => `test_correctContents__bulletin_wrong_station`), to group them, or identify what method you're testing. Alternatively, putting them in a class (like [here](https://github.com/MetPX/sarracenia/blob/ddb228cd9daf1ac4fd3fcb4570ec9a1af5c81262/tests/sarracenia/__init___test.py#L128)) to do grouping would work too. Ultimately, it doesn't matter, as they still work, and are still testing things, but defining as a project how you want to structure your tests means it's easier to debug or write them down the road. 

2 - You're building an `options` class from scratch, which is something I had done initially as well. Until I realized I could just instantiate a `sarracenia.config`, and modify things as needed. I think that ends up being a much better way, as you're not having to emulate what `config` is doing. It's actually doing it, so any changes upstream will break the CBs too. Again, at the end of the day, it all works.

tests/sarracenia/flowcb/gather/am_test.py
```diff
@@ -0,0 +1,365 @@
...
+        self.housekeeping = float(39)
+        self.fileAgeMin = 0
+        self.fileAgeMax = 0
+        self.post_baseUrl = "http://localhost/"
+        self.post_format = "v02"
+
+    def add_option(self, option, type, default = None):
+        if not hasattr(self, option):
+            setattr(self, option, default)
+    pass
```
gcglinton, line 39: This should probably be replaced by importing `sarracenia.config`, and then instantiating that on a per-test bassis, setting the options needed.
If you're going to be re-using the same set of options universally, then you can create a fixture, or method for it.

tests/sarracenia/flowcb/gather/am_test.py
```diff
@@ -0,0 +1,365 @@
...
+
+
+# @pytest.mark.depends(on=['test___init__'])
+
+# Test 1: Check a regular binary bulletin. 
+def test_am_binary_bulletin():
+    from base64 import b64encode
+
+    BaseOptions = Options()
+    renamer = Raw2bulletin(BaseOptions)
```
gcglinton, line 103: Since you're calling this external method, it might be good to add a dependency for it above these tests, but that would mean writing unit tests for that other method.

Alternatively, re-scope your tests here to only cover what `gather/am` is doing.

gcglinton: Hm.. I see `Raw2Bulletin`, and `Bulletin` are getting used in `__init__` as well, so maybe we ignore these dependencies until tests are written for those modules.

andreleblanc11: Yeah. Unfortunately, we need both `Bulletin` and `Raw2Bulletin` for a lot of the tests. The scope of these unit tests are mostly for the "bulletin processing and renaming" functionality

But you're right we'll need unit tests for these classes as well. I'll make sure to take note of that.

petersilva: you just end up with unit tests for Bulletin and Raw2Bulletin... still a good thing.
I mean most of the problems you ran into were bulletins with unexpected content, so those tests are doing useful stuff. It's at worst slightly mislabelled.

petersilva (commented):
Greg's comments are great.  You can keep working on the branch to use the sarracenia classes, or we can merge first, and do that work subsequently.  It's a good idea either way.

more tests is more better, as we clearly don't have enough right now.

andreleblanc11:
I think it'd be better to merge now, so at least we'll have something for future PRs

petersilva (approved):
ok... approving on that basis...

PR #1045, opened 2024-05-14
To convert, one must wololo

MagikEh:
To convert, one must [wololo](https://knowyourmeme.com/memes/wololo)
wololo priest

### DIFF
--- a/sarracenia/sr.py
+++ b/sarracenia/sr.py
@@ -2648,7 +2648,8 @@ class sr_GlobalState:
         if os.path.exists(v3_config_path):
             logger.error( f"{component}/{cfg} already exists in v3, remove {v3_config_path} to convert again")
             return
-
+        
+        logging.info( f'(づ◔-◔)づ Wololo' )
         synonyms = sarracenia.config.Config.synonyms
         accept_all_seen=False
         acceptUnmatched_explicit=False
--- a/sarracenia/sr.py
+++ b/sarracenia/sr.py
@@ -2649,7 +2649,7 @@ class sr_GlobalState:
             logger.error( f"{component}/{cfg} already exists in v3, remove {v3_config_path} to convert again")
             return
         
-        logging.info( f'(づ◔-◔)づ Wololo' )
+        logger.info( f'(づ◔-◔)づ Wololo' )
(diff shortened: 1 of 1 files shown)


reidsunderland (changes requested):
`logging` should be `logger`

MagikEh:
> `logging` should be `logger`

Lol, copy pasta from further down the file has failed me? xD
https://github.com/MetPX/sarracenia/blob/wolololo/sarracenia/sr.py#L2768

andreleblanc11:
Omg lol

andreleblanc11:
Priority not high enough

MagikEh:
@reidsunderland see 1b2e374

petersilva:
fwiw... logging will work, but it will use a different logger.  that's why the original error isn't caught. it's not invalid... merely wrong.

petersilva:
Looking at the patch it's right after telling us ''converted config exists... federal off"
we could make --wololo an option to overwrite an existing configuration if it exists.

petersilva:
or just make wololo a synonym of convert.

MagikEh:
Lol, I love the option to convert a config in place via `--wololo`.
Synonym makes less sense on the grounds of the player needing to tell their monk to go convert something which is where they then go wololo on it to actually convert it.

I'll spend some time looking into implementing that

petersilva:
fwiw... implementation guide:

* in config.py ... 
   * add "wololo" to *flag_options*   ... will make it work in a config file.
    * add "wololo": False, to *default_options*.  ... will ensure it defaults to False.
    * find the lines with parser.add_argument( '--debug' ... there's three lines make a copy of them and and change --debug to --wololo.   ... will make it work as a command line flag.
* sr.py
   * with the above changes, when you get into sr.py, *options.wololo* will now exist.  So, in the same part of sr.py where you put your original message, look at the if statement above it, and add consideration of that value.  The "wololo" message should be guarded by the if.
  
so... it's less than 10 lines... then have to add an explanation in the docs.

That's just the first option... aka only overwrite an existing option if --wololo is given.

petersilva:
uh... I decided to work on it... it's finished.. writing it up in an issue.

petersilva:
see #1054

PR #1048, opened 2024-05-15
Unit Test Engineering wmo00

petersilva:
Unit Test Engineering (Ute) branch building some unit tests for wmo00 class.
I don't know when I'll get back to this, so before it gets to stale, I'll try to get it merged.

It's just unit tests for the support plugins that implement WMO-00 file encapsulation.


petersilva:
@reidsunderland happy.

Issue #494, opened 2022-03-29
hierarchical configurations (getting rid of "components")


petersilva:
In the Development meeting on 2022/03/28, it was like a light-bulb was suddenly lit, and people understood that all the components are just the flow algorithm with different defaults for different settings.  People noted:
* it was hard to understand what "a component" is.
* It was kind of cool that they were all implemented by sub-classing sarracenia.flow.

Consensus seemed to be that getting rid of the concept of components would be a win.  but it's a bit late in the sr3 cycle for that.
(and it wasn't really possible until all the 1st round of re-factors was done, because it wasn't until the last re-factor of poll that this became true.  until then poll was a huge outlier in the sub-classing of flow, but now that flowcb/poll/__init__.py has taken the main body of poll, the ¨poll¨ flow is just a selection of callbacks, like the other components.  so now it should be possible.

petersilva:
Step 1:
* express all the components using configuration files (shovel.conf should connect the defaults for shovel, subscribe.conf (or inc?) should.

petersilva:
@reidsunderland  was suggested replacing subclassing by include files.... ie.. 
```
     include shovel.inc

```

petersilva:
so this is implemented in #575... the thing I noticed is that there are very common configurations, senders, watches, etc... and if we have a sub-directory for them, that denotes the same information as the include, but the name of the configuration does not have that information... so in many cases you end up writing configuration files named:   sender_....  that contain the option *include sender.inc*  so you are writing "sender" twice... which is worse from a DRY perspective.

ie.
You could have a:
```

      sender/hoho.conf
                <contains a normal sender config>

```

vs.

```
     sender_hoho.conf
                 include sender.inc
                <contains a normal sender config>

```
So in a very common case, you end up writing the component name twice instead of once.
big deal? dunno... 

hmm...I'm thinking to keep the directory stuff around as shorthand... 
* if the *last* directory in the config path matches a sub-class, then it is assumed to be that kind of flow.
* if not it defaults to a parent class (Flow) (or explicit specification with flowMain option.)

so then you can group things like:

```
    pitcher_use_case/
                          watch/
                                  server.conf
                                  client.conf
                        sender/
                                  push.conf
                       subscribe/
                                  pull.conf
                      bobo.conf
                      lala/
                              fifi.conf
```

resulting in sr status showing lines like:
```

pitcher_use_case/watch/server
pitcher_use_case/watch/client
pitcher_use_case/sender/push
[...]

PR #1049, opened 2024-05-15
adding new_pathWrite() as per Reid suggestion

petersilva:
I worked on this last week and got distracted by ddsr-dev replacing the gswob feed to the datamart.  This is kind of the complement to getContent() that is used to write local files and return a well-formed message for it.

It should be helpful in callbacks that transform data.  I don't know when I will get back to this either.

### DIFF
--- a/sarracenia/__init__.py
+++ b/sarracenia/__init__.py
@@ -391,7 +391,7 @@ class Message(dict):
         self['_deleteOnPost'] = set(['_format'])
 
 
-    def computeIdentity(msg, path, o, offset=0):
+    def computeIdentity(msg, path, o, offset=0, data=None) -> None:
         """
            check extended attributes for a cached identity sum calculation.
            if extended attributes are present, and 
@@ -400,6 +400,10 @@ class Message(dict):
            then use the cached value.
 
            otherwise, calculate a checksum. 
+           If the data is provided, use that as the file content, otherwise 
+           read the file form the file system.  
+
+           Once the checksum is determined,
            set the file's extended attributes for the new value.
            the method of checksum calculation is from options.identity.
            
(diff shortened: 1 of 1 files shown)


petersilva:
@reidsunderland worry:
* if message has inlined file data... what happens?
  *  current implementation ignores it.   result if present... new data will be over-written by message data if download==True.
 
what is a good behaviour?
* if data==None
  *  use data from inline, write to disk.

* if data!=None
 * consider this an override for inline data...
 * look at options... 
     * if inlining is active, then update the inline field, in addition to writing the data.
     * if ! inlining, then remove the field.
  
* if download==True... potential for double file write.
   *  perhaps not because mtime & checksum.... 
  
For converter plugins in general, all the options feel weird...
* shovel(flowcb/filter/wmo2msc)
   * change line termination,
   * changes file tree.
   * shovels feel weird because they are supposed to be about messages... not data.
   * renamer is natural
* sarra
  *  feel less weird than shovels... not convinced they are wrong... 
  * renamers are natural here also.
  
* sender (flowcb/filter/wmo00 ) 
  * wmo00_accumulate makes sense... but is it a general thing, or it just makes sense in that case?
  * renamers are natural here also.
  
* flow ... should transformers use a *flow* to denote that they aren't doing what any standard components do.

* transform/xform/convert/fx/    ... should we establish another type of flow for conversion?
   * just a sarra with download off, because we will be transforming data.
   * this makes it identical to a shovel... ;-)

 This is essentially re-inventing the discussion in #494... and we the hierarchy makes sense again...

for now, conclusion is to use *flow/cvt_*   to represent transformers.




  

Also need to set msg['size'] = len(data)

petersilva:
last commit included "size" support as well. so I guess it's good to go.

petersilva:
I'm still puzzled as to why I haven't figured out how to leverage Transfer classes for this stuff, or have code from transfer moved here... or... Transfer/\__init__.py there is read_writelocal which is exactly the same logic... but a bit more complete.

PR #1050, opened 2024-05-15
Add post_baseUrl to delete on post, changes to logging

reidsunderland:
I had made some changes that I forgot to push for #1042 

The most important one is adding the post_baseUrl to the delete on post, so it doesn't propagate to the downloading component.

### DIFF
--- a/sarracenia/flowcb/poll/poll_NASA_CMR.py
+++ b/sarracenia/flowcb/poll/poll_NASA_CMR.py
@@ -331,20 +331,21 @@ class Poll_nasa_cmr(sarracenia.flowcb.FlowCB):
                             data_url = url['URL']
                 
                 # Now that above loop is done, should have a data URL and maybe an md5_url
-                sumstr = None
+                new_identity = None
                 if md5_url and self.o.dataSource == "podaac":
                     try:
                         md5_resp = requests.get(md5_url)
                         md5 = md5_resp.text.split(" ")[0]
                         logger.debug(f"MD5 Checksum: {md5}")
-                        sumstr = {"method":"md5", "value":md5}
-                    except:
-                        logger.warning(f"Could not get MD5 Checksum for {data_url}, posting without it.")
-                        sumstr = {"method":"cod", "value":"md5"}
+                        new_identity = {"method":"md5", "value":md5}
+                    except Exception as e:
+                        logger.debug("Exception details:", exc_info=True)
+                        logger.warning(f"Could not get MD5 Checksum for {data_url}, posting without it. {e}")
+                        new_identity = {"method":"cod", "value":"md5"}
                 # source is podaac but md5_url isn't available
                 elif self.o.dataSource == "podaac":
                     # tell downloads to use md5
-                    sumstr = {"method":"cod", "value":"md5"}
+                    new_identity = {"method":"cod", "value":"md5"}
 
                 # finally create the message!
                 if data_url:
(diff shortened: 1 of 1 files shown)


petersilva:
confused... if you are in a poll, you should just set baseUrl ... why do you need post_baseUrl?

reidsunderland:
Because it eventually gets overridden later in the code. https://github.com/MetPX/sarracenia/issues/951

For this plugin, `pollUrl` should be set to https://cmr.earthdata.nasa.gov/search/granules.umm_json in the config file. Because post_baseUrl is not set in the config, self.o.post_baseUrl gets set to the value of pollUrl. (Here's an example config: https://github.com/MetPX/sarracenia/blob/development/sarracenia/examples/poll/nasa_cmr_podaac.conf).

For example, if the poll creates a message with `msg['new_baseUrl'] = "https://downloadfromhere.com"` and `msg['baseUrl'] = "https://downloadfromhere.com"` and does not set `msg['post_baseUrl']`...

Once the list of messages is returned from the poll, they get filtered in the flow algorithm. filter() calls updateFieldsAccepted(), which calls updatePaths(). And updatePaths has this logic:

https://github.com/MetPX/sarracenia/blob/9f61e90d3ea3810559f7423a37c22f03b7951a9c/sarracenia/__init__.py#L790-L800

If I **don't** set msg['post_baseUrl'], then we'd go into the elif case, because options.post_baseUrl **is** set (to the value of pollUrl from the config). This causes updatePaths to set msg['new_baseUrl'] to the value of options.post_baseUrl (== pollUrl).

So now the message has been updated to be this:

`msg['new_baseUrl'] = "https://whatever_I_set_pollUrl_to_in_the_config.com"` and `msg['baseUrl'] = "https://downloadfromhere.com"`

Then finally, we get to this part of the code:

https://github.com/MetPX/sarracenia/blob/47dd8dc3325f96d49befc2609556e0d1615cf69d/sarracenia/flow/__init__.py#L517-L522

new_baseUrl is in m, and m['baseUrl'] and m['new_baseUrl'] are not equal, so this part of the code sets the value of baseUrl to whatever is in new_baseUrl.

Now we have: `msg['new_baseUrl'] = "https://whatever_I_set_pollUrl_to_in_the_config.com"` and `msg['baseUrl'] = "https://whatever_I_set_pollUrl_to_in_the_config.com"`.

---

There's a little bit of history missing from #951. This plugin used to work (see original code and examples here https://github.com/MetPX/sarracenia/commit/f1d7697c14244d7e4ad84abe976c2b89b99d0358) because I was leaving both pollUrl and post_baseUrl undefined in the config file. In that situation, we'd fall into the else on line 796 and updatePaths would not change the 'baseUrl' and 'new_baseUrl' fields in the message.

But something changed between October 11, 2023 and now where the poll won't start if pollUrl is not set in the config file. I used to be able to set `self.o.pollUrl` in the plugin's __init__ and it would work. (I did that because I figured it was easier for the user to just define the plugin in the config and not have to worry about setting pollUrl themself, because it's always the same URL)

reidsunderland:
I also tried setting self.o.pollUrl and self.o.post_baseUrl in the poll, but that doesn't propagate to other parts of the code, because I think it has a deepcopy of the options, not the original instance used in other parts.

petersilva:
Can you try putting self.o.add_option( 'pollUrl' , type='str', default='thevalueUwant' )  in the __init__ ...
and see if that works. I saw you experimented with just setting it in the plugin, but I'm wondering if using add_option will work better.

petersilva (approved):

PR #1052, opened 2024-05-16
Build Unit test boilerplate

gcglinton:
Added a ton of files, most of which are empty, containing no tests.

They're just there to setup the ground-work for future tests.

I also centralized the "debug" stuff into the `conftest.py` file in the root of the tests directory. Now when looking to debug anything during test writing, you can just call `debug.pretty(YourThingToDebug)`, and it should work.
It's also possible to centralize some fixtures there if needed. Things like building a base config, or messages could be put there, and used by all tests. 
You can also create a `conftest.py` file in any given directory, and have those contents be available in that context, either building on those defined above it, or over-riding them.

Some of that is outlined [here](https://docs.pytest.org/en/8.2.x/how-to/fixtures.html)


petersilva:
You didn't ask for any reviewers... (top-right?) 
We normally don't merge without at least one peer review.
Is this a WIP or do you want it merged? fwiw... I have a bunch of unit test engineering (ute) branches... they are not yet merged...
they all just use plain print() what does the pretty print stuff give us?

Issue #927, opened 2024-02-07
sr3 sanity not picking up missing instance


andreleblanc11:
## What happened

A poll failed, and its instance was set to "missing" at 22:47Z

        Component/Config                         Processes   Connection        Lag                              Rates
                                                State   Run Retry  msg data   Queued   LagMax  LagAvg  %rej     pubsub   messages     RxData     TxData
                                                  -----   --- -----  --- ----   ------   ------  ------  ----   --------       ----     ------     ------
        poll/airnow_http_poll                    miss    0/1     0   0%   0%      0    0.00s    0.00s  0.0%  0 Bytes/s   0 msgs/s  0 Bytes/s  0 Bytes/s
              Total Running Configs:   0 ( Processes: 24 missing: 1 stray: 0 )
                              Memory: uss:1.7 GiB rss:1.9 GiB vms:5.1 GiB
                          CPU Time: User:221354.24s System:4841.57s
                 Pub/Sub Received: 0 msgs/s (0 Bytes/s), Sent:  0 msgs/s (0 Bytes/s) Queued: 0 Retry: 0, Mean lag: 0.00s
                       Data Received: 0 Files/s (0 Bytes/s), Sent: 0 Files/s (0 B

In the sanity logs, we see that it never got restarted after it failed

    [2024-02-05 22:49:03] sanity: killing hung processes... (no point in SIGTERM if it is hung)
    [2024-02-05 22:49:03] no missing processes found
    [2024-02-05 22:49:03] notice: python module mqtt is missing: cannot connect to mqtt brokers
    [2024-02-05 22:49:03] notice: python module redis is missing: cannot use redis implementations of retry and nodupe
    [2024-02-05 22:49:03]
    [2024-02-05 22:56:04] sanity: killing hung processes... (no point in SIGTERM if it is hung)
    [2024-02-05 22:56:04] no missing processes found
    [2024-02-05 22:56:04] notice: python module mqtt is missing: cannot connect to mqtt brokers
    [2024-02-05 22:56:04] notice: python module redis is missing: cannot use redis implementations of retry and nodupe
    [2024-02-05 22:56:04]

___ 

## Questions
- Was the system in a weird state, where sanity couldn't find the missing instance?
- Is this isolated to `poll/airnow.py` flowcb?  @junhu3 wonders

petersilva:
tried a trival case: running a dynamic_flow test, then do kill -9 of the poll process, then run sr3 sanity.  It started it up as expected, no problem.
[...]

PR #1053, opened 2024-05-17
Fix sanity not starting missing instances

reidsunderland:
We had a group debugging session and fixed this issue.

### DIFF
--- a/sarracenia/sr.py
+++ b/sarracenia/sr.py
@@ -1981,7 +1981,7 @@ class sr_GlobalState:
             component_path = self._find_component_path(c)
             if component_path == '':
                 continue
-            if self.configs[c][cfg]['status'] in ['stopped']:
+            if self.configs[c][cfg]['status'] in ['missing']:
                 numi = self.configs[c][cfg]['instances']
                 for i in range(1, numi + 1):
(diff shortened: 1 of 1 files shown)


petersilva (approved):
This will save the government thousands of dollars... because the failure to detect was causing pager calls.

petersilva:
Please Note: We need to document the change in behaviour... with this patch, sr3 sanity will not longer start up "stopped" configs... only "missing" ones.  That's what we decided was least astonishing... but it is still a change from how it worked before.

petersilva:
close #927

reidsunderland:
I think the documentation always said what it was supposed to do. "sanity: looks for instances which have crashed or gotten stuck and restarts them." https://metpx.github.io/sarracenia/Reference/sr3.1.html

reidsunderland:
I grepped for sanity and couldn't find anywhere (in English) that said it started stopped instances, only crashed or stuck instances.

petersilva:
I updated HOWTO/UPGRADING.rst ... to indicate the change in behaviour. same in french ... committed direct to development.
yeah... I old behaviour was a mistake in the first place, but one with seniority.

PR #1055, opened 2024-05-17
to convert overwriting, one must Wololo

### LINKED ISSUE #1054: to convert one must wololo
started with pr #1045 ... but that was cosmetic... perhaps a joke?

anyways it was an interesting idea.  Currently, sr3 convert

* will only ever convert one configuration per invocation... which is different from most other command that accept groups.
* will never overwrite an existing sr3 configuration.

In this issue we add:

* ability to convert multiple configurations in one invocation.
* ability to overwrite existing config if --wololo provided
* ability to overwite multiple existing configurations if *--dangerWillRobinson=n --wololo convert* provided.

petersilva:
close #1054 

replaces #1045

### DIFF
--- a/docs/source/Explanation/CommandLineGuide.rst
+++ b/docs/source/Explanation/CommandLineGuide.rst
@@ -70,7 +70,7 @@ the broker, or manage the configurations.
  - remove:        remove a configuration.
  - disable:       mark a configuration as ineligible to run. 
  - enable:        mark a configuration as eligible to run.
- - convert:       converts a v2 config to a v3 config.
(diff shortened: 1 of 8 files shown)


petersilva:
after merging this... merge that: https://github.com/MetPX/sr_insects/pull/40
to have flow tests run faster.

petersilva:
after thinking about it for a while... do we need the --wololo part? I mean is just dangerWillRobinson enough? not as funny... but...

petersilva:
@reidsunderland happy...

PR #1058, opened 2024-05-18
Unit Test Engineering of the sarracenia/config.py module.

petersilva:
Unit Test Engineering ... a reasonable number of unit tests for the config.py 
should bring the coverage up from close to zero. but not complete.


tests/sarracenia/config_test.py
```diff
@@ -301,21 +301,55 @@ def test_source_from_exchange():
...
+     assert( options.directory == os.path.expanduser( '~/ex1' ) )
+     assert( len(options.bindings) == 1 )
+     assert( options.exchange == 'xs_bunnypeer' )
+     assert( options.post_exchange == [ 'xs_bunnypeer' ] )
+     assert( hasattr(options,'nodupe_ttl') )
+     assert( hasattr(options,'metricsFilename') )
+     assert( hasattr(options,'pid_filename') )
+     assert( hasattr(options,'retry_path') )
+     assert( hasattr(options,'novipFilename') )
+     assert( hasattr(options,'bindings') )
```
gcglinton, line 351: Does it make sense to write a helper function to check all these at once? Pass it `option`, and a list of the attrs you want to check, and have the helper loop through them, returning `False` if any are missing?

tests/sarracenia/config_test.py
```diff
@@ -90,6 +91,265 @@ def test_variableExpansion():
...
+     options = sarracenia.config.default_config()
+     logger.info( f" {options.fileEvents=} " )
+
+     assert( options.fileEvents == set( ['create', 'delete', 'link', 'mkdir', 'modify', 'rmdir' ] ) )
+
+     options.parse_line( "subscribe", "ex1", "subscribe/ex1", 1, "fileEvents -rmdir" )
+     logger.info( f" {options.fileEvents=} " )
+     assert( options.fileEvents == set( ['create', 'delete', 'link', 'mkdir', 'modify' ] ) )
+
+     #FIXME: must print an error message... how to test for that?
```
gcglinton, line 169: You can capture logs using the `caplog` fixture. I've used it in the `retry` driver tests, like [here](https://github.com/MetPX/sarracenia/blob/4dea8827765ab1b7ac05a8705f20e07be5cf8765/tests/sarracenia/diskqueue_test.py#L236), [here](https://github.com/MetPX/sarracenia/blob/4dea8827765ab1b7ac05a8705f20e07be5cf8765/tests/sarracenia/diskqueue_test.py#L260), and [here](https://github.com/MetPX/sarracenia/blob/4dea8827765ab1b7ac05a8705f20e07be5cf8765/tests/sarracenia/flowcb/retry_test.py#L313).

You set a variable to say if it's there (`False` by default), and then check if a particular message exists in the logs, setting the variable to `True` if it does, and then finally asserting on that variable.

tests/sarracenia/config_test.py
petersilva: optimization can be done post merge... unclear of benefit in any event ( yes suggestion is prettier, but a bit more to understand... )

tests/sarracenia/config_test.py
petersilva: will do that as action later... change is still good.

PR #1072, opened 2024-05-28
Doc updates 

petersilva:
changelog updates here...
I do changelogs separately because parallel development invariably results in merge conflicts for this file.

There's also a change made on launchpad.net documented here.


reidsunderland (approved):
I added a commit to rename idlethreshold -> runStateThreshold_idle in the Command Line Guide. I think it was missed because it was idlethreshold, not idleThreshold.

PR #1084, opened 2024-06-03
Add scheduled flow with http metadata

reidsunderland:
Normal scheduled flow, except it makes an HTTP HEAD request for each file to try to get file size and modification time before posting.


petersilva (approved):
that's super cool, no mods, just an additional standard plugin...

petersilva:
the unit test is result of previous merge. not an issue here.

Issue #1035, opened 2024-05-01
cpu time used in sr3 is higher than v2 for equivalent configuration.


petersilva:
Comparing two pumps, it looks like the sr3 one consumes more cpu time than the v2.
This issue will be used to investigate a bit.

petersilva:
I notice that in v2 in the amqp consumer, when no message arrives, it does exponential backoff....
```python

        self.get_message()
        self.raw_msg = self.consumer.consume(self.queue_name)

        # if no message from queue, perhaps we have message to retry

        if self.raw_msg is None:
            self.raw_msg = self.retry.get()

        # when no message sleep for 1 sec. (value taken from old metpx)
        # *** value 0.01 was tested and would simply raise cpu usage of broker
        # to unacceptable level with very fews processes (~20) trying to consume messages
        # remember that instances and broker sharing messages add up to a lot of consumers

        should_sleep = False

        if self.raw_msg is None:
            should_sleep = True
        elif self.raw_msg.isRetry and self.last_msg_failed:
            should_sleep = True

        if should_sleep:
            try:
               time.sleep(self.sleep_now)
            except:
               self.logger.info("woke from sleep by alarm.. %s " % self.msg.notice)

            self.sleep_now = self.sleep_now * 2
            if self.sleep_now > self.sleep_max:
                self.sleep_now = self.sleep_max
```

in sr3, it instantaneously checks if a message is there, and returns immediately relying on the global loop.

petersilva:
... but it looks like sr3 has the same logic... in the main loop...

```python
                if not stopping:
                    self.gather()

                last_gather_len = len(self.worklist.incoming)
                if (last_gather_len == 0):
                    spamming = True
                else:
                    current_sleep = self.o.sleep
                    spamming = False
 .
 .
 . 
               if spamming and (current_sleep < 5):
                    current_sleep *= 2

petersilva:
Over the period of a week, we see that, on the servers in question:
* when things initially start, there is high cpu usage for a few polls and some sarra's, but polls calm down.
* over a number of days, the cpu usage of busy sarras, which is initially quite low (2%) gradually creeps up.  after a week or so, it's hanging at 99%.
* restarting resets the clock.

my guess: memory fragmentation.  the GC is spending more and more time to find bytes in ever more fragmented chains.

petersilva:
reading about GC:

* https://stackoverflow.com/questions/6115066/how-much-time-is-the-garbage-collector-using

[...]

PR #1086, opened 2024-06-03
Issue1035 retry too much, reduce cpu to some extent... 

petersilva:
improves #1035 hopefully.

trying to reduce cpu time usage.   It creeps up after having run for a long time.  Not sure if this fixes it, but it should help.
So this eliminates all file i/o in the main *run* loop.  there will be no i/o if:
* the retry_queue is empty
* the metrics file has been recently written.

eliminated a lot of file i/o noise for checks for directory existence also.

We'll see how much this helps.  It does help... but the extent is unclear.


andreleblanc11:
Even if this doesn't fix the main load problem, it should still help drag it down at least a bit.

After we merge to `development`, we should pull the latest changes on ddsr development and see if the CPU keeps climbing high.

andreleblanc11 (approved):

petersilva:
I think the basic test was failing because the metrics dir was not being created in one case.  Re-added to beginnning of run() routine.

PR #1094, opened 2024-06-05
Doc feedback

petersilva:
feedback from discussions this morning:

* logFormat should work after this is merged.
* some messages demoted from info to debug.
* other messages demoted to oblivion.
* set option documented.


petersilva:
the logFormat stuff still does not work... that's probably why it is not documented... I think I messed with it for a while, but the programmable log format does not work.

petersilva:
for clarity... I don't know how to fix the logFormat thing... don't plan on working on it further... this PR improves things regardless.

reidsunderland (approved):
I changed the RateMax log messages - i upgraded the message that says when it's throttling back to INFO. And downgraded a different message (that prints the current rate vs ratemax) down to DEBUG.

And French translation, from @andreleblanc11 / Google

petersilva:
I fixed the logFormatting after all...

PR #1099, opened 2024-06-06
Add directory to send, etc. errors

### LINKED ISSUE #1098: Sender IO errors don't show paths
We lost some useful information with this change (#1037). Now we only see which file failed to send, before we got a more generic error, but we saw the full path.

This is what we got before:

https://github.com/MetPX/sarracenia/blob/6ad476d5d1438bfa101f5754ed11fbc0cccd5b00/sarracenia/flow/__init__.py#L2545-L2547

The new message doesn't include the path:

https://github.com/MetPX/sarracenia/blob/6ad476d5d1438bfa101f5754ed11fbc0cccd5b00/sarracenia/flow/__init__.py#L2403-L2404

_Originally posted by @reidsunderland in https://github.com/MetPX/sarracenia/issues/1037#issuecomment-2127901038_

reidsunderland:
Closes #1098. Sorry to steal this from @mshak2, i was trying to debug errors on dev and it was super hard without this.

Creating as a draft, because I need to make sure the flow tests pass.

### DIFF
--- a/sarracenia/flow/__init__.py
+++ b/sarracenia/flow/__init__.py
@@ -2214,7 +2214,8 @@ class Flow:
     # generalized send...
     def send(self, msg, options):
         self.o = options
-        logger.debug( f"{self.scheme}_transport sendTo: {self.o.sendTo}" )
+        sendTo=self.o.sendTo 
+        logger.debug( f"{self.scheme}_transport sendTo: {sendTo}" )
         logger.debug("%s_transport send %s %s" %
                      (self.scheme, msg['new_dir'], msg['new_file']))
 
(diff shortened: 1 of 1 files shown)


petersilva (approved):
I'm wondering if we should make these messages more verbose... as in list both source and destination, and username and host...  my original thinking in not having this was that it is all available on other lines in the log, and so I would just use the grep to find the error and use either vi or grep -B10 -A10 to get the other information.  But I see it's a lot easier to get everything we need if a straight grep will do it... so maybe the error messages should be a lot more explicit.  not sure...

reidsunderland:
I think that would be great, have all the information in one place. It would make troubleshooting a lot easier. 

Especially with batching, the lines immediately above and below don't always have the needed information.

PR #1100, opened 2024-06-06
Issue1089 fileAgeMin and fileAgeMax should be honoured by all components

### LINKED ISSUE #1089: fileAgeMax doesn't work in all situations?
I tried to use fileAgeMax in a sarra, where the incoming messages have mtime and atime headers. It didn't work. 

I think it's because fileAgeMax isn't implemented in gather.message? https://github.com/MetPX/sarracenia/blob/4ca61178718c7e4b8fc458faa6c9e65c5314cc37/sarracenia/flowcb/gather/message.py

I guess this is a feature request? I'm pretty sure it's not something that ever worked before, so not a bug.

petersilva:
closes #1089

* put a check in flow/__init__.py/ filter() routine... which is used by all components.  
* files that are too old, are *rejected* (permanently dropped.)
* files that are too youn, are *failed*, which should result in them being put on the retry list for later.

### DIFF
--- a/sarracenia/flow/__init__.py
+++ b/sarracenia/flow/__init__.py
@@ -962,8 +962,13 @@ class Flow:
             if self.o.messageAgeMax != 0 and lag > self.o.messageAgeMax:
                 self.reject(
                     m, 504,
-                    "Excessive lag: %g sec. Skipping download of: %s, " %
-                    (lag, m['new_file']))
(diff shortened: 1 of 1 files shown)


reidsunderland (approved):
That looks good, thanks!

I see self.o.messageAgeMax/Min, they're not documented. Are they supposed to be user configurable options?

PR #1101, opened 2024-06-06
only run after_post after we actually post

### LINKED ISSUE #1066: after_post runs even though nothing has been posted
This causes a confusing log message saying

```
2024-05-03 02:30:44,644 [INFO] sarracenia.flowcb.log after_post posted to a file with baseUrl: http://example.com/ relPath: data/zps.202405030230.past_hourly.xml
```

If there are after_post plugins, it might also do unexpected things.

@petersilva already found the fix, just indent `self._runCallbacksWorklist('after_post') ` 4 characters so it is inside the if statement.

https://github.com/MetPX/sarracenia/blob/6ad476d5d1438bfa101f5754ed11fbc0cccd5b00/sarracenia/flow/__init__.py#L1125-L1141

petersilva:
close #1066 

kind of obvious.

### DIFF
--- a/sarracenia/flow/__init__.py
+++ b/sarracenia/flow/__init__.py
@@ -1211,23 +1211,26 @@ class Flow:
 
     def post(self,now) -> None:
 
-        if len(self.plugins["post"]) > 0:
+        if hasattr(self.o,'post_broker') and self.o.post_broker:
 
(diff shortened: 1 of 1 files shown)


reidsunderland (approved):
Looks good to me. It's a draft, so maybe i wasn't supposed to review it yet

PR #1106, opened 2024-06-07
Issue1097 fractional rates

### LINKED ISSUE #1097: messageRateMax/messageRateMin documentation/implementation inconsistency
The documentation says that messageRaxMax/Min are float options https://metpx.github.io/sarracenia/Reference/sr3_options.7.html#messageratemax-float-default-0

But they're count options in the code:
https://github.com/MetPX/sarracenia/blob/d095485061da1dbf9e210bfdd38b32da5f033dfc/sarracenia/config.py#L135-L139

https://github.com/MetPX/sarracenia/blob/d095485061da1dbf9e210bfdd38b32da5f033dfc/sarracenia/config.py#L153

I agree with the documentation, messageRateMax should be a float because it allows more flexibility in the values we choose.

petersilva:
close #1097 

does as discussed in the bug report.

### DIFF
--- a/docs/source/Reference/sr3_options.7.rst
+++ b/docs/source/Reference/sr3_options.7.rst
@@ -238,7 +238,7 @@ flag
     an option that has only True or False values (aka: a boolean value)
 
 float
-    a floating point number.
+    a floating point number (3 decimal places max, numbers > 1000 are integers)
 
 list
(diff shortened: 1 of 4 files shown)


reidsunderland (approved):
reviewed in the meeting

Issue #693, opened 2023-06-06
MQTT error recovery (get restart_server and flakey_broker flow tests working.)


petersilva:
To get #506 resolved to the point where MQTT is at least within CI/CD... had to break off the tests of error recovery... because those do not work, and would prevent CI/CD for the more basic stuff that does.
So this issue is about fixing up error recovery for MQTT.

petersilva:
so in restart server, in a run, I noticed that between the poster and the first subscriber, it already lost messages.  To me that problem looks like this problem in the mqtt library itself, rather than anything we can fix: https://github.com/eclipse/paho.mqtt.python/pull/554

But there are possibly other problems to address as well.

petersilva:
MQTT is a marquee feature for version 3.  For compatibility with WMO, need for this to work fully.

petersilva:
so paho-mqtt v2 has been released, and it includes support for manual acks that we need to pass these error recovery tests.
So there is a *paho-mqtt-v2-support* branch, which is a work in progress.   The API is changed, as is error handling, so it isn't a straight-forward migration.  current status:

* *development* branch passes static only....  no_mirror  22/29 ... flakey (does not terminate) restart (tbd?), dynamic (17/37 after 70 minutes?) 

* the  *paho-mqtt-v2-support* branch passes static (35/35), but nothing else: no_mirror 27/30, flakey 34/37 in 23 minutes, dynamic 16/37 in 10 minutes.

so it's not passing... but it's doing better than the main dev branch.


(in progress... posted but will update and results become available.)

petersilva:
updates:  no on *paho-mqtt-v2-support* passing: static, no-mirror, dynamic ... 
flakey_broker... getting 35/37 ... ran in 15 minutes...  so much better.

``` 
test 20 FAILURE: poll sftp_f62+3	 (1924) should have the same number of items of sender	 (1425)
test 21 FAILURE: poll sftp_f62+3	 (1419) should see the same number of items as poll sftp_f62 posted	 (1924)

```
but content of all the directories is correct... so looks like a test artifact...

petersilva:
restart_server still has lots of problems though 28/37

reidsunderland:
I hoped these two PRs would get flakey_broker working on GitHub: 
- https://github.com/MetPX/sarracenia/pull/1415
- https://github.com/MetPX/sr_insects/pull/68

[...]

PR #1119, opened 2024-06-22
Paho mqtt v2 support

petersilva:
This branch changes the mqtt driver to use version 2.1 of paho-mqtt python client... the current one (released in 2024.)
Previous versions have no way of passing the flakey_broker test, and will fail in other ways because there
it no application control of acknowledgements.  The patch I submitted to the paho-mqtt for application acknowledgements was only merged for v2.   V2 also brings many api changes, so it's not straight-forward.

The stack, as-is, passes more tests than the old mqtt support did on the old mqtt version, but not yet all tests.
It's still better than what is there, and will probably not come back to it for a while, so better not to have a long running parallel branch.

It will be a good basis for finishing the work later.

This is part of #693  ... it might actually solve flakey... but there might be some timing issues... (false positives, where the results seem ok,. but the test indicates a failure.)

### DIFF
--- a/sarracenia/__init__.py
+++ b/sarracenia/__init__.py
@@ -497,6 +497,54 @@ class Message(dict):
         for h in d:
             msg[h] = d[h]
 
+    def deriveSource(msg,o):
+        """
+           set msg['source'] field as appropriate for given message and options (o)
+        """
+        source=None
+        if 'source' in o:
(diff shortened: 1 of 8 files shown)


reidsunderland (approved):
discussed in our meeting

PR #1128, opened 2024-07-17
Issue33 - default queue name change.

### LINKED ISSUE #33: sr - add hostname on default queue name
pattern: 
queue_name q_${BROKER_USER}.${PROGRAM}.${CONFIG}.${HOSTNAME}

shows up in a lot of definitions in the DDSR cluster, where there are 8 worker nodes, and 1 broker.
...
I don't know about this. If you have multiple nodes and a shared file system, which is a pretty common case, putting the hostname in there will cause them to use separate queues. it isn't clear that it is smart to change that default... but there are specific cases where it is clearly desirable.

on ddsr clusters, where we have multiple nodes and NO shared disk, the hostname in queue names is clearly always necessary.
...
on DDSR now that each node has it's own broker, the HOSTNAME should not be in the queue names.
but it is still often useful.   It is a common thing, but whether it should be the default is a question... i dunno.
...
was: https://sourceforge.net/p/metpx/feature-requests/58/

petersilva:
fix #33 

* adds: *queueShare* setting, defaulting to ${USER}_${HOSTNAME}
* essentially *queueShare* should be used wherever one used to set *queueNames*... should be no more need for queueName, and more often than before, the default should be good enough so *queueName* setting can just be removed.
* changes default queue to end in *queueShare*, replacing the random-ids.
* add RAND8 variable for people to use queueShare to get the old behaviour back by putting *queueShare ${RAND8}_${RAND8}* in their configuration.
* documenting above.

### DIFF
--- a/docs/source/Explanation/CommandLineGuide.rst
+++ b/docs/source/Explanation/CommandLineGuide.rst
@@ -658,7 +658,7 @@ Creating the Queue
 Once connected to an AMQP broker, the user needs to create a queue.
 Common settings for the queue on broker :
 
-- **queueShare <strin>         (default: ${USER}_${HOSTNAM})**
+- **queueShare <strin>         (default: ${USER}_${HOSTNAME}_${RAND8})**
 - **expire        <duration>      (default: 5m  == five minutes. RECOMMEND OVERRIDING)**
 - **message_ttl   <duration>      (default: None)**
 - **prefetch      <N>            (default: 1)**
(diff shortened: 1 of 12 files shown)


petersilva:
after discussion adding _${RAND8} to default value.

petersilva:
OK... it looks good. 
*  as per discussion: *queueShare* defaults to ${USER}\_${HOSTNAME}\_${RAND8}, 
* separator is period.
* documentation in both languages consistent with that.
* documentation modified to promote use of queueShare, where queueName would have been used before.
* unit tests updated, so they pass.

junhu3 (approved):
Finally!

Issue #1043, opened 2024-05-10
GeoJSON filter after_accept plugin


reidsunderland:
We already support including GeoJSON in messages. 

One poll plugin I wrote adds GeoJSON information, because I noticed the API provided it, and I thought it would be interesting to use it. https://github.com/MetPX/sarracenia/blob/development/sarracenia/flowcb/poll/eumetsat.py

Now we need an after_accept plugin that can filter based on the GeoJSON in the message.

i.e. if a region defined in a config file overlaps with the region in the message, accept it. Otherwise, reject the message and don't bother downloading the file.

This after_accept plugin could be used either in a poll (to prevent messages from being posted) or in a sarra/subscribe (to filter messages before downloading).

There are multiple non-Sarracenia polls in use at CIS that do region filtering and only download files that cover the region they want. Implementing this plugin would be one step in the process of replacing these custom polls with sr3.

petersilva:
That´s a great idea.  My guess, from googline a bit, is that the fully general answer is to use something like geopandas. 

* https://stackoverflow.com/questions/75072717/calculating-overlap-of-polygons-using-python-unsure-of-which-package-to-use
* https://geopandas.org/en/stable/docs/user_guide/set_operations.html

There are pointers to *shapely* also...  but that seems more geometry and less geography focused.

perhaps a lighter, simpler one is turfpy:

* https://turfpy.readthedocs.io/en/latest/index.html

but not sure how active it is.

I´m wondering how complex the region specifications are from ice... if they are just 2D rectangles or circles, we can likely do  something manually, easily.  If we just want to compare an arbitrary geometry specification, then I think we will quickly end up using something like GeoPandas or turfpy.

It´s a big library, big dependency... fine if we need it, but heavy if our needs are simpler.

petersilva:
how to the CIS polls do the region filtering? The client is likely more versed in python + GIS than we are.
We should at least study what they have done.

petersilva:
I could imagine something dead simple like:
https://turfpy.readthedocs.io/en/latest/measurements/distance.html
```
from turfpy import measurement
from geojson import Point, Feature
interested_in = Feature(geometry=Point((-75.343, 39.984)))
product_covers = Feature(geometry=Point((-75.534, 39.123)))

if measurement.distance(interested_in,product_covers) == 0:
     accept
else:
    reject

``` 

That might be enough for a proof of concept.

reidsunderland:
[...]

PR #1130, opened 2024-07-18
Issue1043 geojson

Jeanc16888:
### Fix Issue 1043
This PR has the goal of fixing issue #1043. This is done by adding a filter for whether two features overlap (be that a polygon or a circle)

**Changes**
- Added GeoJSON filter file
- Added GeoJSON filter testing file
- Added GeoJSON markings in the config

**What is expected input**
- The message should be a dictionary which contains a field named "*geometry*" which is filled with a string or a series of string containing a standard GeoJSON like this: 
```Python
"geometry": '{"type": "Polygon", "coordinates": [[[-75.833, 39.284], [-75.332, 39.284], [-75.332, 40.584], [-75.833, 40.584], [-75.833, 39.284]]]}'
```
The message can contain only a GeoJSON Point or a Polygon. Refer to the GeoJSON documentation for this.

- The config should contain the same thing: an option called self.o.geometry (list of strings) and optionally (if the config feature is a Point), a self.o.geometry_maxDistance (in km). Which will look like this:
```Python
     options.geometry = [
        '{"type": "Polygon",',
        ' "coordinates": [',
        '    [',
        '        [-10.0, -10.0],',
        '        [10.0, -10.0],',
        '        [10.0, 10.0],',
        '        [-10.0, -10.0]',
        '    ]',
        ' ]',
        '}'
        ]
    options.geometry_maxDistance = 1.5
```

**Usage, logic and output**

- The GeoJSON file is a class and so calling the after_accept function like the standards that were set in Sarracenia is the way to go.
- The plugin tests whether there is overlap between the features. So in the case of two polygons (one in the message, one in the config), it tests whether they overlap. For one Point and one Polygon, it tests if the Point is within the Polygon. If there is two points being tested, it checks if the point in the message is within a certain radius (maxDistance) set in the config of the other point.
- There is error settings (outputs failed) for the input message being a wrong geometry (other than Polygon or Point), bad geometry set (a point with two coordinates etc.) or no geometry set.
- If there is no overlap or the points are not within a certain distance, the plugin outputs the message as being rejected.
- If there is overlap or the points are within a certain distance of each of other, the output is set as accepted.

**Testing**
A file for testing all possible combinations was designed. Be that:
- Testing when the config is a Polygon (messages are points or polygons).
- Testing when the config is a Point (with MaxDistance set) with messages being a Point or a Polygon.
- Testing an invalid incoming message geometry.
- Testing when comparing two points with missing MaxDistance.
- Testing where the message geometry field is an invalid JSON.
- Testing when the config does not contain the geometry option at all.

**Possible future additions if needed**
- Adding a bool in the config that checks if we accept the boundaries of a features (so if the point or other polygons is only touching the boundary of the radius / maxDistance or the boundary of the Polygon, it could be refused is that bool is set to true. Right now, it accepts as a default.
- Other features can be added like LineStrings, multiPoint or multiPolygon but the use would be greatly niche.
- Adding a bool which makes so that the plugin checks whether there is overlap on the bbox of a Polygon within the config, not the polygon itself. The use could maybe exist here but that is uncertain.


petersilva (commented):
Visually reviewing, it looks like it should work. Great feature to have. we have long wanted it. This is going to be great!

I think using the "GeoJSON as the Class/file name is misleading... it implies that we process GeoJSON files... GeoJSON is certainly the inspiration for this logic, and extracting the *geometry* field from a GeoJSON field is a great way to get geometry fields for use by this module.   

There is also a short-hand convention that means that we prefer for module names to be named with a single capital letter and the rest lower case.  If you don't follow this convention then people have to put:
```

callback sarracenia.flowcb.filter.geojson.GeoJSON  

``` 
when they want ot use the module.  I suggest naming the class *Geometry*, and the file *geometry.py* as the filtering is based on the *geometry* field of GeoJSON, rather than the entire file format (there is actually a lot more to those files than just this.)  Also such a name scheme would mean users can invoke it with:
```

callback filter.geometry

```
which is more compact and eloquent.





Regarding dependency management.  It's great you  found the feature detection stuff.
Can you move the feature detection into the module itself? Example for another plugin:

https://github.com/MetPX/sarracenia/blob/development/sarracenia/flowcb/clamav.py#L24-L38

The idea is that a user can do *sr3 features* and see a list of what features/support is active in the sr3 installation.   for example, if clamav packages are not installed, the clamav callback will not work.

but if geometry is not active (via callback), then the feature will not be listed.


```

fractal% sr3 features

Status:    feature:   python imports:      Description: 
Installed  amqp       amqp                 can connect to rabbitmq brokers
Absent     azurestorage azure-storage-blob   cannot connect natively to Azure Stoarge accounts
Installed  appdirs    appdirs              place configuration and state files appropriately for platform (windows/mac/linux)
Installed  filetypes  magic                able to set content headers
Absent     ftppoll    dateparser,pytz      not able to poll with ftp
Installed  humanize   humanize,humanfriendly humans numbers that are easier to read.
Installed  mqtt       paho.mqtt.client     can connect to mqtt brokers
Installed  process    psutil               can monitor, start, stop processes:  Sr3 CLI should basically work
Installed  reassembly flufl.lock           can reassemble block segmented files transferred
Installed  redis      redis,redis_lock     can use redis implementations of retry and nodupe
Installed  retry      jsonpickle           can write messages to local queues to retry failed publishes/sends/downloads
Installed  s3         boto3                able to connect natively to S3-compatible locations (AWS S3, Minio, etc..)
Installed  sftp       paramiko             can use sftp or ssh based services
Installed  vip        netifaces            able to use the vip option for high availability clustering
Installed  watch      watchdog             watch directories
Installed  xattr      xattr                will store file metadata in extended attributes

 state dir: /home/peter/.cache/sr3 
 config dir: /home/peter/.config/sr3 

```

so uh... normally we approve stuff unless it breaks something... but I figured getting these things patched up might be simpler than asking for another PR.   If you want us to approve this anyways, give another nudge.

petersilva:
This looks like it should work great! @reidsunderland do you want to have a look too? thoughts?

petersilva (approved):
looks great! Thanks!

PR #1134, opened 2024-07-24
Fix #1133

### LINKED ISSUE #1133: NASA Earthdata token expiry bug
The NASA Earthdata bearer token isn't auto-updating.

We think this is because the token expires on the day of the expiry at 23:59:59Z. 
 
The current code doesn't account for the expiry date being after today's date.

https://github.com/MetPX/sarracenia/blob/1533f85ea8cf2c23a5d93b419db43e74c09e9cfe/sarracenia/flowcb/authenticate/nasa_earthdata.py#L97-L102

Need to change `==` to `>=` on line 97 to account for this

andreleblanc11:


reidsunderland:
i did it a different way in #1142 , but that's a bigger change so we could merge this in for 2.00.54rc3 as a temporary fix.

reidsunderland (approved):

PR #1135, opened 2024-07-24
fix #1089 - created getIDStr function for Sarracenia.Message

### LINKED ISSUE #1089: fileAgeMax doesn't work in all situations?

petersilva:
close #1089 

Logic already existed under flowcb/log.py, just promoted it to the Message class itself. This provides an identifying string that deals with messages when fields are missing still giving something.

### DIFF
--- a/sarracenia/__init__.py
+++ b/sarracenia/__init__.py
@@ -774,6 +774,27 @@ class Message(dict):
 
         return sarracenia.Message.fromFileData(path, o, stat(path))
 
+    def getIDStr(msg) -> str:
+        """
+           return some descriptive tag string to identify the message being processed.
+
+        """
+        s=""
+        if 'baseURl' in msg:
+            s+=msg['baseUrl']+' '
+        if 'relPath' in msg:
+            if s and s[-1] != '/':
+                s+='/'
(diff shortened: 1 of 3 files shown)


reidsunderland (approved):

sarracenia/__init__.py
```diff
@@ -774,6 +774,27 @@ def fromStream(path, o, data=None):
 
         return sarracenia.Message.fromFileData(path, o, stat(path))
 
+    def getIDStr(msg) -> str:
+        """
+           return some descriptive tag string to identify the message being processed.
+
+        """
+        s=""
+        if 'baseURl' in msg:
```
andreleblanc11, line 783: Typo?

PR #1136, opened 2024-07-24
Release doc touchups

petersilva:
notes from Monday, when people were working on a release.


andreleblanc11 (approved):
More documentation is always good. 
We will improve it even more next pre-release :)

PR #1142, opened 2024-07-26
Authentication plugin refactor

### LINKED ISSUE #1133: NASA Earthdata token expiry bug

reidsunderland:
Adds a BearerToken class, with a standardized after_accept method that calls get_token (to be implemented by subclasses) and inserts the token into the in memory credentials DB.  This eliminates duplicated code that I had in my bearer token authentication plugins, and makes it easier for other people to write new authentication plugins. (Part of #565)

Refactored the NASA, Copernicus and Eumetsat auth plugins to use this BearerToken class as their parent class.

Also fixes #1133 (in a different way than #1134, sorry @andreleblanc11 😄 I was worried about string comparisons for dates not being reliable, especially with the MM/DD/YYYY format ).

I'm okay with waiting until after we release a stable version of 3.00.54 to merge this (i.e. only include it in 3.00.55).

### DIFF
--- a/sarracenia/examples/subscribe/get_copernicus.conf
+++ b/sarracenia/examples/subscribe/get_copernicus.conf
@@ -1,7 +1,7 @@
 broker amqp://COPERNICUS@localhost/
 exchange xs_COPERNICUS
 
-callback accept.auth_copernicus
+callback authenticate.copernicus
 
 directory /tmp/copernicus
 accept .*L2__O3.*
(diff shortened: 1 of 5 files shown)


petersilva (approved):
great! Looks like it will be easier to understand going forward.

petersilva:
@reidsunderland... there is a conflict  ... can you resolve it in the branch before we merge?

reidsunderland:
I fixed the conflict

Issue #1144, opened 2024-08-02
v2 sr sanity crashes when a pid file exists but the config doesn't


reidsunderland:
I marked this v2 only, but haven't yet confirmed that sr3 handles this situation any better.

When a pid file exists in the .cache directory, but the corresponding config file has been removed, sr sanity crashes.

We're observing it on 2.21.4 but I don't see any changes in the related code, so I don't think it was fixed.

```
sanity: missing: [['sender', 'configname', 1]]
starting them up...
Traceback (most recent call last):
  File "/usr/bin/sr", line 11, in <module>
    load_entry_point('metpx-sarracenia==2.21.4', 'console_scripts', 'sr')()
  File "/usr/lib/python3/dist-packages/sarra/sr.py", line 1094, in main
    gs.sanity()
  File "/usr/lib/python3/dist-packages/sarra/sr.py", line 766, in sanity
    self._start_missing()
  File "/usr/lib/python3/dist-packages/sarra/sr.py", line 728, in _start_missing
    self._launch_instance(component_path, c, cfg, i)
  File "/usr/lib/python3/dist-packages/sarra/sr.py", line 90, in _launch_instance
    s = self.configs[c][cfg]['options'].statehost.lower()
KeyError: 'configname'
```

reidsunderland:
I reproduced the bug on the v2_dev branch, so it wasn't fixed.

reidsunderland:
with my fix, we get this:

```
sanity: missing: [['sarra', 'get_copernicus', 1]]
starting them up...
sarra get_copernicus is not a known config, cannot start it. Maybe the config file is missing?
killing strays...
```
$? = 0 

instead of 

```
sanity: missing: [['sarra', 'get_copernicus', 1]]
starting them up...
Traceback (most recent call last):
  File "/net/local/home/sunderlandr/.local/bin/sr", line 33, in <module>
    sys.exit(load_entry_point('metpx-sarracenia', 'console_scripts', 'sr')())
  File "/net/local/home/sunderlandr/sarra_v2/sarra/sr.py", line 1106, in main
    gs.sanity()
  File "/net/local/home/sunderlandr/sarra_v2/sarra/sr.py", line 778, in sanity
    self._start_missing()
  File "/net/local/home/sunderlandr/sarra_v2/sarra/sr.py", line 740, in _start_missing
    self._launch_instance(component_path, c, cfg, i)
[...]

PR #1145, opened 2024-08-02
Fix for #1144 - don't crash when the config doesn't exist for a missing instance

reidsunderland:
This prevents sr sanity from crashing, it just ignores the problem and moves on.

It might be better to have sr sanity/sr audit delete the pid file in this case, but that would be a bit more complicated to implement and test.


petersilva:
I looked at the failures in the flow tests... It's weird that the dynamic is failing... but I looked back, and it's been failing for a year.  It would be a good thing to fix that, but not a reason not to merge this change.

petersilva (approved):

PR #1149, opened 2024-08-06
Issue339 part2 queue names

petersilva:
This is the second phase of work in issue339, being able to gather messages from multiple sources.  It adds all of the queue settings into the binding structure.

Bindings used to have three elements: exchange, topicPrefix, subtopic, but now they need:
broker, queueName, and all the settings related to declaring and binding a queue... so: expiry, durable, queueBind, queueDeclare, auto_delete, prefetch....

All of those settings become binding/consumer specific.

so that is too much to keep track of in a tuple, so changed the binding into a dictionary.

This is mean to be merged after part1 is merged... the patches in part1 break flow tests, so they need to be collapsed into a single patch.  In contrast, these patches all preserve full flow_test compatibility, so a merge is fine (desirable actually to have smaller patches to work on bisection.)

The feature still isn't finished though... need to replace .qname files with .binding files, and figure out how to read them in.... there will beed to be a part 3.

### DIFF
--- a/sarracenia/config.py
+++ b/sarracenia/config.py
@@ -650,7 +650,7 @@ class Config:
          cfg.component = 'subscribe'
          cfg.config = 'flow_demo'
          cfg.action = 'start'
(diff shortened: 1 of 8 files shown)


petersilva:
closing to  to feature branch.

PR #1150, opened 2024-08-06
Add new BC Trans API to sr3

andreleblanc11:
* This plugin already exists (on gitlab.science), but it wasn't using the bearer token authentication class.

* Now we are using it, which enables us to have the API working just through the authenticate plugin.

* We also have another after_accept entry point to rename the file in its correct CSV format.

* Examples also added in the examples folder

### DIFF
--- a/sarracenia/examples/flow/bc_trans.conf
+++ b/sarracenia/examples/flow/bc_trans.conf
@@ -1,0 +1,27 @@
+#
+# Post the BC Trans URL to the sarra.
+# The sarra will authenticate through the authenticate plugin after
+#
+
+vip # Enter your vip here
+
+# You want to callback the scheduled flow AND the post plugin so the URL gets posted downstream
+callback scheduled
+callback post.message
+
+
+# Run everyhour
+scheduled_interval 3600
+
+# post_baseUrl + path combined make up the URL to be posted downstream
+post_baseUrl https://sawsx.api.gov.bc.ca/
+path /api/v1/motisite/report7100/all/3/
+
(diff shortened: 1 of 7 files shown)


andreleblanc11:
Unit tests failing looks unrelated

```
___________ ERROR collecting sarracenia/flowcb/poll/eumetsat_test.py ___________
import file mismatch:
imported module 'eumetsat_test' has this __file__ attribute:
  /home/runner/work/sarracenia/sarracenia/tests/sarracenia/flowcb/authenticate/eumetsat_test.py
which is not the same as the test file we want to collect:
  /home/runner/work/sarracenia/sarracenia/tests/sarracenia/flowcb/poll/eumetsat_test.py
HINT: remove __pycache__ / .pyc files and/or use a unique basename for your test file modules
```

reidsunderland:
Yeah, I guess I broke that and didn't notice :(

reidsunderland (approved):
It looks good to me.

The new_messages list in the renamer after_accept isn't really needed, but isn't hurting anything. You only need to create a new list if you are rejecting messages -- then you add messages to worklist.rejected and replace the old incoming list with the new one `worklist.incoming = new_messages`.

If you aren't rejecting, then you can just iterate through worklist.incoming and modify the message in place (that's what you're doing anyways), like this one: https://github.com/MetPX/sarracenia/blob/development/sarracenia/flowcb/accept/httptohttps.py

reidsunderland:
I guess I can't reapprove because I didn't request changes in the previous review. it looks good to me

PR #1152, opened 2024-08-07
Fix #1151

### LINKED ISSUE #1151: `ComputeIdentity` crashing instance when xattr identity not found
The instance will crash when `xattr` can't find the `identity` when it performs a `listxattr`. `identity` will be `null` value

```
  File "/net/local/home/leblanca/.local/lib/python3.10/site-packages/sarracenia/instance.py", line 249, in <module>
    i.start()
  File "/net/local/home/leblanca/.local/lib/python3.10/site-packages/sarracenia/instance.py", line 240, in start
    self.running_instance.run()
  File "/net/local/home/leblanca/.local/lib/python3.10/site-packages/sarracenia/flow/__init__.py", line 600, in run
    self.work()
  File "/net/local/home/leblanca/.local/lib/python3.10/site-packages/sarracenia/flow/__init__.py", line 1228, in work
    self._runCallbacksWorklist('after_work')
  File "/net/local/home/leblanca/.local/lib/python3.10/site-packages/sarracenia/flow/__init__.py", line 300, in _runCallbacksWorklist
    p(self.worklist)
  File "/net/local/home/leblanca/.config/sr3/plugins/rename/sumstr_ext.py", line 65, in after_work
    msg.computeIdentity(old_path,self.o)
  File "/net/local/home/leblanca/.local/lib/python3.10/site-packages/sarracenia/__init__.py", line 431, in computeIdentity
    if fxainteg['method'] == o.identity_method:
TypeError: string indices must be integers

```

https://github.com/MetPX/sarracenia/blob/7c6313cc60da9f9765ca1d16abe21fae03025fbd/sarracenia/filemetadata.py#L137-L151
https://github.com/MetPX/sarracenia/blob/7c6313cc60da9f9765ca1d16abe21fae03025fbd/sarracenia/__init__.py#L416-L435

andreleblanc11:
When `identity` is missing from `xattr` it returns a `null` string

### DIFF
--- a/sarracenia/__init__.py
+++ b/sarracenia/__init__.py
@@ -425,9 +425,10 @@ class Message(dict):
                 if xattr.get('mtime') >= msg['mtime']:
                     logger.debug("mtime remembered by xattr")
                     fxainteg = xattr.get('identity')
-                    if fxainteg['method'] == o.identity_method: 
-                         msg['identity'] = fxainteg
-                         return
+                    if fxainteg != 'null':
+                        if fxainteg['method'] == o.identity_method: 
+                            msg['identity'] = fxainteg
+                            return
                     logger.debug("xattr different method than on disk")
                     calc_method = o.identity_method
                 else:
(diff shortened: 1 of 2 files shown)


petersilva (changes requested):
this patch does not make sense.  why would a routine return  'null' as a string?  looking at the source of the other routine, it returns either the value or python None.

This code effectively disables every reading the extended attributes because the return value never == 'null'.

petersilva (approved):
excellent patch, addresses the root cause ... invalid checksum for calculate on download case.

PR #1160, opened 2024-08-09
Fix #953 re-enable and fix passive queue declare

### LINKED ISSUE #953: sr3: PRECONDITION_FAILED - unknown delivery tag
```
2024-02-07 16:27:36,038 [WARNING] sarracenia.moth.amqp getNewMessage failed q_MSC-SCI-CMC.winnow.winnow-ss5-09.smco501: Basic.ack: (406) PRECONDITION_FAILED - unknown delivery tag 3
2024-02-07 16:27:36,038 [WARNING] sarracenia.moth.amqp getNewMessage lost connection to broker

```

On HPC mirroring use case, versions >= 3.0.47 are seeing this error in one particular mirror. Error goes away when client downgrades, and re-appears on all versions above the threshold.

reidsunderland:
Re-enables the passive queue declare that we use to get the number of messages in the queue.

#953 is fixed by only calling self.__connect from _queueDeclare when it's necessary. If we're already connected, connecting again will cause problems.

### DIFF
--- a/sarracenia/moth/amqp.py
+++ b/sarracenia/moth/amqp.py
@@ -242,7 +242,7 @@ class AMQP(Moth):
             next_time = self.last_qDeclare + 30
             now=time.time()
             if next_time <= now:
-                #self._queueDeclare(passive=True)
+                self._queueDeclare(passive=True)
                 self.last_qDeclare=now
 
         super().metricsReport()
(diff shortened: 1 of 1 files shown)


petersilva (approved):
looks very likely correct. I hope this fixes things... doing it early in the release cycle (for after 54) will give us time to validate that it's really solid.  The old one seemed fine in the flow tests, but blew up in use.

PR #1162, opened 2024-08-09
Issue1159

### LINKED ISSUE #1159: more obvious message when wrong config pattern given.
```

sr3 start sender/sender_hpfx-collab-wxo-dd-direct (should have been sender/hpfx-collab-wxo-dd-direct)
starting:( 0 ) Done

```
So it got the list of configs from the command line, and they did not match any existing flow configurations, so it says it started 0 things... technically accurate... but kind of uninformative at the same time.

It would be a lot more useful if it gave some kind error feedback to say it did not find any matching configs.

petersilva:
close #1159 

give the kinds of error messages normal people expect when invalid configurations (usualyl typos) are given on the command line.  If given a configuration that does not match anything... be conservative and do nothing.

Note, the flow tests currently do operations on non-existent/invalid configurations, so they need a PR ...
this one: https://github.com/MetPX/sr_insects/pull/54

in order to pass (the flow tests when automatically invoked below will fail because the above sr_insects PR isn't merged.)

### DIFF
--- a/sarracenia/config.py
+++ b/sarracenia/config.py
@@ -2055,7 +2055,7 @@ class Config:
                     self.sleep=1
 
         if self.runStateThreshold_hung < self.housekeeping:
-            logger.warning( f"{component}/{config} runStateThreshold_hung {self.runStateThreshold_hung} set lower than housekeeping {self.housekeeping}. sr3 sanity might think this flow is hung kill it too quickly.")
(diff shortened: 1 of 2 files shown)


petersilva:
ohh... it's only the cleanups that fail... so the other PR is needed, to do cleanups.. but the flow tests do not check if the cleanup succeeds.  That's why stuff is passing.

reidsunderland (approved):

Issue #1157, opened 2024-08-08
Add stat call to transfer protocols, and logic to avoid useless retries when source changes.


petersilva:
Based on inconsistencies between messages and upstream servers caused by changes in the products and lag on the consumer side, it is relatively common to download a file and the result to have a different size than what is expected based on the message. Current logic tries to download again until size matches (unless AcceptSizeWrong is set), and failing a few attempts puts the message on the retry queue for a subsequent attempt (after 5 minutes, typically.)

It might be better if, when it downloads the file, rather than blindly trying the download again, it would check with the server if everybody agrees on the size. Essentially, it should do a stat() call... the implementation of stat would vary per transfer protocol.

* get_http_headers() could be used to implement stat() for https.
* ls could do it for SFTP and perhaps FTP.
* actual os.stat() would work for file protocol.

So you do a stat() of the upstream resource. if the stat succeeds and if its size does not match the message. then retrying the download will never succeed. so should give up, rather than retry.

observations:

* if we give up, and the subscriber is always far behind the source, then the subscriber will never get any data at all... kind of a resource starvation thing.

* if you try to download something, then put it on the retry queue, then get a new message... that works... but a minutes later you pull the same resource out of the retry queue... it will fail again (because the newer version is already downloaded.) ...  logically... we should be updating the retry queue with subsequent successes.

petersilva:
We have one product stream (citypage_weather) where the files are re-written constantly, like three or four times within 30 seconds... and then are static for an hour or so... but then get re-written multiple times in a short interval.

So this creates a stream where you have four messages of updates for the same "file" and if the consumers are more than a 30 seconds behind the source, then by the time they try to download, they will get a newer file than what is expected, and then... how to deal with that.

petersilva:
It would appear that the most robust answer is "Don't do that" ... put a robust date stamp in the name the file to differentiate it and have four products issued in a minute if you must, but don't re-publish the same product (with the same name) too quickly.

petersilva:
Here's a sample message:
```

2024-08-11 00:00:47,978 [INFO] sarracenia.flow download upstream resource is newer, so message https://hpfx.collab.science.gc.ca //20240811/WXO-DD/citypage_weather/xml/NB/s0000653_e.xml is obsolete. Discarding.

```
if it's not newer, but the size does not match:

```

[...]

PR #1163, opened 2024-08-13
Issue1157 add stat so subscribe/citypage downloads just work.

petersilva:
progress on #1157 ...

* add a stat() entry point to the transfer protocols.
* implemented using @reidsunderland 's and @andreleblanc11 's logic to fetch headers for https.
* tested with subscribe/hpfx_amis

so before this change:

* when downloading citypages, typically at the top of the hour, a subscriber would download a citypage and find it didn't match what the message said it should be.
* so the downloaded file is discarded, and the message it placed on the retry queue.
* five minutes later It will again attempt to download.

in the case of citypages, the file has been changed before the subscriber attempted the first download:
*  so all retries will fail.... for three days (by default retries continue for three days.)

with the change:

* after a length mismatch on download, stat() the file on the server. to ask it what the size is on the server.

* if the date of the file is newer than on the message, discard the message rather than requeing... 
* If the size matches on the server... then downloading again, won't change anything. so better to discard.

in both cases, subsequent retries will never succeed.,,, so just throw it out.

* so there will be no error reported, the subscribe flow now understands what is going on and just prints and information message.
* nothing will be put on the retry queue for this understood phenomenon.  

So testing was done downloading subscribe/hpfx_citypage ... and with these changes, it is clean now, with no retries.

### DIFF
--- a/.github/workflows/flow.yml
+++ b/.github/workflows/flow.yml
@@ -32,7 +32,7 @@ jobs:
       fail-fast: false
       matrix:
        which_test: [ static_flow, no_mirror, flakey_broker, dynamic_flow, restart_server, partitioned_flow ]
-       osver: [ "ubuntu-20.04", "ubuntu-22.04" ]
+       osver: [ "ubuntu-20.04", "ubuntu-22.04", "ubuntu-24.04" ]
 
     runs-on: ${{ matrix.osver }}
   
--- a/.github/workflows/flow_basic.yml
+++ b/.github/workflows/flow_basic.yml
@@ -26,7 +26,7 @@ jobs:
       # Don't cancel the entire matrix when one job fails
       fail-fast: false
       matrix:
-       osver: [ "ubuntu-20.04", "ubuntu-22.04" ]
+       osver: [ "ubuntu-20.04", "ubuntu-22.04", "ubuntu-24.04" ]
 
     runs-on: ${{ matrix.osver }}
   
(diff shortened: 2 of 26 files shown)


petersilva:
Also added fileSizeMax option to be able to avoid downloading files that are too much bigger than accepted.

petersilva:
something weird has happenned to ubuntu 20.04... all tests on that OS are failing with some kind of SSL error.
I don't think it's related to this PR... on 22.04, all tests pass.

petersilva:
otoh, the problem seems to occur when running sr3 entry point and it does an import requests (which is added by this PR) it seems to fail with an inscrutable SSL something or other not found... so it's like the python packages being installed in 20.04 are somehow incompatible with eachother.

petersilva:
kind of frustrating... we have test machines, and one runs 18.04... and all the tests pass when I run them on a vm... tested so far on 24.04 and 18.04... hmm...

petersilva:
weirdnesses so far:
* perhaps python3-requests is a new dependency that must be added (I think it is pulled in by other deps anyways... I did not have to install it, was there on my test machines.)
* the metpx-sr3c packages might be a little janky on someOS's... it seems to be looking for the wrong versions in the flow tests... I built a local metpx-sr3c package to get it to pass on 18.04.

reidsunderland:
We could change the code to use urllib instead. Requests is just a little bit nicer to work with.

https://docs.python.org/3/library/urllib.request.html#urllib.request.Request --> can set `method='HEAD'`.

petersilva:
I'm going to close this PR, merge from development to the branch, and try opening the PR again.

petersilva:
re-submission behaves the same... it's something to do with the change... but no idea what.

petersilva:
The failure on gitlab of images on 20.04 where the SSL crashes prevent running any tests... we are considering re-writing the https stat() call to use urllib.request()

petersilva:
This branch will require work and cleaning... will open a different PR at some point.

PR #1174, opened 2024-08-19
Issue1155 AMQP ack refactor

reidsunderland:
AMQP: refuse to attempt to ack a message if it was received on a different channel, connection or broker (preventing unnecessary
connection tear down and re-establish)

AMQP: when ack is attempted and fails because the connection to the broker is broken, cleanup (close) the broken connection and don't bother attempting to retry the ack

Also changed the ack method to return a boolean, which wasn't needed but could possibly be useful someday if we want to change how we handle a failed ack.


petersilva (approved):

petersilva:
The tests are messed up atm... but I think this change is ok anyways... in an ideal world we would fix the tests... but no clue what is with 20.04.

PR #1175, opened 2024-08-19
fix #1166 improved download decision criteria

### LINKED ISSUE #1166: file_should_be_downloaded checks size before checksum... hmm...
Some identity algorithms are built to identify files that have the same meaning, but different byte streams. It's a domain specific property.  In such cases, the size will be different, but the checksum could be the same.

If the checksum is the same, one should not download.   The routine in flow/__init__.py  def file_should_be_downloaded() checks the file size first and returns True if the size is different... 

That's probably wrong when certain identity algorithms are in use.

petersilva:
re-organized the file_should_be_download routine a bit, now only check size and mtime if using an unknown or binary checksum where those characteristics would be definitive.

so now it will not check size for arbitrary, or other checksums where it is irrelevant....
When adding methods to identity, can update the binary_methods list.

### DIFF
--- a/sarracenia/flow/__init__.py
+++ b/sarracenia/flow/__init__.py
@@ -1457,7 +1457,7 @@ class Flow:
             method='unknown'
         
         # if no method given, then assume binary comparison is good.
-        if method in ['sha512', 'md5', 'unknown']: 
+        if method in sarracenia.identity.binary_methods: 
             if 'size' in msg:
                 end = msg['size']
                 # compare sizes... if (sr_subscribe is downloading partitions into taget file) and (target_file isn't fully done)
--- a/sarracenia/flow/__init__.py
+++ b/sarracenia/flow/__init__.py
@@ -1451,7 +1451,7 @@ class Flow:
         #   part of partitioning deferral.
         #end   = self.local_offset + self.length
         # if using a true binary checksum, the size check is enough.
-        if 'identidy' in msg and 'method' in msg['identity']:
+        if 'identity' in msg and 'method' in msg['identity']:
             method=msg['identity']['method']
         else:
             method='unknown'
--- a/sarracenia/flow/__init__.py
+++ b/sarracenia/flow/__init__.py
@@ -1451,8 +1451,8 @@ class Flow:
         #   part of partitioning deferral.
         #end   = self.local_offset + self.length
         # if using a true binary checksum, the size check is enough.
-        if 'identify' in msg and 'method' in msg['identity']:
-            method=msg['identity']
+        if 'identidy' in msg and 'method' in msg['identity']:
+            method=msg['identity']['method']
         else:
             method='unknown'
         
--- a/sarracenia/flow/__init__.py
+++ b/sarracenia/flow/__init__.py
@@ -1450,58 +1450,62 @@ class Flow:
         # FIXME... local_offset... offset within the local file... partitioned... who knows?
         #   part of partitioning deferral.
         #end   = self.local_offset + self.length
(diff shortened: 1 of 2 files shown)


petersilva:
one thing that isn't clear atm... if a file:
* has no checksum
* has the same size in the message as the one already on disk...

we currently assume it is different, since that is safest... we don't have a checksum to know it is the same.
It is debatable that without a checksum... do we just assume a file of the same size is... the same file?
I think what we currently do is safer... but likely wasteful... but data sources should really provide checksums.

sarracenia/flow/__init__.py
```diff
@@ -1450,58 +1450,62 @@ def file_should_be_downloaded(self, msg) -> bool:
...
-            # This check prevents random halting of subscriber (inplace on) if the messages come in non-sequential order
-            # target_file is the same as new_file unless the file is partitioned.
-            # FIXME If the file is partitioned, then it is the new_file with a partition suffix.
-            #if ('self.target_file == msg['new_file'] ) and ( fsiz != msg['size'] ):
-            if (fsiz != msg['size']):
-                logger.debug("%s file size different, so cannot be the same" %
-                             (msg['new_path']))
-                return True
+        # if using a true binary checksum, the size check is enough.
+        if 'identify' in msg and 'method' in msg['identity']:
```
reidsunderland: I think `if 'identify'` is a typo and should be identity

sarracenia/flow/__init__.py
```diff
@@ -1450,58 +1450,62 @@ def file_should_be_downloaded(self, msg) -> bool:
...
+            method=msg['identity']
```
reidsunderland: I'm also not completely sure if this is right? You're checking that `'method' in msg['identity']` but assigning `method=msg['identity']` not `method=msg['identity']['method']`?

reidsunderland (changes requested):
This is a draft so maybe I shouldn't be looking at it yet, but I noticed a couple things that might be preventing it from working as intended

sarracenia/flow/__init__.py
petersilva: totally right... good eye.

petersilva (commented):
look good...

petersilva:
darn... I committed the same changes to the branch... not realizing you had queued them up also... unclear how to sort this out.

reidsunderland:
I didn't make any changes to this branch, unless i did it by accident 😄

sarracenia/flow/__init__.py
```diff
@@ -1450,58 +1450,62 @@ def file_should_be_downloaded(self, msg) -> bool:
...
-            # This check prevents random halting of subscriber (inplace on) if the messages come in non-sequential order
-            # target_file is the same as new_file unless the file is partitioned.
-            # FIXME If the file is partitioned, then it is the new_file with a partition suffix.
-            #if ('self.target_file == msg['new_file'] ) and ( fsiz != msg['size'] ):
-            if (fsiz != msg['size']):
-                logger.debug("%s file size different, so cannot be the same" %
-                             (msg['new_path']))
-                return True
+        # if using a true binary checksum, the size check is enough.
+        if 'identidy' in msg and 'method' in msg['identity']:
```
reidsunderland: I think there's a new typo? `identify` got changed to `identidy` 😆

reidsunderland (changes requested):

sarracenia/flow/__init__.py
petersilva: offs

reidsunderland (approved):

PR #1182, opened 2024-08-22
V2 fix the problems found during v2 upgrade of main pump.

### LINKED ISSUE #1176: v2 poll adds trailing slash to post_base_url
Documenting what @petersilva found yesterday and caused @andreleblanc11 some problems.

https://github.com/MetPX/sarracenia/blob/fa78678fb34831957d092622e9aede3de8d30136/sarra/sr_poll.py#L131-L136

When you do not define a post_base_url explicitly in the poll config, v2 will set post_base_url to the value of destination, but it always adds a trailing slash. 

This causes sr3 consumers and recent v2 versions to treat all paths posted by v2 polls as absolute paths, even if you wanted them to be relative.

Recent v2 downloaders now look at the baseUrl in the message and prepend that to the relPath, so if the baseUrl ends in /, it will get added to the relPath that tries to download. This is the change that broke something that was working before, but we do not plan to undo this change. 

If you need to post messages with relative paths from v2 polls, you can workaround the problem by explicitly setting post_base_url in the poll config:

```
# relative 
destination      sftp://user@example.com
post_base_url    sftp://user@example.com

# Many possibilities for absolute... 

# absolute
destination      sftp://user@example.com/
# no post_base_url

# absolute
destination      sftp://user@example.com/
post_base_url    sftp://user@example.com/

# absolute - this is the case that really should be changed to post a baseUrl with no trailing slash
destination      sftp://user@example.com
# no post_base_url
```

We might want to change the behaviour by deleting line 133 above. It probably requires a manual audit of all our v2 polls, but we'll have to do the audit anyways if we want to upgrade to the current latest 2.24.08 version.

### LINKED ISSUE #1180: v2 poll `file_time_limit` doesn't work properly
We've found a problem when we're starting/stopping polls between sarracenia versions (2.21.04 and 2.24.08). The `ls` file in the poll cache doesn't recognize the new entries, before it gets the VIP back, so it will repost all the old files in the `ls` entry -> https://github.com/MetPX/sarracenia/issues/1184.

To mitigate this problem, we've found that we can specify a 'file age limit' within the poll, with `file_time_limit`.

https://github.com/MetPX/sarracenia/blob/bd095d86ece157e0fbd4a40b41ef1cace2dcec8b/sarra/sr_poll.py#L177-L189

When testing this config option on dev, I found that the `.seconds` method **ONLY** returns the seconds from the **CURRENT DAY** and not the total number of seconds. This means that

- If you have `file_time_limit Xd` (default is 60d) it should always return True and never filter out old files. 
- If you have `file_time_limit Xh or Xm or Xs` the files will be ignored for some parts of the day, but then unignored once the timestamp is reached again on the following day.
   - We currently don't have any configuration specifying the `file_time_limit` so this is the reason why it has never caused problems until now

```
>>> current_date
datetime.datetime(2024, 8, 22, 14, 51, 47, 640140)
>>> file_date
datetime.datetime(2024, 8, 15, 12, 40)
>>> abs((file_date - current_date)).days
7                                                                    <- The days field is separated
>>> abs((file_date - current_date)).seconds
7907                                                                 <- The seconds field is separated
```

The fix for this should be to pass `.total_seconds()` instead of `.seconds` to get the actual total number of seconds.

```
>>> abs((file_date - current_date)).days
7
>>> abs((file_date - current_date)).seconds
7907
>>> abs((file_date - current_date)).total_seconds()
612707.64014
```

petersilva:
close #1180 
close #1176 

based on attempts to deploy to main data pump.  Cleans up the issues we found.
It should probably go to dev, then pre-release, and gradually work it's way to ops.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,4 +1,10 @@
-metpx-sarracenia (2.24.08) UNRELEASED; urgency=medium
+metpx-sarracenia (2.24.08post1) UNRELEASED; urgency=medium
+
+  * 
+
+ -- SSC-5CD2310S60 <peter@bsqt.homeip.net>  Thu, 22 Aug 2024 12:25:11 -0400
+
+metpx-sarracenia (2.24.08) unstable; urgency=medium
 
   * Fix for #1144 - sr_audit don't crash when the config doesn't exist for a missing instance
 
--- a/sarra/__init__.py
+++ b/sarra/__init__.py
@@ -25,4 +25,4 @@
 #  Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA 02111-1307  USA
 #
 #
-__version__ = "2.24.08"
+__version__ = "2.24.08post1"
--- a/sarra/sr_poll.py
+++ b/sarra/sr_poll.py
@@ -179,7 +179,7 @@ class sr_poll(sr_post):
                 date = line_split[5] + " " + line_split[6]
                 file_date = parse(date)
                 current_date = datetime.datetime.now()
-                file_within_date_limit = abs((file_date - current_date).seconds) < self.file_time_limit
+                file_within_date_limit = abs((current_date - file_date).total_seconds()) < self.file_time_limit
             except:
                 self.logger.error("Assuming ok, incorrect date format for line:  %s" % ls[f])
                 if not dateparser_available:
--- a/sarra/sr_poll.py
+++ b/sarra/sr_poll.py
@@ -130,7 +130,8 @@ class sr_poll(sr_post):
 
         if self.post_base_url == None:
             self.post_base_url = self.details.url.geturl()
-            if self.post_base_url[-1] != '/': self.post_base_url += '/'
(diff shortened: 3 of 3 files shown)


reidsunderland:
I'm okay with it the way it is, but I think the absolute value could cause weird behaviour if the remote file's timestamp is significantly newer than the local system's current time?

petersilva:
> I'm okay with it the way it is, but I think the absolute value could cause weird behaviour if the remote file's timestamp is significantly newer than the local system's current time?

won't it just be a small negative number, the abs makes it psitive... either way it should not cause an issue unless the max_file_age is really short?

reidsunderland (approved):

reidsunderland:
yeah, i think it just means that a file with a newer timestamp than the current time on the node running sarra could get ignored as "too old" for a while (when it's actually too new). But yes, unlikely unless file_time_limit is << 24 hours.

andreleblanc11:
For canadian servers, they'll also likely be using local timezones instead of UTC, which would all be in the past in comparison to UTC.

PR #1186, opened 2024-08-23
fix Ubuntu 20.04 flow tests

reidsunderland:
pr created to trigger flow tests


reidsunderland:
Initial results are looking good, Ubuntu 20.04 tests are passing again. 

The change is that we're installing paramiko with apt instead of pip and that seems to prevent the version conflicts that were causing problems before. When installing paramiko with pip, we were also getting a much newer cryptography package that was incompatible with *something* else that was installed. 

We use system packages when possible, so this change is good, it makes the GitHub tests more similar to our real deployments.

petersilva (approved):

PR #1190, opened 2024-08-26
Issue1189 SFTP FSETATTTR fix + timeout 0 support

### LINKED ISSUE #1189: Can't send to SFTP destinations that don't support FSETSTAT
Sending to SFTP destinations that don't support FSETSTAT causes a failure:

```
[ERROR] sarracenia.flow send could not send /source/filename to inflight=None sftp://user@example.com/ /destination/filename: FSETSTAT unsupported
```

Looking at the Paramiko code revealed that FSETSTAT is used for `chmod`, `chown`, `utime` and `truncate`. 

This happens even with `timeCopy` and `permCopy` set to False, so we shouldn't be calling chmod, chown or utime.

We do call truncate in 2 places:

1) in `readlocal_write` of the Transfer class: https://github.com/MetPX/sarracenia/blob/f7563bc70c0805f18768298df5e6bef0e64a7856/sarracenia/transfer/__init__.py#L401
    - I think this is the case we're encountering, because length is set to 0. This case is for overwriting files that already exist. We want to ensure that remnants of the old file aren't left behind when the new file is smaller than the old one.
2) in the SFTP transfer class: https://github.com/MetPX/sarracenia/blob/f7563bc70c0805f18768298df5e6bef0e64a7856/sarracenia/transfer/sftp.py#L486

In the SFTP code, there's already logic to determine if a file already exists on the destination, so we could possibly do something with that to not call truncate in the case where the file doesn't already exist.

We could additionally add a try/except around the truncate call(s), but need to be careful to make sure we're not ignoring failures to truncate in cases where truncation really was needed (when a new file overwrites a larger old file).

The workaround is to use binary acceleration for all file transfers to these destinations.

petersilva:
close #1189 

These are changes that deal with sftp servers.
* wrap truncate/chmod/utime calls in sftp in try/excepts so that the transfers are not considered "failed" if one of those operations fails.
* report a warning about the call failing instead.
* add an nofsetstat option to the sr3_options configs, so that when we know that those operations won't work, don't even generate the warning message.   

passes flow tests, but No tested with a server that restricts the permissions.

@iMacadan  perhaps you can test this in sending to such a client.

### DIFF
--- a/docs/source/Reference/sr3_options.7.rst
+++ b/docs/source/Reference/sr3_options.7.rst
@@ -2035,11 +2035,12 @@ a topicPrefix, followed by subtopics derived from the *relPath* field of the mes
 Some networks may choose to use different topic conventions, external to sarracenia.
 
 
-timeout <interval> (default: 0)
+timeout <interval> (default: 300)
 -------------------------------
 
 The **timeout** option, sets the number of seconds to wait before aborting a
 connection or download transfer (applied per buffer during transfer).
+Setting to zero disables timeouts.
 
 
 timezone <string> (default: UTC)
--- a/docs/source/Reference/sr3_options.7.rst
+++ b/docs/source/Reference/sr3_options.7.rst
@@ -1240,6 +1240,18 @@ off and output with *post_exchangeSplit*, which route notification with the same
 the same member of a **second layer of subscribers (winnow) whose duplicate suppression caches 
 are active.**
 
+nofsetstat <off|on> (default: off)
+----------------------------------
+
+Error messages such as this::
+
+    [ERROR] sarracenia.flow send could not send /source/filename to inflight=None sftp://user@example.com/ /destination/filename: FSETSTAT unsupported
+
+Some SFTP servers will restrict permissions in ways which limit functionality. Cannot set modification times, or 
+truncate files (when new file is shorter than earlier version.), cannot set mode/permission bits.
+To avoid warnings or errors being generated about FSETSTAT, one can accept the limited functionality
+by setting **nofsetstat on**
+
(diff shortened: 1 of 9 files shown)


reidsunderland (commented):
I'm confused by the change to alarm_cancel?

The change adds `if time <= 0: return` which changes the code from cancelling the alarm to doing nothing? And if a time > 0 is passed in, it would cancel the previous alarm but set a new one that will cause a signal to be sent after _time_ seconds.


> signal.alarm(time)
> If time is non-zero, this function requests that a [SIGALRM](https://docs.python.org/3/library/signal.html#signal.SIGALRM) signal be sent to the process in time seconds. Any previously scheduled alarm is canceled (only one alarm can be scheduled at any time). The returned value is then the number of seconds before any previously set alarm was to have been delivered. If time is zero, no alarm is scheduled, and any scheduled alarm is canceled. If the return value is zero, no alarm is currently scheduled.

https://docs.python.org/3/library/signal.html#signal.alarm

petersilva:
The change is symmetric.  If you cancel a non-existent alarm... that's an error.  so when I changed the thing setting the alarm to... not do that... the cancel has to match...
I think... that was what I was thinking...  

I don't know why one would want to cancel an alarm that we didn't set... how do we know?
so yeah... setting it to sero disables setting and clearing timeouts....

petersilva:
This branch will never be merged. Kept here as a place holder, reminder for a future branch.
* drop the alarm clearing with arg. is already addressed by try/finally logic already merged.
* need to extract the nofsetattr option stuff from this, and build a new PR with just that in it.

PR #1191, opened 2024-08-27
Issue1023 take2 topicCopy fix

### LINKED ISSUE #1023: What is correct "topic" processing in sr3?
Proposed behaviour:
      
* a topicCopy flag is added to sr3 along the model of timeCopy and permCopy to have the topic received reflected in what is sent.
* when topicCopy is set, a msg['topic'] is set, and on publication, that topic will be used.
    * cannot work with amqp->mqtt transitions
    * cannot work with message format changes (e.g. v02 -> v03)
* default is always off.

In *sr3 convert* the *topicCopy  on* option will be added for best compatibility with v2, but default in sr3 will be off, as the sr3 behaviour is likel more correct than what v2 does.

Starting point:

* on receipt of a message, a topic is split into a list (based on topicSeparator for protocol (. or / )) and the topicPrefix removed.   the result is stored as msg['subtopic']
* on transmission, logic is in postformat/\_\_init\_\_.py/deriveTopic() 
    * if mqtt then scheme is exchange/topicPrefix/path
    * else amqp then scheme is topicPrefix/path
    * if there is a "tropic" header set in the message, then use that to override the normal derivation and use that.

so whenever a message is received/parse it is split to have msg['subtopic'] set, and that is combined with the post_topicPrefix on publication.

The above is how it works now.

petersilva:
close #1023 

* um... again... topicCopy should work now.

### DIFF
--- a/sarracenia/postformat/__init__.py
+++ b/sarracenia/postformat/__init__.py
@@ -64,18 +64,18 @@ class PostFormat:
         pass
 
     @staticmethod
-    def exportAny(msg, post_format='v03', topicPrefix=[ 'v03' ], options={ 'post_format': 'v03', 'topicPrefix':'v03' } ) -> (str, dict, str):
+    def exportAny(msg, post_format='v03', topicPrefix=[ 'v03' ], options={ 'post_format': 'v03', 'topicPrefix':'v03' }, topicOverride=None ) -> (str, dict, str):
         """
           return a tuple of the encoded message body, a headers dict, and content_type
           and a completed topic as a list as one header.
        """
         for sc in PostFormat.__subclasses__():
             if post_format == sc.__name__.lower():
-                return sc.exportMine( msg, options ) 
+                return sc.exportMine( msg, options, topicOverride ) 
 
         return None, None, None
 
-    def topicDerive(msg, options ) -> list:
+    def topicDerive(msg, options, topicOverride=None ) -> list:
         """
            Sarracenia standard topic derivation.
 
@@ -96,11 +96,11 @@ class PostFormat:
             topic_prefix = options['topicPrefix']
             topic_separator='.'
 
-        if 'topic' in msg:
-            if type(msg['topic']) is list:
-                topic = msg['topic']
+        if topicOverride in msg:
+            if topicOverride is list:
+                topic = topicOverride
             else:
-                topic = msg['topic'].split(topic_separator)
+                topic = topicOverride.split(topic_separator)
         elif 'topic' in options and options['topic'] and (type(options['topic']) is not list):
             topic = options['topic'].split(topic_separator)
         else:
--- a/sarracenia/postformat/__init__.py
+++ b/sarracenia/postformat/__init__.py
@@ -96,7 +96,7 @@ class PostFormat:
             topic_prefix = options['topicPrefix']
             topic_separator='.'
 
-        if topicOverride in msg:
+        if topicOverride:
             if topicOverride is list:
                 topic = topicOverride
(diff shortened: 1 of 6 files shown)


sarracenia/postformat/__init__.py
```diff
@@ -96,11 +96,11 @@ def topicDerive(msg, options ) -> list:
             topic_prefix = options['topicPrefix']
             topic_separator='.'
 
-        if 'topic' in msg:
-            if type(msg['topic']) is list:
-                topic = msg['topic']
+        if topicOverride:
+            if topicOverride is list:
```
reidsunderland: I think this should be `if type(topicOverride) is list:`

reidsunderland: It worked in my test, because the topicOverride was a string. But if we pass in a topicOverride list it would break

petersilva: ugh... revised the whole stanza. different now.

reidsunderland:
I will do a bit more testing tomorrow. 

My initial quick test seems like this is always copying the source message's topic, even with `topicCopy False`. But I could have done something wrong in my test too.

petersilva:
OK.  I think with this last patch it does what people want.
* if the topicCopy option is set, the inbound topic overrides and is copied to the published message.
* if topicCopy is not set, then it tries to build a normal 'subtopic' ... if it can't it doesn't ... the message just doesn't get a topic or a subtopic...

(with previous patch, when the 'subtopic' could not be built, it forced on topicCopy... which "broke" the use case of interest.)

reidsunderland:
I posted on Teams, just adding here for proper documentation.

topicCopy works now. But with topicCopy off, the topic that gets posted is `v02.post` and nothing else. It should be `v02.post.data.msc ...`

petersilva:
@reidsunderland I looked over the code again... I think it only gets properly set when it is being published, so the log of the component itself would show what you described... is there a way of looking at the next component in the chain ... a subscriber to what was posted?  I have the feeling it is published with a good topic, but it will show that truncated one within the component.

petersilva:
The topic is only properly set on publish... based on the relPath.

reidsunderland:
I rebased this on the latest dev branch and will finally test it now

reidsunderland:
Good news: `topicCopy True` works correctly, bad news: `topicCopy False` doesn't work :(

With `topicCopy false`:

- the logs say the topic being posted is just `v02.post`
- It should match the relPath, e.g. `v02.post.data.rwin.observation.atmospheric` for relPath `data/rwin/observation/atmospheric...`
- I also double checked that this is not just a log issue, the broker says the routing key the messages were published with is `v02.post`.

With `topicCopy true`:

- it works
- the logs show that the topic is the same as we received from the source, e.g. `v02.post.dms-dev.nws.observation.atmospheric`, for relPath `data/nws/observation/atmospheric/...`
- I confirmed that the broker also gets the correct routing key

I can try to fix it when I have some free time.

petersilva:
This is endlessly confusing, but I think what you are describing is the intended, correct  behaviour.
The issue here is that the topic given is invalid, so when topicCopy is off, if the inbound topic is not valid, do not set it.  you get v02.post because that comes from the post_topicPrefix.

If the inbound topic were correct, then it would be fine.  The DMS input topic is not correct, so there is no correct behaviour possible when *topicCopy off*  I could be wrong... but that´s my guess.

petersilva:
This is really old... I don't even remember what behaviour we were aiming for.
We should probably close this and start over at some point.

PR #1192, opened 2024-08-27
fix rejectPercent maybe other stuff?

petersilva:
@reidsunderland  noticed a metric calculation that is off.

Now that people can see the metrics on real deployments, problems should be a lot more obvious.  it might be worthwhile rooting around to see if other problems are detected.
We can maybe catch a couple with this PR.

### DIFF
--- a/sarracenia/__init__.py
+++ b/sarracenia/__init__.py
@@ -359,17 +359,45 @@ def durationToSeconds(str_value, default=None) -> float:
     if default and str_value.lower() in [ 'on', 'true' ]:
         return float(default)
 
-    if str_value[-1] in 'sS': factor *= 1
-    elif str_value[-1] in 'mM': factor *= 60
-    elif str_value[-1] in 'hH': factor *= 60 * 60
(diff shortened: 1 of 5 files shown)


petersilva:
second patch changes "lag from 595483.20s to 6d21h" ... e.g:
```

 sarra/get_marine_weather                      lag     1/1    49 100%  47%     0 595483.20s 296574.88s 1141s  0.0%      0 B/s      0 m/s      0 B/s      0 B/s

```
becomes:
```
 sarra/get_marine_weather                      lag     1/1    49 100%  47%     0 6d21h 3d10h 1141s  0.0%      0 B/s      0 m/s      0 B/s      0 B/s

```
so it does not upset the columns that way either.

petersilva:
I just noticed my fancy formatting has a problem... I made a conversion routine to make like 3h4m (3 hours and 4 minutes) but I didn't do the reverse conversion, which is needed.  another commit coming.

petersilva:
fixed now.

reidsunderland (approved):

Issue #1156, opened 2024-08-07
Option to log when a message is rejected because it's a duplicate, but not for any other reason


reidsunderland:
Right now, we have the `logReject` option (which adds `reject` to the `logEvents`). We can turn this on to see why messages are rejected. 

But in cases where messages are rejected because they're duplicates, it would be nice to see that in the logs, without logging rejections for any other reasons. 

We could add a new option, maybe `logReject_duplicate` to turn on logging duplicates.

It should just take a couple changes in this area: https://github.com/MetPX/sarracenia/blob/301519e6edf0c4288f8fc2fbe25208a7d50d12e3/sarracenia/flowcb/log.py#L143-L150

to check for the new option and check if the report message contains nodupe?

https://github.com/MetPX/sarracenia/blob/301519e6edf0c4288f8fc2fbe25208a7d50d12e3/sarracenia/flowcb/nodupe/disk.py#L160-L174

v2 winnows would do this, and @junhu3 wants the same in sr3. I think it would be good to have this new option enabled by default in winnows.

Maybe @mshak2 has time to look at this one?

reidsunderland:
I also see this inconsistency, I think maybe the cache check should be changed to nodupe check but I'm not sure if it was left as cache (from v2) for a specific reason:

https://github.com/MetPX/sarracenia/blob/301519e6edf0c4288f8fc2fbe25208a7d50d12e3/sarracenia/flowcb/nodupe/disk.py#L178-L182

petersilva:
uh... logDuplicates as a flag?

reidsunderland:
yeah that works too, i said `logReject_duplicates` because I thought of it as kind of a logReject subset option, but it doesn't matter to me what the option is called

petersilva:
* logDuplicates ... add to flags in config.py.. default to False. (or add an event to logEvents.

* modify flowcb/log.py to honour...
   * there are multiple places where duplicates get refused... gotta check them all...
   * can check msg['report']['code'] == 304.
   * some fixes needed to flowcb/nodupe/disk.py (perhaps redis as well... )
   * fix typos modififi, cache vs. nodupe
   * some 304's are not always right... "too new" should become 425.
   * too old is also 304 ... should be ... 406 ... not acceptable.

petersilva:
question: do we want to tack on the the other improvements to log to this issue?
* log sundew_extension if present.
* log first seven (based on git example) chars of identity value, like so "id: ab13f12"  
* log size.

[...]

PR #1193, opened 2024-08-27
Fix for #1156

mshak2:
Doesn't pass static_flow test on dirt22-2 (33/35 tests passed).

Doing a draft PR to test on github.

Note: In winnow.py, logDuplicates is set to True by default as requested by Jun/Reid


petersilva:
I disagree with setting options different in different components.  That should be avoided unless there is a really good reason.  Therefore I ask that the default setting be changed in all components, rather than just winnow.

petersilva (changes requested):
don't want different defaults for  winnow vs. other components.
please set this for all.

petersilva (changes requested):
```

'nodupe' in msg['report']['message']:

```
that text will never be in that field... the test should be something like:
```
msg['report']['code'] in [ 304 ] 

```
perhaps add codes later (discussion) ...

petersilva:
Looking over things... this patch only covers the winnow case, not the general subscription.  In general subscriptions there is a second point where duplicate suppression happens... just before a file is downloaded, if it would overwrite an existing one, it is checked... and a 304 rejection done if they are considered the same.

(can see this in flow/__init__.py/file_should_be_downloaded... )

So I think a second part of this improvement should be to check for duplicates in after_work, in addition to after_accept.

petersilva:
fwiw... I just ran the original branch on my wsl, and it passes.  Still want some changes though.  will re-run later.

petersilva (changes requested):
There was another required change... have the same stuff done to after_accept also done to after_work in log.py because some rejections for duplicate happen in the download phase.

petersilva:
The request was to make rejects only of duplicate suppression be logged, but not rejections for any other reason.  I looked for 304 in the sarracenia/flow/__init__.py and there are:
```

     self.reject(
                                m, 304, "mask=%s strip=%s url=%s" %
                                (str(mask), strip, urlToMatch))

```

```

               self.reject(msg, 304,
                            "mtime not newer %s " % (msg['new_path']))

```

need to change the numbers for those.

mshak2:
> The request was to make rejects only of duplicate suppression be logged, but not rejections for any other reason. I looked for 304 in the sarracenia/flow/**init**.py and there are:
> 
> ```
> 
>      self.reject(
>                                 m, 304, "mask=%s strip=%s url=%s" %
>                                 (str(mask), strip, urlToMatch))
> ```
> 
> ```
> 
>                self.reject(msg, 304,
>                             "mtime not newer %s " % (msg['new_path']))
> ```
> 
> need to change the numbers for those.

So first one should be 404, and second one should be 406?

petersilva:
@mshak2 there are conflicts with development branch. Can you merge dev to this branch so the conflicts get resolved?

petersilva (approved):

PR #1202, opened 2024-08-30
Implement HTTP(S) stat using urllib, and HTTPS transfer class minor refactor

### LINKED ISSUE #1157: Add stat call to transfer protocols, and logic to avoid useless retries when source changes.

reidsunderland:
Implements HTTP(S) stat using urllib, which allows the existing authentication to work with stat calls.

While working on this, I had to make some changes to the existing `__open__`, which should fix problems and be a general improvement from the existing code. As long as nothing breaks...

This also includes @petersilva's implementation of stat in other protocols and the code that calls stat when a download results in a file with an unexpected size.

Fixes #1157.

### DIFF
--- a/.github/workflows/flow_basic.yml
+++ b/.github/workflows/flow_basic.yml
@@ -26,7 +26,7 @@ jobs:
       # Don't cancel the entire matrix when one job fails
       fail-fast: false
       matrix:
-       osver: [ "ubuntu-20.04", "ubuntu-22.04" ]
+       osver: [ "ubuntu-20.04", "ubuntu-22.04", "ubuntu-24.04" ]
 
     runs-on: ${{ matrix.osver }}
   
--- a/.github/workflows/flow_mqtt.yml
+++ b/.github/workflows/flow_mqtt.yml
@@ -29,7 +29,7 @@ jobs:
       fail-fast: false
       matrix:
        which_test: [ static_flow, no_mirror, flakey_broker, dynamic_flow ]
-       osver: [ "ubuntu-22.04" ]
+       osver: [ "ubuntu-22.04", "ubuntu-24.04" ]
 
     runs-on: ${{ matrix.osver }}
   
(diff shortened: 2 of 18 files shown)


petersilva (approved):
Looks great! I ran the subscribe/hpfx_citypage.conf and noticed that actually most of the re-writes are caught by the change in file_should_be_downloaded now properly respecting the arbitrary checksums... still every hour there are still dozens of:

```
 "2024-08-31 20:02:25,812 [INFO] sarracenia.flow download upstream resource is newer, so message https://hpfx.collab.science.gc.ca /20240901/WXO-DD/citypage_weather/xml/QC/s0000611_f.xml is obsolete. Discarding."

```

Message... so it definitely works. This will be a great Enhancement reducing wrongful retries substantially.

PR #1208, opened 2024-09-04
Fix badly expiring timers

petersilva:
more work on #966  That issued dealt with timers expiring instances being killed because transfers of large files appeared to be dead.  This is about timers being set and forgotten. In normal processing, it looks like all timers should be cleared, but in a number of failure cases, the timers might not be cleared.
It's unclear what specific case is causing a client issue.

Review of the source code showed:
* only place alarm timers are set is in the Transfer class hierarchy.
* only client report of failures is in ftp transfers.
* review of source revealed many, many places where errors could cause alarm clearing to be missed.
* potential code readability improvement from try/finally applicaiton.

So decided to look for and correct all the vulnerable parts of code where an exception would cause the timer to not be cleared.

### DIFF
--- a/sarracenia/transfer/azure.py
+++ b/sarracenia/transfer/azure.py
@@ -30,7 +30,6 @@ import stat
 import json
 
 from sarracenia.transfer import Transfer
-from sarracenia.transfer import alarm_cancel, alarm_set, alarm_raise
 
 from azure.storage.blob import ContainerClient
 import azure.core.exceptions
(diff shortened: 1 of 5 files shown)


reidsunderland (approved):
The `finally` in https is especially nice, it cut 5 alarm_cancel calls down to one :)

andreleblanc11:
Testing in a dev environment...

andreleblanc11:
From my tests:

- Running without this branch, the problematic sender timed out in about 10 minutes.
- Running with this branch, the problematic sender never timed out for over an hour of usage.

I think this is enough validation

andreleblanc11 (approved):

Issue #1209, opened 2024-09-04
Add a new option to poll at specific times of the day, `scheduled_time`


andreleblanc11:
We currently only have 2 ways of adjusting how we poll/scheduled flow files:

1. At a specific time interval, `scheduled_interval`, meaning 'poll at each X seconds' no matter what
2. At the same hours, minutes, `scheduled_minute` and `scheduled_hour`

The problem with `scheduled_minute` and `scheduled_hour` is that it doesn't allow you to have specific times within a list.
Also, every minute specified in `scheduled_minute` will appear in `scheduled_hour` which can attribute to wrong behaviour for specific use cases.

Adding a new option to only poll at specific times of the day would solve this problem, `scheduled_time`.

With the way that the scheduled flow is setup in https://github.com/MetPX/sarracenia/blob/development/sarracenia/flowcb/scheduled/__init__.py, makes it pretty easy to add this option.

PR #1210, opened 2024-09-04
Add `scheduled_time` for #1209

andreleblanc11:
`scheduled_time` allows us to poll for files when `scheduled_minute`/`scheduled_hour` offers a too broad spectrum of appointments to choose from.

`scheduled_time` should be used when we want to specify specific appointment times when we want to poll/scheduled flow.

### DIFF
--- a/docs/source/Explanation/CommandLineGuide.rst
+++ b/docs/source/Explanation/CommandLineGuide.rst
@@ -954,8 +954,13 @@ to poll the remote resources every thirty minutes. Alternatively::
    scheduled_hour 1,13,19
    scheduled_minute 27
 
-specifies that poll be run at 1:27, 13:27, and 19:27 each day.
- 
+specifies that poll be run at 1:27, 13:27, and 19:27 each day. Alternatively::
+
+   scheduled_time 15:30,20:30,23:15
+
+will poll the remote resources at exactly 15:30, 16:30 and 23:15.
+
+
 By default, sr_poll sends its post notification message to the broker with default exchange
 (the prefix *xs_* followed by the broker username). The *post_broker* is mandatory.
 It can be given incomplete if it is well defined in the credentials.conf file.
--- a/docs/source/Reference/sr3_options.7.rst
+++ b/docs/source/Reference/sr3_options.7.rst
@@ -1762,7 +1762,7 @@ If the speed observed is to be considered normal for that flow, then set the thr
 appropriately.
 
 
-scheduled_interval,scheduled_hour,scheduled_minute
+scheduled_interval,scheduled_hour,scheduled_minute,scheduled_time
 --------------------------------------------------
 
 When working with scheduled flows, such as polls, one can configure a duration
@@ -1772,7 +1772,7 @@ given activity::
   scheduled_interval 30
 
 run the flow or poll every 30 seconds.  If no duration is set, then the 
-flowcb.scheduled.Scheduled class will look for the other two time specifiers::
+flowcb.scheduled.Scheduled class will then look for the hour/minute time specifiers::
 
   scheduled_hour 1,4,5,23
   scheduled_minute 14,17,29
@@ -1780,6 +1780,14 @@ flowcb.scheduled.Scheduled class will look for the other two time specifiers::
 
 which will have the poll run each day at: 01:14, 01:17, 01:29, then the same minutes
 after each of 4h, 5h and 23h.
+
+If no scheduled_time nor scheduled_hour is given, then flowcb.scheduled.Scheduled class will
+then look for the remaining time specifier::
+
+  scheduled_time 15:30,16:30,18:59
+
+this will poll the data at 15:30, 16:30 and 18:59 every day. This option allows you to fine tune more
+your time field then previous options.
 
 sendTo <url>
 ---------------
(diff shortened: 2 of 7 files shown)


petersilva (changes requested):
*if scheduled_minute and scheduled_hour*
*  means you cannot use either one alone, have to specify both always.

If somebody specifies times that aren't in ascending order, your method won't work.  need to sort the times before adding them.  Entire appointment list must be in ascending order through the day.

There are unit tests for this module.  You should run them and add new tests with
* one time
* multiple times
* multiple times in wrong order.
* invalid times.

andreleblanc11:
Interesting, I thought they came in pairs and you had to specify both. I will change the `and` to a `or` then


> If somebody specifies times that aren't in ascending order, your method won't work. need to sort the times before adding them. Entire appointment list must be in ascending order through the day.

I've added 

https://github.com/MetPX/sarracenia/pull/1210/files#diff-fcf5177e0914e714809f1ff64076b7d252de5e6553b577e6ce6b197105ac6ff8R91

```python3 
self.sched_times.sort()
```

for this. I've made some tests before submitting the PR to validate that it was working

petersilva:
thinking about it... 
```
>>> import datetime
>>> appointments = [ datetime.time(12,10), datetime.time(15,15), datetime.time(7,7) ]
>>> appointments
[datetime.time(12, 10), datetime.time(15, 15), datetime.time(7, 7)]
>>> appointments.sort()
>>> appointments
[datetime.time(7, 7), datetime.time(12, 10), datetime.time(15, 15)]
>>>

```
so you can just add the appointments to the list and then sort appointments.
You don't need to bother if else making separate lists, just split into hours and minutes and add them to the list, and then sort the result. like after the first loop, add a second loop:
```
... first loops about minutes and hours...

for t in scheduled_times:
    h,m=t.split(':')
    appointment=datetime.time(h,m, tz... like there already)
   next_time = datetime.datetime.combine(when,appointment)

self.appointments.sort()

```
After the second loop, we have the whole series of minutes and hours, and then a second series potentially out of order, of times that could intersperse with the previous set... so need to sort the whole thing anyways.


maybe less code than you have so far.

andreleblanc11:
I think with that logic we can also remove

```python3
self.hours.sort()
# and
self.minutes.sort()
# in the code
```

andreleblanc11:
I also never thought of someone using both options together.. it doesn't hurt to have that option available

andreleblanc11:
I've found a bug when trying to use the `default_wait` value. It wasn't using a `datetime.delta` format so I encapsulted it with that and it fixed the problem.

Here's the error we were getting

```
2024-09-05 13:51:05,965 [ERROR] sarracenia.flow gather flowCallback plugin <bound method Scheduled.gather of <sarracenia.flowcb.scheduled.Scheduled object a
t 0x7f34eff24be0>> crashed: '<' not supported between instances of 'int' and 'datetime.timedelta'
2024-09-05 13:51:05,965 [DEBUG] sarracenia.flow gather details:
  File "/net/local/home/leblanca/.local/lib/python3.10/site-packages/sarracenia/flow/__init__.py", line 1133, in gather
    retval = p(self.o.batch-so_far)
  File "/net/local/home/leblanca/.local/lib/python3.10/site-packages/sarracenia/flowcb/scheduled/__init__.py", line 100, in gather
    self.wait_until_next()
  File "/net/local/home/leblanca/.local/lib/python3.10/site-packages/sarracenia/flowcb/scheduled/__init__.py", line 221, in wait_until_next
    self.wait_seconds(self.default_wait)
  File "/net/local/home/leblanca/.local/lib/python3.10/site-packages/sarracenia/flowcb/scheduled/__init__.py", line 140, in wait_seconds
    if sleepfor < nap:
TypeError: '<' not supported between instances of 'int' and 'datetime.timedelta'

```

petersilva:
I added tests for when one specifies only the minute, or only the hour ... and the current code fails both tests.

petersilva:
the message being printed out for default interval...:
```
 no scheduled_interval or appointments (combination of scheduled_hour and scheduled_minute) set defaulting to every 0:05:00 seconds

```

* maybe mention schedule_hour/minute/time (no mention of new option in message.)
* 00:05:00 seconds... looks weird.  Not sure what to do about it, but it looks odd.

andreleblanc11:
> I added tests for when one specifies only the minute, or only the hour ... and the current code fails both tests.

I think this is expected based on how the code is setup now? If `scheduled_minute` or `scheduled_hour` is not specified, there is no default that is given to resolve the missing option. Maybe we should default to **all of the minutes and all of the hours**  when either option is not specified?

andreleblanc11:
> 00:05:00 seconds... looks weird. Not sure what to do about it, but it looks odd.

That's because the default_wait is now in datetime format. Need to change the output to match.

petersilva (commented):
I've seen that bug before... seconds? or total_seconds() ?

andreleblanc11:
`seconds` is wrong and `total_seconds` should be used , but it's probably fine to leave it as `default_wait` will never have a changed value.

petersilva (approved):

petersilva:
oh darn... I merged it, but we didn't put the slack ("loose") in... so it will still fail.

andreleblanc11:
I was going to spawn a different branch for that different issue #1214, and create another PR

PR #1211, opened 2024-09-05
Promote 2.24.08post2 to v2 stable?

reidsunderland:
I think we're ready to release 2.24.08post2 as the new stable v2 version?
UPDATE: v2 flow tests were slightly broken... fixed.. now re-run...


andreleblanc11:
I think we've found most of the new V2 bugs so I think we're ready. All of the changes were also tested.

petersilva (approved):

PR #1223, opened 2024-09-16
Issue1214 another scheduled refactor

petersilva:
Ha-ha... PR for @reidsunderland 's stuff... this topic is confusing, so I wrote a bunch of unit tests, and  yay! this formulation seems correct as far as I can tell.  so made the PR.

This fixes the regression we got from the #1217 ... and adds much more confidence in that we now have unit tests of what went wrong there.
If somebody breaks it the unit tests should squawk.

### DIFF
--- a/sarracenia/flowcb/scheduled/__init__.py
+++ b/sarracenia/flowcb/scheduled/__init__.py
@@ -46,6 +46,54 @@ class Scheduled(FlowCB):
 
     """
 
+    def __init__(self,options,class_logger=logger):
+        super().__init__(options,class_logger)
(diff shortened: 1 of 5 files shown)


reidsunderland (approved):
It looks good to me, as long as we don't need the unit tests to run on Ubuntu 18.04/Python 3.6 because it has the nice `variable=` fstrings that don't work there. I think that's okay.

Thanks for doing the unit test :)

petersilva:
unit tests don't run on 18.04 at all...  unfortunately... requires 22.04.

Issue #989, opened 2024-03-20
credentials.conf handling of certain characters in usernames and passwords


reidsunderland:
I was not able to use a password that had `/` in it in sr3.

```
https://username:pass/word@example.com
```

I was using a plugin, so I was able to workaround it by using `%2F` in credentials.conf and `urllib.parse.unquote` in the plugin.

```
[ERROR] 2181978 sarracenia.credentials isValid credential not found
```

The problem seems to come from urllib.parse.urlparse:

```
>>> urllib.parse.urlparse("https://username:pass/word@example.com")
ParseResult(scheme='https', netloc='username:pass', path='/word@example.com', params='', query='', fragment='')
```

petersilva:
perhaps we could have an alternative syntax with a separate keyword... aka:

amqp://user@hoho password=lovel/

so use something completely different than urlparse.  There is already syntax for other keywords later in the line...

petersilva:
look at sarracenia/credentials.py  syntax examples given with passive= and binary keywords.

reidsunderland:
Yes, I like that idea! 

Another option would be using a keyword to tell sr3 whether the username and password is url encoded or not. 

```
https://u%2Fs3rn%40me:P%40%24%24w0r%2Fd%2F@example.com urlencoded=True
```

username = `u/s3rn@me`
password = `P@$$w0r/d/`

petersilva:
I like the urlencode keyword idea also... probably that is a better approach

petersilva:
I vaguely recalle we tried *always* url encoding in the past, and something bad happenned, so we had to roll back... so use of the keyword does seem necessary. #884 is the thing I remember...

reidsunderland:
The documentation does say that credentials are URL encoded:

https://metpx.github.io/sarracenia/Reference/sr3_credentials.7.html#credential-details 

> "These strings are URL encoded, so if an account has a password with a special character, its URL encoded equivalent can be supplied. In the last example above, %2f means that the actual password isi: /dot8 The next to last password is: De:olonize. ( %3a being the url encoded value for a colon character. )"

reidsunderland:
@robjarawan had problems with a password containing `#`:

```
>>> urllib.parse.urlparse("amqps://USERNAME:pass#word@broker.example.com/")
ParseResult(scheme='amqps', netloc='USERNAME:pass', path='', params='', query='', fragment='word@broker.example.com/')
```

reidsunderland:
[...]

PR #1228, opened 2024-09-18
Unquote secret_access_key in s3CloudSender plugin

reidsunderland:
This plugin is kind of redundant because we have a built in S3 transfer class now, but we're still using it right so we might as well have the latest code in the repo. I don't think the transfer class handles URL encoded passwords.

In this change, I'm assuming the password is always URL encoded.

We have an issue to add an option to credentials.conf to let the user specify whether it's URL encoded or not. Right now most (all?) of the rest of the code assumes that passwords are not URL encoded. 

https://github.com/MetPX/sarracenia/issues/989

### DIFF
--- a/sarracenia/flowcb/send/s3CloudSender.py
+++ b/sarracenia/flowcb/send/s3CloudSender.py
@@ -104,6 +104,7 @@ from sarracenia.credentials import Credential
 from sarracenia.flowcb import FlowCB
 import boto3
 from botocore.exceptions import ClientError
+from urllib.parse import unquote
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


petersilva:
This one worries me... there might be unintentional escape sequences in our existing db's and also all the secrets need to be encoded... not clear how the migration works.

petersilva:
Is this only for s3 or for all passwords?  Should we do it for all passwords?

reidsunderland:
This PR is just for the s3CloudSender plugin, there's no change to the S3 transfer class or any other part of the code. It should be safe. This is the version of the code that we've been using in ops for a few months from the internal plugins repo.

reidsunderland:
For all other parts of the code, we decided #989 that we'd add an option to allow the user to specify if a password is URL encoded or not. I still think that's a good approach

petersilva (approved):

Issue #1206, opened 2024-09-04
lag metric reporting incorrect statistics


andreleblanc11:
Sometimes, some configurations will report a high lag time, which triggers the `lag` state in `sr3 status`.

Example

```
Component/Config                              Processes   Connection        Lag                              Rates
                                              State   Run Retry  msg data   Que   LagMax   LagAvg  Last  %rej     pubsub   messages     RxData     TxData
sarra/get-cmems-oceanobs                      lag     1/1     0 100%   0%     0  171.24s   90.80s 2.43s  0.0%    564 B/s      0 m/s 40.3 KiB/s      0 B/s
sarra/get_airnow_http                         lag     1/1     1 100%  67%     0 541312.32s 541312.32s 4196s  0.0%      0 B/s      0 m/s      0 B/s      0 B/s
sarra/get_airnow_prov_depot                   lag     2/2    13 100%  69%     0 586328.54s 310211.80s 6940s  0.0%      0 B/s      0 m/s      0 B/s      0 B/s
sarra/get_marine_weather                      lag     1/1    53 100%  72%     0 533804.73s 212781.40s 8549s  0.0%      0 B/s      0 m/s      0 B/s      0 B/s
sarra/get_rvas_rem_pxpaz                      lag     1/1     1 100%  91%     0 270340.14s 270340.14s 6424s  0.0%      0 B/s      0 m/s      0 B/s      0 B/s
```

The first value looks to make sense. The other 4 values span from 3 to 6 days which don't make sense when looking at their `last` values.
So far this problem looks to only be affecting sarra components. Still unsure why

We checked the output of `sr3 dump` it also looked to report a similar story. `lagMax` and `lagTotal` are of huge values

```
[...]

PR #1229, opened 2024-09-18
Issue1206 2nd try 404

petersilva:
Second shot at #1206 (yeah branch name is off by one...)

* change send signature to match the change in the download signature... they both return integers now. <0 means permanent failure, 0 means queue for retry > 0 means yay! it worked.
* also catch http 404's and call them permanent.

This should address the sort of bogus (will never recover) lag we see
on ops with 3.0.54.

### DIFF
--- a/sarracenia/flow/__init__.py
+++ b/sarracenia/flow/__init__.py
@@ -2312,7 +2312,15 @@ class Flow:
         return 1
 
     # generalized send...
-    def send(self, msg, options):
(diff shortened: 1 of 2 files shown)


petersilva:
@reidsunderland and I are both nervous about the 404 change, because we imagine too many situations where 404 is a temporary problem that should be retried.   likely will take the sender commit, and make a new PR for that.

for 404 ? maybe make an option "drop 404" ...

Issue #339, opened 2020-06-10
Can We Gather from Multiple Sources?


petersilva:
with v3 and  use of plugins for gather... can we just subscribe to x different brokers and get messages from all of them with a single flow?

this probably means that the error recovery needs to change so the we retry the broken gather sources, but also come back, and use the others...

one could just list multiple brokers, and the subscribe would connect to all of them at once, checking each one in turn.

(implementation for that case would be adding plugin.gather.message multiple times, each with a different broker argument...) would require some other changes... perhaps specify different exchanges, per server?)

Can also do a winnow in a single subscriber process... (using cache, and two upstream sources.)

would be cool, and not that hard to get done in v3.

petersilva:
I thought of a grammar for this.  We add a *subscribe* keyword to the config grammar.
for AMQP: 
```

subscribe *broker* *exchange* *topic*

```
an example would be:
```

subscribe amqps://hpfx.collab.science.gc.ca xpublic  v02.post.*.WXO-DD.bulletins.alphanumeric.#
subscribe amqps://dd.weather.gc.ca xpublic v02.post.bulletins.alphanumeric.#

```

This would subscribe to different topics on different brokers.

the same verb for MQTT would be missing the exchange... just have a topic.

@reidsunderland  you mentioned publishing to multiple ... who was interested in multiple subscribes?

petersilva:
actually, we don't need to change anything in the grammar... we just have multiple exchange and subtopic declares and make relate them (so that each subtopic refers to the last exchange...) could do the same thing with broker... just repeat it.

the same way directory & accept work for scoping.

reidsunderland:
That sounds good to me -- making it work the same way as directory and accept.

reidsunderland:
I don't know of anyone with an actual need for publishing or subscribing to multiple exchanges/brokers

petersilva:
use case:

Sample use case:

* if you're doing HA, then it would be really cool to define two upstream sources, and use duplicate suppression on them.   vs:

* What we do today:  we define two shovels that feed a winnow, and subscribe to the winnow output.   ... so four configurations to one... and you need a local broker.

so not absolutely essential... but a heck of a lot simpler.

petersilva:
reading the source...  gathering from multiple exchanges already works in the python one...
```

exchage  zzz
topicPrefix ...
subtopic A
exchange yyy
subtopic B

```
[...]

Issue #35, opened 2018-06-02
Stateful Queue Binding maintenance


petersilva:
currently, when one starts up sr_subscribe with

subtopic A

then stops the sr_subscribe, and edits the file to show

subtopic B

Then restart sr_subscribe, the result is that both bindings are set on the queue. The new binding is added, but the old binding is not, so the result is a subscription to A and B, rather than just B.

The reason for this is that the queue must be preserved one wishes to receive items which arrive during transient failures of the consumer. But the very permanence of the queue causes the bindings previously in place to be preserved across instances.

One can correct this unfortunate effect by invoking sr_subscribe with --reset option, to delete the queue and with it, the old binding. The queue will be re-created with only the new binding. Unfortunately that has the effect of discarding all items already queued.

AMQP does not provide any way to query what bindings are in effect, nor does it allow to delete bindings without specifying them exactly. In order to delete a binding one must know exactly what it is. When the configuration has changed, the program no longer has access to what the binding was before the change, and so it does not have the information required to delete the old bindings.

To correct the behaviour, one needs to treat bindings as statefully. Currently, sr_subscribe stores the name of the queue in the user_cache_dir. One would need to add more content to the state directory to document the bindings in place when the subscribe was initiated.

When sr_subscribe starts up, need to compare the stored bindings with the bindings set by the current configuration, removing obsolete ones and adding the new ones. Then save the bindings to the binding state file.

This was all discussed with sr_subscribe as an example, but it applies equally to all consumers.

Was: https://sourceforge.net/p/metpx/feature-requests/49/

michelgrenier:
saving previous bindings and cleaning up what is not matching the config options on restart may
work... but needs to be tested if this example would work....  
 ex. old = exchange x, subtopic *.a.#
      new =  exchange = x subtopic = *.a.b.#   

unbonding    subtopic *.a.# ...   should not break  binding   subtopic *.a.b.# ...
if this is true... it is easy.

petersilva:
That will work fine... detail:  you want to:   1. diff the bindings... 2. apply net new ones, 3. remove net removed ones. in that order.

btw... AMQP supports this, but I don´t think Sarracenia does...

exchange XA

subtopic A

exchange XB

subtopic B

does binding a single subscriber to multiple exchanges work?  I don´t think so... I don´t know if it should.

michelgrenier:
yes this should work...
[...]

PR #1232, opened 2024-09-23
Issue35 try 2 stateful binding part1

petersilva:
Hey folks, This is work in progress... it does no harm, but it doesn't do much helpful either.  I started this work a month ago. It is related to #35, and #339.  It adds:
* *subscriptions.json* in the state directory, as a more featureful future replacement for *\*.qname* files, which will eventually permit multiple subscriptions for one subscriber.
* it refactors the Sarracenia.config a bit... sarracenia.credentials. is now sarracenia.config.credentials.
* there is a new sarraenia.config.subscriptions class.

This is prep-work for making sr3 notice when queue configs change, and then give you an error message instead of just hanging with PRECONDITION failed... giving you a message like "you changed expiry, you need to cleanup, or it won't work."

Right now, it creates a *subscriptions.json* file in the state directory.
That file stores the same information as the *\*.qname* file (and more.)
So far the work is to make it consistent with the qname stuff.

This is a breaking change, because:

* At some point after this happens, sr3 will start using the *subscriptions.json* file instead of *\*.qname*  If we have a transition version that maintains both, then it makes the eventual transition easier (the fully featured one could just  ignore the .qname files.)
* the class hierarchy changes... config.py -> config/__init__.py, and credentials.py becomes config/credentials.py ... so the sooner this small re-factor takes effect, the better.

The new *subscriptions.json* file is designed to permit multiple subscriptions (to multiple brokers) per subscriber... (as per #339) this is actually work that started by working on that, but backtracked when it turned out to be necessary to re-work the *\*.qname* files and ended up working on #35 as a pre-requisite for #339.

The new format supports that, but the code doesn't actually do that yet. It's just there is room in the format to put in multiple brokers.


petersilva:
yes... the changes are useless, but harmless. they pass all the tests.
Please review.

reidsunderland (approved):
I don't see any obvious problems that merging this would cause

PR #1235, opened 2024-09-24
Switch from global QoS to per-consumer QoS

reidsunderland:
Since quorum queues don't support global QoS, I think it makes sense to switch to per-consumer QoS.

It will also show the Prefetch Count for each consumer when looking at a queue in the RabbitMQ GUI, so that's a nice improvement as well.

I don't think there is any practical difference. We only ever have one consumer per connection and channel, so it should behave the same way.

-- edit to re-trigger tests

### DIFF
--- a/sarracenia/moth/amqp.py
+++ b/sarracenia/moth/amqp.py
@@ -349,7 +349,8 @@ class AMQP(Moth):
                 #logger.info('getSetup connected to {}'.format(self.o['broker'].url.hostname) )
 
                 if self.o['prefetch'] != 0:
-                    self.channel.basic_qos(0, self.o['prefetch'], True)
+                    # using global False because RabbitMQ Quorum Queues don't support Global QoS, issue #1233
+                    self.channel.basic_qos(0, self.o['prefetch'], False)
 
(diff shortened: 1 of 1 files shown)


petersilva (approved):

petersilva:
@reidsunderland I approved it, but it's still draft... go ahead and merge it if you're comfortable.

reidsunderland:
The tests are all failing because of the sr3c instances thing that has been fixed now. I'll rebase the branch.

```
  File "/home/sarra/.local/bin/sr3", line 33, in <module>
    sys.exit(load_entry_point('metpx-sr3', 'console_scripts', 'sr3')())
  File "/home/sarra/sr3/sarracenia/sr.py", line 3074, in main
  File "/home/sarra/sr3/sarracenia/sr.py", line 1323, in __init__
    self._read_states()
  File "/home/sarra/sr3/sarracenia/sr.py", line 556, in _read_states
    self._read_state_dir()
  File "/home/sarra/sr3/sarracenia/sr.py", line 493, in _read_state_dir
    i = int(pathname[0:-4].split('_')[-1])
ValueError: invalid literal for int() with base 10: 'i01'
```

PR #1237, opened 2024-09-26
Add recursion back

petersilva:
I just looked at the code and guessed where it needs to be added back.
There is one patch to add it to watch, one to document, and one to add it to poll.

I ran the flow tests and they pass as well as before.
Have not actually tested what happens when you turn recursive off.  someone perhaps should before merging this.

It's harmless, but I don't know if it does what it should.


reidsunderland (approved):
I haven't had a chance to test it, but the changes look good and we will test on dev after merging

andreleblanc11 (approved):

reidsunderland:
I tested with a poll and it works!

Issue #716, opened 2023-07-13
JSON log formatting.


petersilva:
user request:

```
Is there an easy want to customize the logging config in sr3? We do
most of our logging in JSON (with the python-json-logger package), so
that it can be easily ingested to our elasticsearch/logstash/kibana
cluster. Usually with a logging config such as:

handlers:
  file:
    class: logging.handlers.RotatingFileHandler
    level: INFO
    formatter: json
    backupCount: 10
    maxBytes: 10000000
    filename: 'crmprtd_log.txt'
```

petersilva:
This would also be great for NAGIOS integration and other monitoring stuff.

petersilva:
some thoughts:

* current logging (in sr3) uses builtin python logging modules in a standard way.
the logs are TimedRotatingFileHandles, there is probably a way to add some json output method using the module mentioned above.

* worries about ouroboros of log options being set in the config.... The CLI (sr3 itself) parses the config files the first time to figure out what logs are needed for instances started... then it starts them up with std files re-directed to them.   The sub-processes thus started then parse the configs a second time, so the re-directed logs should contain complete config parsing errors, if any... in json in the log.

* some stuff will be printed to stderr on startup before logs get re-directed... unavoidable.

petersilva:
more thoughts... the way things work, the stdout/stderr are re-directed to the log files, so while most output is published via python log API, when forking sub-processes, such as scp, or when doing integration with other systems, the output of those other systems is just catenated into the file.  so while most messages in the log a log formatted, some are just raw output.

seeing unadulterated outputs of sub-processes is felt to be the most practical form of output from them, which is why it is the way it is... so... even if we converted things, there would be some non-JSON data in the logs to be ignored.

jameshiebert:
> when forking sub-processes, such as scp, or when doing integration with other systems, the output of those other systems is just catenated into the file

Yeah, you'd have to decide how you want sr3 to handle that. It's technically possible to capture stdout/stderr from the subprocess and do something else with it (e.g. create log messages containing them), but you'd have to decide whether that's desirable or not.

Arguably, one could assume that if the user is configuring JSON logging, then it's intended to be ingested into another system for computational processing and a human will not be reading individual log entries. In which case, having subprocess output captured may be what is wanted.

petersilva:
see #1238

petersilva:
Avoided the problems mentioned above by making the .json a second log format and leaving the text one around for all the exceptional strings.

PR #1238, opened 2024-09-27
Json logging (EXPERIMENTAL)

petersilva:
See #716 

This option adds a flag to enable json formatted logging.

It has been requested by several people. but the problem was that
some of the text stuff currently being logged isn't log messages, just random
stdout of subprocesses (e.g. scp, wget) which will give a json parser fits.

So just made a second log file (with .json suffix) and tools that prefer that
can ingest that.  the  txt log  remains... so it logs twices as much.

experimental enabling of json formatted logging.

Does this make any sense?


petersilva:
purpose of this experimental feature is to let people kick the tires. it is documented as EXPERIMENTAL ... meaning we may remove it in a future release, or it could change... it is just to people to comment on for now.

reidsunderland (approved):

PR #1244, opened 2024-09-27
remove sleep loop from moth/setup routines.

petersilva:
working on #339.

We if we have multiple broker connections, then we can't loop waiting to connect to each one... because every time 
one broker connection goes down, it will just get stuck there.  

* old behaviour: getSetup and putSetup loop waiting to get a connection.
* new behavious: getSetup and putSetup fail and return.
* main loop then calls them again.

This is also consistent with the pattern of trying to have a single place where the code sleeps.

The only change in this code is to remove the loop (which outdented the entire routine by 4... sigh...)
changing "break" to "return" in a few places, and removing the sleeps.

### DIFF
--- a/sarracenia/__init__.py
+++ b/sarracenia/__init__.py
@@ -427,6 +427,7 @@ known_report_codes = {
     #FIXME : should  not have 503 error code 3 times in a row
     # 503: "Service unavailable. delete (File removal not currently supported.)",
     503: "Unable to process: Service unavailable",
+    504: "Gateway Timeout: message too old"
     # 503: "Unsupported transport protocol specified in posting."
 }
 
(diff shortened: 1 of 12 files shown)


petersilva:
so... first patch or two was to make sure that, we don't loop when a broker is down in the connection establishment phase.  A second step is that... when the connection establishment does fail, we should notice and back off for a while.

The later patches:
* notice how long a connection attempt took.
* implement exponential backoff as a function of how many connection failures, and how long a connection failure takes (if it takes 5 minutes, then it doesn't make much sense to wait 1 second before trying again.)  It should be at least 10 minutes before we try again.

petersilva:
There is still an inner loop in the mqtt getsetup stuff... Just doing the same thing to mqtt that was done to amqp isn't enough...  The change should work in some failure modes, but perhaps not others.

petersilva:
oh my goodness... every time I try to resolve conflicts... I just get more conflicts... 
I'm going to try a fresh branch and some cherry-picking.

petersilva:
replaced by #1247

PR #1247, opened 2024-10-03
Issue339 try4 part1 fail connections in main loop

petersilva:
replaced #1244 by cherry picking just the relevant commits and adding them to an uptodate development tree... #1244 was getting very confusing with > 20 commits of mixes of the two branches.  With this restart, it's only a handful of commits, much easier to review.

goal: If we have multiple broker connections, then we can't loop waiting to connect to each one... because every time
one broker connection goes down, it will just loop there.

* old behaviour: getSetup and putSetup loop waiting to get a connection.
* new behaviour: getSetup and putSetup fail and return.
main loop then calls them again.

This is also consistent with the pattern of trying to have a single place where the code sleeps.

The only change in this code is to remove the loop (which outdented the entire routine by 4... sigh...) changing "break" to "return" in a few places, and removing the sleeps.

So that's the first few patches... later it dawned that even if you loop from mainline... when you do a connection attempt and it hangs... you will spend all your time looping and hanging.  To get some free time (so other connections can be used.) added exponential back off to connection attempts.  That's the rest of the commits.  Look at how long it took for the connection to fail, and then back off for at least twice that long.

note that when failures are quick, they can take variable amounts of time, so the ebo might not continuously increase because the backoff period is a multiple of how long it took to fail, so if it fails faster, it might want to retry sooner.

you get an odd pattern of retries... that generally does the right thing, but looks wrong...

### DIFF
--- a/sarracenia/moth/__init__.py
+++ b/sarracenia/moth/__init__.py
@@ -284,6 +284,10 @@ class Moth():
         self.metrics = { 'connected': False }
         self.metricsReset()
 
+        now = time.time()
+        self.next_connect_time = now
+        self.next_connect_failures = 0
+
         if (sys.version_info.major == 3) and (sys.version_info.minor < 7):
             self.o = {}
             for k in default_options:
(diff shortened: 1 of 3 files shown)


petersilva:
people are worried about this one.  idea is to defer until .57 to stablize the changes we have in .56 and get that out the door first.

petersilva:
I'm going to do this over again later... with cleaner patches.

PR #1251, opened 2024-10-07
Fixing unit tests

petersilva:
These commits fixed the unit tests... 
kinda dumb they failed for these... but at least it's fixed...


petersilva:
The first one was, I think a mistake I made during a previous merge conflict... I had taken the "append" over the "add" ... it turned out we needed "add"

reidsunderland (approved):

PR #1258, opened 2024-10-10
tweaks to flowcb/gather/file.py after experiments with noticing large files.

petersilva:
some message naming things.

@reidsunderland to supplement. (edit to re-trigger tests)


reidsunderland:
I added my tweaks

andreleblanc11 (approved):
I think this is good.

PR #1272, opened 2024-10-23
Issue1270 - fixing relative symlinks with (inappropriate) ../../ in them... 

### LINKED ISSUE #1270: mirrorring relative symlinks with ../.. in them which are wrong...
so.. when mirroring with libsr3shim:

shim posting:
```
realpathFilter on
realpathPost on
realpathAdjust -1

```

subscriber:
```
baseDir /source
directory /dest

```

now there are paths which are /home/user/excellent (where excellent is a link into /source/excellent) so the
path of operations would be:
```
cd /home/user/excellent
mv a b
mkdir oneDirDeep
mkdir oneDirDeep/twoDirDeep
echo "hoho" >>oneDirDeep/twoDirDeep/2dd_file1
ln -s 2dd_file1 oneDirDeep/twoDirDeep/link_to_2dd_file1
mv  oneDirDeep/twoDirDeep/link_to_2dd_file1  oneDirDeep/twoDirDeep/renamed_link_to_2dd_file1

```

you end up with links with (wrong) extra ../../ in them.
also rename... maybe same issue maybe different.

petersilva:
close #1270

removed special relative translation that turns out to have been wrong-headed.
requires/works best with: https://github.com/MetPX/sarrac/pull/168

with the two fixes, renames and symlinks should be a lot better.

### DIFF
--- a/sarracenia/flow/__init__.py
+++ b/sarracenia/flow/__init__.py
@@ -890,7 +890,7 @@ class Flow:
                     if f in msg['fileOp']:
                         fopv = msg['fileOp'][f].split('/') 
                         # an absolute path file posted is relative to '/' (in relPath) but the values in
-                        # the link and rename fields may be absolute, requiring and adjustmeent when stripping
+                        # the link and rename fields may be absolute, requiring and adjustment when stripping
                         if fopv[0] == '':
                             strip += 1
                         elif len(fopv) == 1:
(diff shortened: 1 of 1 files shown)


reidsunderland (approved):
reviewed in a call with @petersilva  and @andreleblanc11

Issue #1269, opened 2024-10-22
U18.04 packages missing `sarracenia/flowcb/rename` directory w/ files


andreleblanc11:
The directory is available on U22.04 servers

```
root@ncp1-op:~# cat /etc/issue
Ubuntu 22.04.4 LTS \n \l
root@ncp1-op:~# ls -alFd /usr/lib/python3/dist-packages/sarracenia/flowcb/rename/
drwxr-xr-x 3 root root 48 Aug 15 15:56 /usr/lib/python3/dist-packages/sarracenia/flowcb/rename//
```

But is not on U18.04 servers?

```
root@ddsr-cmc-dev01:~# cat /etc/issue
Ubuntu 18.04.6 LTS \n \l
root@ddsr-cmc-dev01:~# ls -alFd /usr/lib/python3/dist-packages/sarracenia/flowcb/rename/
ls: cannot access '/usr/lib/python3/dist-packages/sarracenia/flowcb/rename/': No such file or directory
```

It looks like this has been the case since the inception of my `rename/raw2bulletin.py` plugin - [V3.0.53](https://github.com/MetPX/sarracenia/releases/tag/v3.0.53)

## Work around (kinda?)
Even when I tried to add the plugin to the gitlab repo it gave problems. 
The work around is to `mkdir rename` where it's supposed to be in the package directory. The code will then stop crashing, if the plugin is found from the gitlab repo.

petersilva:
you need to have a \_\_init__.py (could have just *pass* in it... ) in the directory in order for the packaging utilities to include it... I guess the newer ones don't need it but the older ones do?
could that be what is going on?

andreleblanc11:
The new pre-release package (3.00.56rc2) includes the `rename` directory now and its files. Closing issue.

PR #1273, opened 2024-10-23
Issue1271 - Change raw2bulletin renamer to an `after_accept` entry point

andreleblanc11:
- Changing to `after_accept` entry point to be callable from regular flow processes
- The change log is kinda chaotic for the renamer because I had to indent the code to adapt for the after_accept `for` loop.

- Unit tests have been corrected for this and pass.

- Ran the AM flow tests on U22.04 and U18.04. Tests had 10/11 results, because the watch was posting `modify` events when only `create` was specified in the config. So looks to be a separate issue?

- I also added the `__init__.py` file for #1269


petersilva (approved):
looks good! you even added tests... Excellent!

PR #1279, opened 2024-10-29
Issue1271 - Fix accomodation to sarra component

andreleblanc11:
I originally thought that `getData` under `sarracenia/bulletin.py` was correct to fetch data that's not inline, but I was wrong. I don't think it was ever tested.

I've been testing a sarra component on dev with these changes and it looks OK now. It can now get the data through the path.

The goal moving forward is to use this renamer in a non-AM setting through a sarra (which posts to a sender), or a flow/sender that has `download True`.

### DIFF
--- a/sarracenia/__init__.py
+++ b/sarracenia/__init__.py
@@ -1019,10 +1019,11 @@ class Message(dict):
 
         # inlined/embedded case.
         if 'content' in msg:
+            logger.info("Getting msg from inline'd content")
             if msg['content']['encoding'] == 'base64':
                 return b64decode(msg['content']['value'])
             else:
-                return msg['content']['value'].encode('utf-8')
+                 return msg['content']['value'].encode('utf-8') if not options.inputCharset else msg['content']['value'].encode(options.inputCharset)
 
         path=''
         if msg['baseUrl'].startswith('file:'):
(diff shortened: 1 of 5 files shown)


petersilva:
There is another routine sarracenia.Message.getContent() in __init__() that seems to do the same thing as getData...  it would be nice to leverage the existing impl, rather than having to maintain 2. but not a big deal.  This looks fine.

petersilva (approved):

petersilva:
the PR is marked draft... feel free to merge when you think it is good to go... or if you make substantive chances, ask for another review.

andreleblanc11:
It doesn't look like that much work and would simplify future code. Currently testing this

andreleblanc11:
I added some logic so that `getContent` checks for the `inputCharset` if it exists. We need this for french chars in bulletins.

petersilva (approved):
in the bulletin class \__init__ function, it says super().__init__() which is odd, because it's invoking the parent class initializer, when sarracenia.Bulletin has no parent class.  I don't know what it does.  tried an example on my own... seemed harmless... but strange... the super thing is for flowcb's ... but bulletin class isn't one...


but that's harmless.   everything else looks good!

andreleblanc11:
That was my initial attempt at getting options which didn't work. Forgot to remove it after, but isn't causing problems

Issue #1285, opened 2024-11-05
Add `fileSizeMax` to sender


andreleblanc11:
We currently can use `fileSizeMax` in a component that downloads, to avoid downloading too big of files.

https://github.com/MetPX/sarracenia/blob/4a9f6424eaee6ae50f6343bf6a7643a615c3740a/sarracenia/flow/__init__.py#L1795-L1799

It would be nice to also have this for a sender.

Likely need to add same logic to `do_send`

https://github.com/MetPX/sarracenia/blob/4a9f6424eaee6ae50f6343bf6a7643a615c3740a/sarracenia/flow/__init__.py#L2814-L2827

petersilva:
I very much prefer what we are doing here (refusing to send incomplete data by refusing to try to send it at all if it is too big.)   Some clients might request truncation... sending the bytes until we get to the max size ... perhaps with *TRUNCATED* being the last bytes...

That's:
* a bit more complicated to implement.
* what is a correct checksum? (how do you know a product is correct if you cannot see all of it.)
* does sending a partial product actually help anyone... what if the product has issues later on...?

I think refusing to send (what we have now) is the better approach.  We may, after conversations with the client ... be obliged to implement truncation.

PR #1286, opened 2024-11-05
Add fileSizeMax to sender #1285

andreleblanc11:
The sender will now refuse to send files if `fileSizeMax` is set and if a message has a size greater then the `fileSizeMax` value.

```
'fileSizeMax': 32768,
...
2024-11-05 20:58:47,278 [INFO] sarracenia.flowcb.log after_accept accepted: (lag: 2.40 )  exchange: xpublic subtopic: 20241105.SSC-DATAINTERCHANGE.MSC-BULLETINS.20241105.SSC-DATAINTERCHANGE.MSC-BULLETINS.FP.KOKX.20 a file with baseUrl: http://ddsr-cmc-dev01.cmc.ec.gc.ca relPath: 20241105/SSC-DATAINTERCHANGE/MSC-BULLETINS/20241105/SSC-DATAINTERCHANGE/MSC-BULLETINS/FP/KOKX/20/FPUS51_KOKX_052058___234 sundew_extension: cvt_nws_bulletins-sr:KOKX:FP:3:Direct:20241105205843 id: OKcNKq5 size: 40834
2024-11-05 20:58:47,279 [WARNING] sarracenia setReport unknown report code supplied: 413:Payload Too Large http://ddsr-cmc-dev01.cmc.ec.gc.ca /20241105/SSC-DATAINTERCHANGE/MSC-BULLETINS/20241105/SSC-DATAINTERCHANGE/MSC-BULLETINS/FP/KOKX/20/FPUS51_KOKX_052058___234
```

I also accomodated the AM sender code to use `fileSizeMax` and attribute a default value if ever `fileSizeMax` is missing

### DIFF
--- a/sarracenia/flow/__init__.py
+++ b/sarracenia/flow/__init__.py
@@ -2829,6 +2829,10 @@ class Flow:
                 self.reject(msg, 422, f"new_file message field missing, do not know name of file to write. skipping." )
                 continue
 
+            if self.o.fileSizeMax > 0 and msg['size'] > self.o.fileSizeMax: 
+                self.reject(msg, 413, f"Payload Too Large {msg.getIDStr()}") 
+                continue
+
             # weed out non-file transfer operations that are configured to not be done.
             if 'fileOp' in msg:
                 if ('directory' in msg['fileOp']) and ('remove' in msg['fileOp']) and ( 'rmdir' not in self.o.fileEvents ):
--- a/sarracenia/flowcb/send/am.py
+++ b/sarracenia/flowcb/send/am.py
@@ -86,7 +86,8 @@ class Am(FlowCB):
 
         # Step out of the function if the bulletin size is too big
         if len(strdata) > self.o.fileSizeMax:
-            raise Exception(f"Bulletin length too long. Bulletin limit length: {self.o.MaxBulLen}. Latest bulletin length: {len(strdata)}. Path to bulletin: {msg_path}")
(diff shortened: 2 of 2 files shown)


petersilva (approved):
I'm fine with the sender logic...  the am... is a bit more debatable.
I don't think we should raise an exception.  Raising an exception will cause a retry.
You want to return None as the message, and have the thing check the return value from wrapmsg and return if it is None instead of trying to send...

It seems like any error in wrapmsg is going to result in a retry, but it isn't clear that anything that goes wrong there will go right at a subsequent attempt.

We should generate an error message, fine... but there is no reason to retry a failure that is permanent.

petersilva:
I accepted the change... so... um usually I would just merge at that point...
but I whined about am .also...  @andreleblanc11 do you want to get this merged now and work on AM some other time? or do you want to work on it, while you're already *in the area* ?

andreleblanc11:
That's something that I wanted to change, but I didn't really think about it while writing the PR. The PR was a bit of an end of the day rush.

I'll add it in tommorow.

andreleblanc11:
The only way the AM sender will try sending files greater then `fileSizeMax` is if the option is unset in the configuration.

We are already emitting a logger warning when this happens in the AM sender. 
So an analyst should theoretically see this, then add the option in the configuration (likely along with `retry_refilter True` to get all the current retries out of the loop).

I think this should be enough padding to avoid unnecessary retries.

PR #1292, opened 2024-11-12
Check if there is anything to retry before figuring out if we

petersilva:
have room to actually read from the queue.

If we don't check, we won't warn about being too busy.


reidsunderland (approved):
This is great, I was actually getting annoyed by this when reading logs last week :)

PR #1293, opened 2024-11-13
systemd integration hints were v2 oriented

petersilva:
correcting docs to have sr3 info.  removed support for upstart (non-systemd) init systems.

### DIFF
--- a/docs/source/How2Guides/Admin.rst
+++ b/docs/source/How2Guides/Admin.rst
@@ -196,35 +196,29 @@ gather statisical information.
 Init Integration
 ~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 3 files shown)


petersilva:
So I noticed this stuff is covered in:

HowTo/Admin.rst

and 

Tutorial/Install.rst

The stuff in Tutorial was upto date, the admin guide wasn't... wondering if it should be removed from one place... suggestions welcome.

They both include the cronjobs, and the systemd installation work needed  (or not.)

petersilva:
https://wiki.archlinux.org/title/Systemd/User  literature on usermode systemd

andreleblanc11 (approved):
French/English both look good

PR #1295, opened 2024-11-14
Issue339 try5 fail connections in main loop

petersilva:
This is a third pass at not looping to retry broker connections in the amqp and mqtt moth drivers.  They now attempt to connect once, and have an exponential backoff implementation to that they don't try to connect again too soon.

This is a preparatory change to, in future, being able to work with multiple broker connections, where if one fails the other still works.  Without this change, when a failure in one broker happens, it will just loop trying to re-connect to it, and not even look at the available good connection to the other broker.

previous attempts :  #1247 

The first two patches here are different, in that it was split into two, so that, in the first patch, you see the substantive changes made to getSetup and putSetup (removing the loop, making it return instead.) the second patch, just fixes the indenting and gets rid of the now dead code (mqtt/putSetup was missed on first pass, so both changes combined in second patch for that routine.)

The remainder of the patches implement ebo (exponential back off) of broker connections.

Ran the flow tests... seems fine.

### DIFF
--- a/sarracenia/moth/__init__.py
+++ b/sarracenia/moth/__init__.py
@@ -284,6 +284,10 @@ class Moth():
         self.metrics = { 'connected': False }
         self.metricsReset()
 
+        now = time.time()
+        self.next_connect_time = now
+        self.next_connect_failures = 0
+
         if (sys.version_info.major == 3) and (sys.version_info.minor < 7):
             self.o = {}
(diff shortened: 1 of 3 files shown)


reidsunderland (approved):
reviewed in meeting with @andreleblanc11  and @petersilva

PR #1304, opened 2024-11-25
exclude retries from lag accumulation

petersilva:


petersilva:
I saw these worrisome failures of a bunch of tests in the flakey_broker test... reproduced on my test machine, then reproduced the same thing with the development branch.  So it is pre-existing breakage, nothing to do with this change.

andreleblanc11 (approved):

PR #1308, opened 2024-11-28
Issue1299 - New `after_gather` entry point + better error handling in sundew_dirPattern

### LINKED ISSUE #1299: How to deal with sundew pattern matching when filename is ambiguous
I've been working on migrating Sundew feeds to sr3 and have encountered a problem when a filename doesn't have enough sundew parameters in it. 

## Example

We have a bulletin that is coming in with a quasi-ambiguous filename.

```
fq20qx.191324^
```

In our configuration (in my use case a *sarra* config), we want to download the data with the sundew pattern matching.

```
directory               ${PBD}/${YYYYMMDD}/SOMETHING/BULLETINS/${T1}${T2}/${CCCC}/${GG}
```

However, because the filename doesn't contain any of the information requested in `directory`, the config will crash

```
2024-11-19 19:14:03,343 [DEBUG] sarracenia.flow filter  urlToMatch: http://my-server/20241119/source/of/data/fq20qx.191324^
  File "/usr/lib/python3/dist-packages/sarracenia/instance.py", line 276, in <module>
    i.start()
  File "/usr/lib/python3/dist-packages/sarracenia/instance.py", line 267, in start
    self.running_instance.run()
  File "/usr/lib/python3/dist-packages/sarracenia/flow/__init__.py", line 600, in run
    self.filter()
  File "/usr/lib/python3/dist-packages/sarracenia/flow/__init__.py", line 1068, in filter
    self.updateFieldsAccepted(m, url, pattern, maskDir,
  File "/usr/lib/python3/dist-packages/sarracenia/flow/__init__.py", line 964, in updateFieldsAccepted
    new_dir = self.o.sundew_dirPattern(pattern, urlstr, tfname, new_dir)
  File "/usr/lib/python3/dist-packages/sarracenia/config/__init__.py", line 2233, in sundew_dirPattern
    nddword += self.sundew_matchPattern(BN, EN, BP, dwword, dwword)
  File "/usr/lib/python3/dist-packages/sarracenia/config/__init__.py", line 2257, in sundew_matchPattern
    elif keywd[:6] == "{CCCC}": return EN[1] + keywd[6:]
IndexError: list index out of range
```

***

This feature is necessary in my opinion, because we have lots of feeds from Sundew where files come in without a clear filename.
In sundew, I think the code looks at the file contents directly to determine the filename, when the type is `bulletin-file`.

andreleblanc11:
- Added a new flow callback entry point + documentation. 
   - I'm not 100% sure the location where I added `after_gather` is correct. I put it before `runCallbackPoll` is called because there's `if self.o.component != 'poll':` that comes just after. 
   - So far, it seems that only sarra configs benefit of `after_gather`. Placement might be fine.

- Better error handling of sundew_dirPattern by wrapping it in a try/except.
   - `updateFieldsAccepted` now returns a boolean.
 
- I updated the raw2bulletin renamer to use `relPath` instead of `new_file/new_dir`. There's no `new_dir/new_file` before the filtering.

- I've also updated the unit tests and have ran the AM flow test. Both look OK now.

### DIFF
--- a/docs/source/Explanation/SarraPluginDev.rst
+++ b/docs/source/Explanation/SarraPluginDev.rst
@@ -569,6 +569,17 @@ for detailed information about call signatures and return values, etc...
 |                     |                                                    |
 |                     |                                                    |
 +---------------------+----------------------------------------------------+
+|                     |                                                    |
+| after_gather        | Called after gather and before filter.             |
+| (self,worklist)     |                                                    |
+|                     | Not used often. after_accept should be used        |
+|                     | for most use cases.                                |
+|                     |                                                    |
+|                     | after_gather should only really be used when:      |
+|                     | - There needs to be a change to the worklist       |
+|                     |   of messages before attempting to filter.         |
+|                     |                                                    |
++---------------------+----------------------------------------------------+
 |                     | called after When a transfer has been attempted.   |
 | after_work          |                                                    |
 | (self,worklist)     | All messages are acknowledged by this point.       |
--- a/docs/source/How2Guides/FlowCallbacks.rst
+++ b/docs/source/How2Guides/FlowCallbacks.rst
@@ -17,6 +17,9 @@ the `sarracenia.flowcb <../../sarracenia/flowcb/__init__.py>`_ class.
 Briefly, the algorithm has the following steps:
 
 * **gather** -- passively collect notification messages to be processed.
+
+  * *after_gather* callback entry point
+
 * **poll** -- actively collect notification messages to be processed.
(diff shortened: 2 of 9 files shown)


petersilva (approved):

petersilva:
I approved it, but there are conflicts... can you fix them in your branch so it will apply cleanly?

andreleblanc11:
I thought I had already resolved the merge conflicts in https://github.com/MetPX/sarracenia/pull/1308/commits/af725c6eb58b4b3a244090be1f678495392d8837?

It doesn't look like there are any more conflicts

PR #1314, opened 2024-11-30
Issue339 consistent & persistent rand8 (only changes on declare and cleanup.)

petersilva:
So this is part of eventual multi queue support (which includes multi-broker support.) the RAND8 thing needs to be "reliable" evaluating to the same thing every time we resolve it.   The value only changes on a fresh declare (after a cleanup.)

This set of patches makes a rand8 state file in the state directory, allowing "self.o.rand8" to evaluate to the same thing regardless of how many times it is substituted.   Before this change, in RAND8 showed up in a variable, it would get substituted by a different value every time it got substituted.

The config parser reads the files at least twice... sometimes more often, so it won't result in the same values with each pass.   Second time you resolve a queue name that has RAND8 in it, you get a different value from the first time.  

Until now, there was only one queue. The queue name would get evaluated once and then the resolved value would be used later.  This is difficult to work with in the context of things like "exchangeSplit" where we need to create 1 queue per instance... so it needs to be evaluated n times, and each queue would have a different RAND8.  

 I guess that could work... having each queue with completely different RAND8 values, but it would be confusing... it would be much more obvious if they all used the same RAND8 value.  

Anyways... the idea is to have a more consistent RAND8 value.

The other thing this patch set does is detect in previous versions when there is a RAND8 value already used (in the queueName) and copy it into the rand8 statefile so it is understood by current  and future versions.

### DIFF
--- a/docs/source/Reference/sr3_options.7.rst
+++ b/docs/source/Reference/sr3_options.7.rst
@@ -140,8 +140,8 @@ option, with the use of *${..}* notation:
 * PROGRAM     - the name of the component (subscribe, shovel, etc...)
 * CONFIG      - the name of the configuration file being run.
 * HOSTNAME    - the hostname running the client.
-* RANDID      - a random id (0-64Ki) that will be consistent within a single instance.
-* RAND8       - a random 8 digit wide number that is generated whenever it is evaluated in a string.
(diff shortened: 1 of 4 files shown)


petersilva:
Not sure this is a good idea... working on another approach.

petersilva:
dropping for now.

PR #1324, opened 2024-12-05
add sdnotify support to cli, for Systemd

### LINKED ISSUE #1318: systemd integration limited: systemctl status does not work.
on older operating systemd (redhat8, ubuntu 18.04) at least... the systemd versions do not check for status correctly for sr3.  There is some incompatibility.  e.g.:

```
sarra@host:~$ systemctl status metpx-sr3
● metpx-sr3.service - Sarracenia File Copy Service
   Loaded: loaded (/lib/systemd/system/metpx-sr3.service; enabled; vendor preset: enabled)
   Active: inactive (dead) since Tue 2024-12-03 19:03:55 UTC; 21min ago
  Process: 36353 ExecStop=/usr/bin/sr3 stop (code=exited, status=0/SUCCESS)
  Process: 2889 ExecStart=/usr/bin/sr3 start (code=exited, status=0/SUCCESS)

Dec 03 12:24:43 host systemd[1]: Starting Sarracenia File Copy Service...
Dec 03 12:24:58 host sr3[2889]: starting:.........................( 248 ) Done
Dec 03 12:24:58 host systemd[1]: Started Sarracenia File Copy Service.
Dec 03 19:03:55 host sr3[36353]: Stopping: no procs running...already stopped
sarra@host:~$
```

So one must use *sr3 status* to check properly.

Also monitoring is done by having *sr3 sanity* run periodically using cron (e.g. every 5 minutes.) which will restart crashed processes, and kill strays should any be present.

There is no known approach to fix this at the moment... but newer OS versions seem to understand sr3 better (e.g. ubuntu 24.04.) this problem may go away with time.

petersilva:
fixes #1318 systemd recommends against Type=forking, preferring TYPE=notify. To do that, we need to make little api calls during startup and shutdown. Added those.

This makes systemd happier on startup and seems to show status messages correctly.

### DIFF
--- a/debian/metpx-sr3.service
+++ b/debian/metpx-sr3.service
@@ -11,7 +11,7 @@ After=network-online.target local-fs.target remote-fs.target
 Requires=network-online.target
 
 [Service]
-Type=forking
+Type=Notify
 ExecStart=/usr/bin/sr3 start
 User=sarra
 Group=sarra
--- a/debian/metpx-sr3.service
+++ b/debian/metpx-sr3.service
@@ -11,7 +11,7 @@ After=network-online.target local-fs.target remote-fs.target
 Requires=network-online.target
 
 [Service]
-Type=Notify
+Type=notify
 ExecStart=/usr/bin/sr3 start
 User=sarra
 Group=sarra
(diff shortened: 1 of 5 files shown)


petersilva:
This does not work... existing development branch actually works better on 22.04 and 24.04... dropping this.

PR #1329, opened 2024-12-05
Issue1322 post exchange split revamp

### LINKED ISSUE #1297: hard ordering constraints in HPC mirroring
See Corresponding C implementation issue for background: https://github.com/MetPX/sarrac/issues/174 

Things that need to be done to robustly support that on the python side:

* ${INSTANCE} ...  as per #1296 
* ${NODE} ... this is new...

If we want to implement exchangeSplit properly, then as pointed out here: https://github.com/MetPX/sarrac/issues/174#issuecomment-2476906203

We need to establish instances over an entire cluster, not just a single node, in order to get singleton processing working properly.

e.g... a winnow publishes to 20 exchanges... we have 4 consuming nodes, each with 5 instances... we would want the bindings to look something like:

x1 -> n1i1, x2->n1i2, ...  x5-> n1i5, x6->n2i1 ... x10 -> n2i5, x11 -> n3i1, ... x15 -> n3i5 ....

or another mapping would be to have subsets of instances on each node...

x1 -> n1i1 .... x6 -> n2i6,  x11->n3i11 ... 

Would have to do the math one way or another. and create the exchanges, queues and bindings.

### LINKED ISSUE #1322: sr3 sender fails to post message when post_exchangeSplit is in use and the incoming message does not have a checksum
```
-  File "/usr/lib/python3/dist-packages/sarracenia/flow/__init__.py", line 1253, in post
-    p(self.worklist)
-  File "/usr/lib/python3/dist-packages/sarracenia/flowcb/post/message.py", line 40, in post
-    if all_good and hasattr(self.poster,'putNewMessage') and self.poster.putNewMessage(m):
-  File "/usr/lib/python3/dist-packages/sarracenia/moth/amqp.py", line 665, in putNewMessage
-    idx = sum( bytearray(body['identity']['value'],
-KeyError: 'identity'
```

https://github.com/MetPX/sarracenia/blob/a4af0e2c21f211e1d993c3a183b8eed63dbfb09c/sarracenia/moth/amqp.py#L675-L682

In a sender, we should probably ensure that all messages have an identity. If we receive one that doesn't have identity, we should calculate it.

petersilva:
fix #1322 
partially fix #1297 

The post_exchangeSplit was originally conceived as a means of scaling duplicate suppression beyond what a single process could support. Duplicates were defined as files that had the same checksum, so splitting them among exchanges based on the checksum is what made sense initially.  This method of load distribution will assign the same file path to different exchanges if they are not identical (don't have the same checksum.)

As work with mirroring has proceeded, it turns out that a file is really a stream of changes to a given path, and to make good decisions about it, you want a single process to receive all the updates for a given path.  if a file is modified and then removed... the removal event would not necessarily have a related checksum to the modification.  Similarly for a symbolic link. 

Changing the hasing algorithm to be based on path name, rather than checksum:

*  If files have the same path, and the same checksum... they are still duplicates. so that still works.
* If files have the same path, but different checksums... fine it should be processed same as today, just by the same participant in the exchangeSplit pool of flows.
* If files have the same path but different type (link, directory, file... remove event.) they should now show up in the same participant in the exchangeSplit pool of flows.
* it was a sorry that when hashing based on checksums, we hamper the effectiveness of the duplicate suppresion path by removing some hashes from it, resulting in an unbalanced tree.  hashing on path is likely to yield better/faster associative arrays for duplicate suppression within each participant in the exchangeSplit pool of flows.

### DIFF
--- a/docs/source/Explanation/CommandLineGuide.rst
+++ b/docs/source/Explanation/CommandLineGuide.rst
@@ -2247,7 +2247,7 @@ post_exchangeSplit   <number>   (default: 0)
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 The **post_exchangeSplit** option appends a two digit suffix resulting from 
-hashing the last character of the checksum to the post_exchange name,
+hashing the path to the post_exchange name,
 in order to divide the output amongst a number of exchanges.  This is currently used
 in high traffic pumps to allow multiple instances of winnow, which cannot be
 instanced in the normal way.  Example::
--- a/docs/source/How2Guides/Admin.rst
+++ b/docs/source/How2Guides/Admin.rst
@@ -146,7 +146,7 @@ winnow is used to suppress duplicates.
 first time a posting is received, it could be picked by one instance, and if a duplicate one is received
 it would likely be picked up by another instance. **For effective duplicate suppression with instances**,
 one must **deploy two layers of subscribers**. Use a **first layer of subscribers (shovels)** with duplicate
-suppression turned off and output with *post_exchangeSplit*, which route posts by checksum to
+suppression turned off and output with *post_exchangeSplit*, which route posts by path to
 a **second layer of subscribers (sr_winnow) whose duplicate suppression caches are active.**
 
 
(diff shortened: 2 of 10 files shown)


reidsunderland (approved):
This looks good, it's a draft so i'm not sure if i was supposed to look at it yet

PR #1330, opened 2024-12-05
sender options inflight test fails if None

### LINKED ISSUE #1327: `inflight None` in sender causes messages to get retried
If you have a sender that doesn't have `inflight` set, messages will get retried. This is the exception message we are getting in the logs

```
sarracenia.flow send Exception details:
  File "/net/local/home/leblanca/.local/lib/python3.10/site-packages/sarracenia/flow/__init__.py", line 2418, in send
    'rename') and options.inflight.startswith('.'):
AttributeError: 'NoneType' object has no attribute 'startswith'
```

The `inflight` value defaults to `None`. 

```
sr3 show sender/config-name | grep inflight
'inflight': None
```

This also conflicts with what the documentation states? It says the inflight defaults to `.tmp` when `post_baseDir` isn't set.
https://metpx.github.io/sarracenia/Reference/sr3_options.7.html#inflight-string-default-tmp-or-none-if-post-broker-set

Regardless, `inflight None` shouldn't be failing.

petersilva:
fix #1327 

just check if option.inflight is not None... if it is None, we don't need to rename after transfer... so it is moot.

### DIFF
--- a/docs/source/Explanation/FileCompletion.rst
+++ b/docs/source/Explanation/FileCompletion.rst
@@ -23,16 +23,16 @@ Inflight Table
 |   NONE      |Send `sr3_post(7) <sr3_post.7.rst>`_   |post only when file is complete       |
 |             |by AMQP after file is complete.        |                                      |
 |             |                                       |(Best when available)                 |
-|             | - fewer round trips (no renames)      | - Default on sr_sarra.               |
-|             | - least overhead / highest speed      | - Default on sr_subscribe and sender |
-|             |                                       |   when post_broker is set.           |
+|             | - fewer round trips (no renames)      |                                      |
+|             | - least overhead / highest speed      | DEFAULT                              |
+|             |                                       |                                      |
 +-------------+---------------------------------------+--------------------------------------+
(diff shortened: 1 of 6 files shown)


petersilva:
There is this weird breakage of the flakey on the deve branch for the past week or so...
It is still broken here... thought I had fixed it, but no.

andreleblanc11 (approved):

reidsunderland:
I was taking a close look at this because I was going to use inflight None in a config that I wanted to deploy with 3.00.56 and @andreleblanc11 pointed me to this issue when he reviewed my change.

I noticed a potential problem with the fix.

https://github.com/MetPX/sarracenia/blob/34031df0af0557c13cae6385ae864a649c9272f6/sarracenia/flow/__init__.py#L2470-L2486

Assuming there is a protocol that does support umask and does not support renaming, the new code will set `inflight = None`, which would cause `inflight umask` in the config to be ignored.

But I think FTP is the only protocol that supports umask, and it does support rename as well, so it shouldn't really matter in practice.

Issue #1289, opened 2024-11-07
Possible feature request: additional fields in accept statements that can be used by plugins


reidsunderland:
I've run into a couple of cases where it would be very helpful to let a user define additional fields to the right of accept statements, similar to how we have DESTFN now.

For example, I have this config:

```
accept      .*${HOSTNAME}.*/loggernet/.*T_Out.*.dat
accept      .*${HOSTNAME}.*/loggernet/.*T_TBRG.*.dat
accept      .*${HOSTNAME}.*/loggernet/.*T_RF1.*.dat
accept      .*${HOSTNAME}.*/loggernet/.*T_StnConfig.*.dat
accept      .*${HOSTNAME}.*/loggernet/.*final_storage_.*.dat

rawca2b_map loggernet/||CACN65 CWAO||(...)_T_Out.*.dat|(...)_T_TBRG.*.dat|(...)_T_RF1.*.dat|(...)_T_StnConfig.*.dat|(...)_final_storage_.*.dat
```

The fields are separated by `||`. The first field is a string that is mostly redundant with the accept statements. The second field is the bulletin header that we want. The third field is a regex that identifies where the 3 character station ID is in the filename. It would be more user friendly (for both the plugin developer and config file author) to integrate this with the accept statements.

Like this:

```
accept      .*${HOSTNAME}.*/loggernet/.*(...)_T_Out.*.dat            header=CACN65 CWAO
accept      .*${HOSTNAME}.*/loggernet/.*(...)_T_TBRG.*.dat           header=CACN65 CWAO
accept      .*${HOSTNAME}.*/loggernet/.*(...)_T_RF1.*.dat            header=CACN65 CWAO
accept      .*${HOSTNAME}.*/loggernet/.*(...)_T_StnConfig.*.dat      header=CACN65 CWAO
accept      .*${HOSTNAME}.*/loggernet/.*(...)_final_storage_.*.dat   header=CACN65 CWAO
```

I'm not sure how hard this would be to implement, and it's not a high priority. I'm just submitting this issue because it might be something worth looking into someday.

petersilva:
*Header* is confusing in this context... the fields of a message are called "headers" like when someone wants to post a message with a custom header, they use:  *--header &lt;key>=&lt;value>* ... but I think you mean the *key* is 'header'  and something is putting a field in the message called *header* whose value is the WMO AHL (Abbreviated Header Line) of the message.

if someone were posting the message it would be...
```

[...]

PR #1338, opened 2024-12-10
Refactor email plugin config syntax

reidsunderland:
This is a proposal to change the email plugin configuration syntax to take advantage of the additional accept statement "arguments" added in #1289. 

Previously, the email plugin had to do it's own regex matching/filtering to determine the recipients for a particular file. Now, it just gets the recipients from the mask, removing this redundancy.

I made a copy, called `email_old` before making modifications. I thought it might be useful if someone didn't want to update their config file. 

Old syntax:

```
callback send.email_old

file_email_to AACN27 user@example.com, test@test.com
file_email_relay email.relay.server.ca
file_email_from santa@canada.ca

accept .*AACN27.*
```

New syntax:

```
callback send.email

sendTo smtp://email.relay.server.ca
# sendTo email.relay.server.ca  also works

# optional
email_from santa@canada.ca

email_subject_prepend  Sent by Sarracenia: 

accept .*AACN27.*  user@example.com, test@test.com
```

I also added a new option, `email_subject_prepend`, because we have some other email plugins internally that add extra text in the subject. I wanted this to be able to replace all of our email plugins, and eliminate hard-coded configuration in the plugin code.


reidsunderland:
Oh, and it works in a sender or subscriber/sarra.

petersilva (approved):
This is a great improvement!

We tend to accumulate cruft (v1, 2, 3...) when we keep old stuff around... we have git for that. Would prefer to delete email_old, and have everyone change over... but could be done as second step for transition purposes... anyways not worth rejecting for that.

reidsunderland:
I looked at our configs and nothing is currently using the current/old email plugin, so I removed email_old. There's one config on dev and stage that we will update once we release the new sr3 version.

PR #1342, opened 2024-12-12
Many small issues

### LINKED ISSUE #1340: Problems with the rename option
# Part 1
The `rename` option does not seem to work in a sarra.

I tried using it both before and after accept statements.

```
rename ${%o-6d%Y%m%d}00.grib1

directory ${PBD}/${%Y%m%d}/${SOURCE}/
accept .*grib.*

rename ${%Y%m%d}00.grib1
```

# Part 2

In a poll, it crashes.

```
  File "/local/home/sarra/sr3/sarracenia/instance.py", line 276, in <module>
    i.start()
  File "/local/home/sarra/sr3/sarracenia/instance.py", line 267, in start
    self.running_instance.run()
  File "/local/home/sarra/sr3/sarracenia/flow/__init__.py", line 591, in run
    self.gather()
  File "/local/home/sarra/sr3/sarracenia/flow/__init__.py", line 1160, in gather
    self._runCallbackPoll()
  File "/local/home/sarra/sr3/sarracenia/flow/__init__.py", line 386, in _runCallbackPoll
    new_incoming = plugin()
  File "/local/home/sarra/sr3/sarracenia/flowcb/poll/__init__.py", line 584, in poll
    msgs.extend(self.poll_directory(currentDir))
  File "/local/home/sarra/sr3/sarracenia/flowcb/poll/__init__.py", line 449, in poll_directory
    msgs.extend(self.poll_list_post(pdir, desclst, filelst))
  File "/local/home/sarra/sr3/sarracenia/flowcb/poll/__init__.py", line 552, in poll_list_post
    new_msgs = self.poll_file_post(desc, destDir, remote_file)
  File "/local/home/sarra/sr3/sarracenia/flowcb/poll/__init__.py", line 508, in poll_file_post
    msg = sarracenia.Message.fromFileInfo(post_relPath, self.o, desc)
  File "/local/home/sarra/sr3/sarracenia/__init__.py", line 742, in fromFileInfo
    msg['retrievePath'] = msg['new_retrievePath']
KeyError: 'new_retrievePath'
```

https://github.com/MetPX/sarracenia/blob/9cc7d8c616e77f3e6bb51e9994e8b1a93f2c500a/sarracenia/__init__.py#L750-L754

I'm not quite sure what's happening with retrievePath there. Maybe it's not necessary? Rename shouldn't change the relPath or retrievePath when the file is being posted

petersilva:
close #1340 

Three other defects found in testing by André and Peter.  (no corresponding issue.)

### DIFF
--- a/sarracenia/__init__.py
+++ b/sarracenia/__init__.py
@@ -748,7 +748,10 @@ class Message(dict):
         # rename path given with no filename
 
         if o.rename:
-            msg['retrievePath'] = msg['new_retrievePath']
+            # ensure we have the old name for retrieval usage.
+            if 'new_retrievePath' in msg:
+                msg['retrievePath'] = msg['new_retrievePath']
+
             newname = o.variableExpansion(o.rename)
             if o.rename[-1] == '/':
                 newname += os.path.basename(path)
--- a/sarracenia/flow/__init__.py
+++ b/sarracenia/flow/__init__.py
@@ -1241,8 +1241,11 @@ class Flow:
                     and (m['relPath'] != m['new_relPath']):
                 m['old_relPath'] = m['relPath']
                 m['_deleteOnPost'] |= set(['old_relPath'])
+
+            if 'new_relPath' in m:
                 m['relPath'] = m['new_relPath']
-                m['old_subtopic'] = m['subtopic']
+                if 'subtopic' in m:
+                    m['old_subtopic'] = m['subtopic']
                 m['_deleteOnPost'] |= set(['old_subtopic','subtopic'])
                 m['subtopic'] = m['new_subtopic']
 
--- a/sarracenia/flow/__init__.py
+++ b/sarracenia/flow/__init__.py
@@ -1237,7 +1237,8 @@ class Flow:
                     else:
                         m['new_relPath'] = m['new_file']
 
-            if ('new_relPath' in m) and (m['relPath'] != m['new_relPath']):
+            if ('new_relPath' in m) and ('relPath' in m) \
+                    and (m['relPath'] != m['new_relPath']):
                 m['old_relPath'] = m['relPath']
                 m['_deleteOnPost'] |= set(['old_relPath'])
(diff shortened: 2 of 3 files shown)


andreleblanc11:
I think the rest looks good.

sarracenia/__init__.py
```diff
@@ -748,7 +748,10 @@ def fromFileInfo(path, o, lstat=None):
         # rename path given with no filename
 
         if o.rename:
-            msg['retrievePath'] = msg['new_retrievePath']
+            # ensure we have the old name for retrieval usage.
+            if 'new_retrievePath' in msg:
+                msg['retrievePath'] = msg['new_retrievePath']
+
```
reidsunderland, line 754: I'm still confused about why the retrievePath code is inside `if o.rename`? I might be overlooking something, but I think the two things are independent? Rename is telling the subscriber what name to give the file when it downloads it, retrievePath tells the subscriber where to download from. But retrievePath should work whether or not rename is set?

petersilva: I don't understand it either... I didn't think that far.  I think I am confused about how rename works... relation between *relPath* and *rename* fields... e.g. should rename overwrite *relPath*? in which case, the actual path you use to download should be *retrievePath* .  ... eg.. one way to implement rename... is to copy relPath to retrievePath, and then apply the rename to the relPath and download normally... I don't recall if that's how things ended up or not...

petersilva: I think the implementation vacilated back and forth between changing *relPath* and using the *rename* field... so there are leftovers of each... and can switch between the two easily.

Advantage of *relPath* approach is can specify entire directory tree at the same time...with *rename* it is only the file name.

petersilva: so I think it is correct as-is... for some use cases... but would be a lot more work to explore thoroughly...

reidsunderland (approved):
I mentioned on teams that I think this is good to merge, but I'd like to explore rename/retrievePath again in the future

PR #1343, opened 2024-12-13
fix #1283 disable new config

### LINKED ISSUE #1283: `sr3 disable` doesn't work when a config is first created
# How to reproduce

- Create a new config
- Try to run `sr3 disable`
- When you run the command, there's no stdout output.
- When you check status after, the config shows as `stopped`

# Work around
- Create the cache directory either manually `mkdir -p ~/.cache/sr3/component/config` or (likely) run `sr3 declare component/config`

This isn't a bug but more of an inconvience for analysts that have created a new config but don't want to use it right away

petersilva:
close #1283 

In v2... "disable" use to be done by moving the config file to .off.  this was changed to creating a state file.
but the variables still had "cfg" in them. so this fix is a little more involved.

* determine the statedir (adding support for *statehost* which was missing.)
* create the statedir if it is missing.
* change all the variables to remove _cfg from the name... since not really appropriate anymore.

### DIFF
--- a/sarracenia/sr.py
+++ b/sarracenia/sr.py
@@ -1608,15 +1608,25 @@ class sr_GlobalState:
                 logging.error("cannot disable %s while it is running! " % f)
                 continue
 
-            state_file_cfg = self.user_cache_dir + os.sep + c + os.sep + cfg
-            state_file_cfg_disabled = state_file_cfg + os.sep + 'disabled'
-            if os.path.exists(state_file_cfg_disabled):
+            if self.configs[c][cfg]['options'].statehost:
+                state_file_dir = self.user_cache_dir + os.sep + self.hostdir + os.sep + f.replace('/', os.sep)
+            else:
+                state_file_dir = self.user_cache_dir + os.sep + f.replace('/', os.sep)
+
+            logger.critical( f"{state_file_dir=}" )
+            if not os.path.isdir(state_file_dir):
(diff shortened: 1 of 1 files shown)


sarracenia/sr.py
```diff
@@ -1608,15 +1608,25 @@ def disable(self):
...
 
-            state_file_cfg = self.user_cache_dir + os.sep + c + os.sep + cfg
-            state_file_cfg_disabled = state_file_cfg + os.sep + 'disabled'
-            if os.path.exists(state_file_cfg_disabled):
+            if self.configs[c][cfg]['options'].statehost:
+                state_file_dir = self.user_cache_dir + os.sep + self.hostdir + os.sep + f.replace('/', os.sep)
+            else:
+                state_file_dir = self.user_cache_dir + os.sep + f.replace('/', os.sep)
+
+            logger.critical( f"{state_file_dir=}" )
```
reidsunderland: i think this is probably a leftover from debugging? `logger.critical( f"{state_file_dir=}" )` :)

reidsunderland: there's another one on line 1621 as well. 

Otherwise I think it's good

petersilva: nice catch... those will break on 18.04... 
will fix.

reidsunderland (approved):

PR #1356, opened 2024-12-20
Fix DiskQueue bugs

reidsunderland:
This seems to make the flakey test somewhat reliable again, even without the poll changes that I'm still experimenting with separately.

### DIFF
--- a/.github/workflows/unit-test.yml
+++ b/.github/workflows/unit-test.yml
@@ -19,7 +19,7 @@ permissions:
 
 jobs:
   build:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
 
     name: Unit test
     timeout-minutes: 10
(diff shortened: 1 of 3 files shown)


reidsunderland:
i also changed the unit test github actions to run on Ubuntu 24.04. This fixed these errors: `AttributeError: module 'lib' has no attribute 'X509_V_FLAG_NOTIFY_POLICY'. Did you mean: 'X509_V_FLAG_EXPLICIT_POLICY'?`

But the unit tests are failing (not related to the diskqueue)

petersilva (approved):
I ran the flakey_test on ubuntu 24, ubuntu18.04, and redhat8 and it passed perfectly on all of them.

I love the unit tests...

reidsunderland:
I made the unit tests work with the old pre-June code, then checked out the June changes and confirmed that it broke my new tests, then fixed the diskqueue code until the unit tests passed again :)

PR #1362, opened 2025-01-06
Issue1351 fixing problem getting local checksums of data that has *arbitrary* identity method

### LINKED ISSUE #1351: Can't download files when file already exists locally and message checksum is set to `arbitrary`
When a message comes in with an `arbitrary` checksum, and we want to download this file in a location where
1. The filename already exists, but
2. The checksum is different from that of the incoming message.

The file gets labeled as a duplicate and is rejected.

```
2024-12-18 16:01:21,372 [DEBUG] sarracenia.flow file_should_be_downloaded checksum in message: {'method': 'arbitrary', 'value': 'f47363343f9330bc36652458255f17ab'} vs. local: {'method': 'a
rbitrary', 'value': 'f47363343f9330bc36652458255f17ab'}
2024-12-18 16:01:21,372 [DEBUG] sarracenia.flow file_should_be_downloaded checksum in message: {'method': 'arbitrary', 'value': '3fb870ca02dd4040c0c3c6ba0237a9e8'} vs. local: {'method': 'a
rbitrary', 'value': '3fb870ca02dd4040c0c3c6ba0237a9e8'}
2024-12-18 16:01:21,372 [INFO] sarracenia.flowcb.log after_work rejected: 304 same checksum /apps/sarra/public_data/20241218/WXO-DD-STAGE/observations/swob-ml/latest/CPCI-AUTO-minute-swob.
xml
2024-12-18 16:01:21,372 [INFO] sarracenia.flowcb.log after_work rejected: 304 same checksum /apps/sarra/public_data/20241218/WXO-DD-STAGE/observations/swob-ml/latest/CWEW-AUTO-minute-swob.
xml
2024-12-18 16:01:21,372 [INFO] sarracenia.flowcb.log after_work rejected: 304 same checksum /apps/sarra/public_data/20241218/WXO-DD-STAGE/observations/swob-ml/latest/CZGH-AUTO-minute-swob.
xml
2024-12-18 16:01:21,372 [INFO] sarracenia.flowcb.log after_work rejected: 304 same checksum /apps/sarra/public_data/20241218/WXO-DD-STAGE/observations/swob-ml/latest/CESB-AUTO-minute-swob.
xml
2024-12-18 16:01:21,372 [INFO] sarracenia.flowcb.log after_work rejected: 304 same checksum /apps/sarra/public_data/20241218/WXO-DD-STAGE/observations/swob-ml/latest/CWZW-AUTO-minute-swob.
xml
2024-12-18 16:01:21,372 [INFO] sarracenia.flowcb.log after_work rejected: 304 same checksum /apps/sarra/public_data/20241218/WXO-DD-STAGE/observations/swob-ml/latest/CWOK-AUTO-minute-swob.
xml
2024-12-18 16:01:21,372 [INFO] sarracenia.flowcb.log after_work rejected: 304 same checksum /apps/sarra/public_data/20241218/WXO-DD-STAGE/observations/swob-ml/latest/CVOP-AUTO-minute-swob.
xml
```

For my use case, I found, that `xattr` wasn't installed when I did a `sr3 features`. 
I installed it and see'd if that helped, but some files were still being labeled as duplicates.

I think something might be wrong in `compute_local_checksum`

https://github.com/MetPX/sarracenia/blob/9b2ff58211f3fbbca6ec1a4f60ceb155f46077bc/sarracenia/flow/__init__.py#L1430-L1454

I've added debugging in the source and found that line 1439 returns as False. 
When this happens, this line of code is executed and will always set an `arbitrary` checksum as a duplicate.

https://github.com/MetPX/sarracenia/blob/9b2ff58211f3fbbca6ec1a4f60ceb155f46077bc/sarracenia/flow/__init__.py#L1459-L1460

I think we might need to change the `>=` to a `=<`? If the file at the destination already exists, the mtime should be lesser than that of the file we're trying to download.

# Work around
We can work around this by overriding the checksum with `identity cod,sha512` however this isn't good practice, because we can't truely validate the integrity of the file that comes from the source.

petersilva:
close #1351 

This branch might do the right thing.  It passes the flow tests, but I don't think that exercises the affected code paths
the same way the issue does.  It would be good to try it out.

### DIFF
--- a/sarracenia/flow/__init__.py
+++ b/sarracenia/flow/__init__.py
@@ -1427,7 +1427,25 @@ class Flow:
 
         return True
 
-    def compute_local_checksum(self, msg) -> None:
+    def compute_local_checksum(self, msg, lstat=None) -> None:
+        """
+            For a file whose path is given by the msg, calculate 'local_identity' field.
+
+            when checksums for files are stored in extended attributes, it's ideal to retrieve them,
+            rather than having to read the entire file again and re-calculate.
+
+            The extended attributes have a field:
+              * 'identity' ... the field from the message when the file was written.
+              * 'mtime' ... the mtime of the file when it was written.
+
+            The 'identity' extended attribute should be correct/good/useful if:
+               * the mtime of the file on disk is not newer than the current file mtime.
+
+            If the file has been over-written afterwards, the mtime will be different,
+            and the local checksum must be re-calculated from scratch.
+
+            If the checksum method is arbitrary, no local recalculation is possible.
+        """
 
         if sarracenia.filemetadata.supports_extended_attributes:
             try:
@@ -1436,7 +1454,7 @@ class Flow:
 
                 if s:
                     metadata_cached_mtime = x.get('mtime')
-                    if ((metadata_cached_mtime >= msg['mtime'])):
+                    if (lstat and (metadata_cached_mtime >= lstat.st_mtime)):
                         # file has not been modified since checksum value was stored.
 
                         if (( 'identity' in msg ) and ( 'method' in msg['identity']  ) and \
@@ -1453,11 +1471,12 @@ class Flow:
             except:
                 pass
 
+        # no local recalculation possible.
+        if msg['identity']['method'] in [ 'arbitrary' ]:
+            return
+
         local_identity = sarracenia.identity.Identity.factory(
             msg['identity']['method'])
-
-        if msg['identity']['method'] == 'arbitrary':
-            local_identity.value = msg['identity']['value']
 
         local_identity.update_file(msg['new_path'])
         msg['local_identity'] = {
@@ -1468,7 +1487,7 @@ class Flow:
 
     def file_should_be_downloaded(self, msg) -> bool:
         """
-          determine whether a comparison of local_file and message metadata indicates that it is new enough
+          Determine whether a comparison of local_file and message metadata indicates that it is new enough
           that writing the file locally is warranted.
 
           return True to say downloading is warranted.
@@ -1555,16 +1574,19 @@ class Flow:
             return True
 
         try:
-            self.compute_local_checksum(msg)
+            self.compute_local_checksum(msg,lstat)
         except:
             logger.debug(
                 "something went wrong when computing local checksum... considered different"
             )
             return True
 
-        logger.debug( f"checksum in message: {msg['identity']} vs. local: {msg['local_identity']}" )
+        if 'local_identity' in msg:
+           logger.debug( f"checksum in message: {msg['identity']} vs. local: {msg['local_identity']}" )
+        else:
+           logger.debug( f"checksum in message: {msg['identity']} vs. local: None" )
 
-        if msg['local_identity'] == msg['identity']:
+        if 'local_identity' in msg and msg['local_identity'] == msg['identity']:
             self.reject(msg, 304, f"same checksum {msg['new_path']}" )
             return False
         else:
--- a/sarracenia/flow/__init__.py
+++ b/sarracenia/flow/__init__.py
@@ -1454,7 +1454,8 @@ class Flow:
 
                 if s:
                     metadata_cached_mtime = x.get('mtime')
-                    if (lstat and (metadata_cached_mtime >= lstat.st_mtime)):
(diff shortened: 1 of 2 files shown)


sarracenia/flow/__init__.py
```diff
@@ -1555,16 +1574,19 @@ def file_should_be_downloaded(self, msg) -> bool:
...
             logger.debug(
                 "something went wrong when computing local checksum... considered different"
             )
             return True
 
-        logger.debug( f"checksum in message: {msg['identity']} vs. local: {msg['local_identity']}" )
+        if 'local_identity' in msg:
+           logger.debug( f"checksum in message: {msg['identity']} vs. local: {msg['local_identity']}" )
+        else:
+           logger.debug( f"checksum in message: {msg['identity']} vs. local: None" )
```
andreleblanc11, line 1588: Maybe this log should be a warning? There would be no way of validating if the file is a duplicate.

andreleblanc11:
I tried testing the code and it works, in the sense that the files with arbitrary checksums aren't reported as duplicates anymore, and are getting downloaded. However, I'm surprised that the `local` checksum still isn't being found from the xattr data

```
# mtime reports the same value
>>> os.getxattr('/data/wxofeed/wxapp/dms-stab/hourly_obs/6092919.yze.hourly_obs.en.json', 'user.sr_mtime')
b'20250106T211105.850807667'
>>> os.getxattr('/data/wxofeed/wxapp/dms-stab/hourly_obs/6092919.yze.hourly_obs.en.json', 'user.sr_identity')
b'{"method": "arbitrary", "value": "1b875c0bcfee820a146847e0cccbe721"}'
>>> os.path.getmtime('/data/wxofeed/wxapp/dms-stab/hourly_obs/6092919.yze.hourly_obs.en.json')
1736197865.8508077

# But the local checksum still can't be found, despite this
21:11:05,836 [DEBUG] sarracenia.flow file_should_be_downloaded checksum in message: {'method': 'arbitrary', 'value': '1b875c0bcfee820a146847e0cccbe721'} vs. local: None
```

andreleblanc11:
I've found out why the local checksum isn't being calculated in my use case. The mtime reported from `lstat` is in unix time, but the xattr mtime is in a separate format (the sarracenia format?).

```
# Values
2025-01-07 13:23:24,601 [CRITICAL] sarracenia.flow compute_local_checksum xattr mtime : 20250107T122207.797049761. lstat mtime : 1736252527.7970498
# Format
2025-01-07 13:28:08,137 [CRITICAL] sarracenia.flow compute_local_checksum xattr mtime : <class 'str'>. lstat mtime : <class 'float'>
```

This will always fail in the code, and because we have a parent try/except, where the except has `pass`, the error is never reported in the logs.

andreleblanc11:
I've added a commit. We're now comparing in the same sarracenia date string format.
The local checksum is now being registered from the extended attributes.
```
2025-01-07 13:47:11,681 [CRITICAL] sarracenia.flow compute_local_checksum xattr stats : 20250107T130136.863848209. File mtime : 20250107T130136.863848209
2025-01-07 13:47:11,681 [DEBUG] sarracenia.flow file_should_be_downloaded checksum in message: {'method': 'arbitrary', 'value': 'cc0bf34044dc29993f563f6baed
bfbb8'} vs. local: {'method': 'arbitrary', 'value': '3d07c1e26f93fbd7840a7c0038341c49'}
```

sarracenia/flow/__init__.py
petersilva: So... In sr3 it's allowed to publish things without checksums.  There were changes made a year ago or more (refactoring that resulted in FileOp field) that made integrity fields optional.  The identity fields are good practice, they are really helpful for a number of duplicate suppression and high availability, but they require the data producer to support them.

What DMS does actively defeats the purpose of checksums, and there is nothing we can do to fix that.  Since DMS is the data producer responsible for a heck of a lot of products, we will start spewing out literally millions of "warnings" that nobody can do anything about.

It's just working as designed.  A warning makes sense if there is something the user can do to fix it... but here, there is nothing anyone can do about it.  It's just annoying for everyone.

petersilva:
So @andreleblanc11 's patch is good imo.  Can somebody review again... so we can merge it?

reidsunderland (approved):

andreleblanc11 (approved):

PR #1365, opened 2025-01-08
review and improve error handling when invoking makedirs

petersilva:
client ran into some error/crashes related to makedirs in downloading.
Reviewed all invocations.
  * if the file exists (but isn't a directory)... delete it before invoking makedirs
  * if the invocation fails... was doing bad things in a few cases. improved.

### DIFF
--- a/sarracenia/flow/__init__.py
+++ b/sarracenia/flow/__init__.py
@@ -1654,10 +1654,13 @@ class Flow:
 
         if not os.path.isdir(msg['new_dir']):
             try:
+                if os.path.exists(msg['new_dir']):
+                    os.unlink(msg['new_dir'])
                 os.makedirs(msg['new_dir'], self.o.permDirDefault, True)
             except Exception as ex:
                 logger.warning("making %s: %s" % (msg['new_dir'], ex))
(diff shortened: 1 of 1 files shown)


reidsunderland:
I'm a bit worried about this one. Deleting a file that already exists and replacing it with a directory seems like kind of surprising behavior, at least to me.

It's probably a pretty rare situation, so maybe it's not a big deal. 

I think there should at least be a warning logged when we delete the file, because that could help someone troubleshoot if they run into this.

petersilva:
The way the mirroring works, fine tuned order of operations is not assured. for All data types, not just files, we have a lot of code that deletes old versions of the files.... This is going a step further.  I'm not super certain about it either, .. if you want can revert that (requires separate patch... since the changes are all in one commit.)

Want me to remove that file with the same name remove code?

petersilva:
oh... it will still solve the client problem, because the improved error handling is what was essential.

petersilva:
OK... I was testing the hpc-mirroring example, and the rm causes crashes and data loss when mirroring the linux kernel. I will remove that stuff.

reidsunderland (approved):

PR #1370, opened 2025-01-16
Issue1369 - Add option to send a file as an attachment.

andreleblanc11:
This adds the configurability to send a file as an attachment. 

I also added logic so that the code will guess if the file is an image or not. Images can't be embedded inside the emails content, so it should always be sent as an attachment.


petersilva (approved):
looks great!

PR #1380, opened 2025-02-11
S3 Tweaks - anonymous access, create buckets. docs

reidsunderland:
I was trying to poll a bucket without providing credentials, but couldn't get it to work. The examples given show that `--no-sign-request` is used with the CLI client, so I added support for the boto3 equivalent. 

Once I did that, the current code to check if the bucket exists was failing, so I changed it to use `head_bucket` instead. 

I also added an attempt to create a bucket, if it doesn't already exist, but I haven't tested it.

And I added documentation for the S3 related credential details.

(This is the data we're trying to poll: https://registry.opendata.aws/noaa-nexrad/)

### DIFF
--- a/docs/source/Reference/sr3_credentials.7.rst
+++ b/docs/source/Reference/sr3_credentials.7.rst
@@ -50,6 +50,11 @@ passwords and settings needed by components.  The format is one entry per line. 
 - **ftp://user8:%2fdot8@host:990  implicit_ftps**
 - **https://ladsweb.modaps.eosdis.nasa.gov/ bearer_token=89APCBF0-FEBE-11EA-A705-B0QR41911BF4**
 
+- **s3://bucket-name s3_anonymous**
+- **s3://access_key_id:secret_access_key@bucket-name**
+- **s3://access_key_id:secret_access_key@bucket-name s3_session_token=a_big_string**
+- **s3://access_key_id:secret_access_key@bucket-name s3_endpoint=https://my-endpoint.com/**
+
 
 In other configuration files or on the command line, the url simply lacks the
 password or key specification.  The url given in the other files is looked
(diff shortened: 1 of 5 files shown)


gcglinton:
Looks good to me!

It does seem that at least one S3 Transfer test is failing (among others), but not sure if it's because of this change though.

reidsunderland:
Oh yeah, thanks, it is because of the change. The test is testing connecting to a bucket that shouldn't exist, but the new code successfully creates it. 

I tried to see if I could change something with moto to make the bucket creation fail, but couldn't find anything so I changed the test to expect a True result.

gcglinton:
I could see cases where the user doesn't have rights to create the bucket though, so maybe defining bucket creation as a config option makes sense. Then you could fully test all that code.

petersilva (approved):
Thanks for adding the documentation... We needed that for the recent discussions.

reidsunderland:
The code will try to create the bucket if it doesn't exist, and if that fails, connect will return False. I'm not sure how to test that. I tried to create a bucket without providing credentials, but it's still successful in the unit test. When I tested it manually it does fail and return False.

gcglinton:
> The code will try to create the bucket if it doesn't exist, and if that fails, connect will return False. I'm not sure how to test that. I tried to create a bucket without providing credentials, but it's still successful in the unit test. When I tested it manually it does fail and return False.

You can probably patch the underlying `create_bucket` method so that it returns `False`, or throws an error, thereby replicating the real behaviour.

I did similar things in [a few places](https://github.com/search?q=repo%3AMetPX%2Fsarracenia+path%3A%2F%5Etests%5C%2Fsarracenia%5C%2F%2F+mocker.patch&type=code).

Docs for Mocker are [here](https://docs.python.org/3/library/unittest.mock.html).

PR #1399, opened 2025-03-13
Issue1382 2 add path_only, name_only, data_only.  Improved documentation.

### LINKED ISSUE #1382: `nodupe_basis path` doesn't look at the path when the checksums are different
Currently, when `nodupe_basis path` is selected as the basis of the duplicate suppression, it will get overridden when the checksum of a file is different

@reidsunderland and I found that this is because the default routing key for a `path` is set to a checksum.

https://github.com/MetPX/sarracenia/blob/a5fd5ed06ce7faf75dd2862011df500eed8ab722/sarracenia/flowcb/nodupe/__init__.py#L33-L42

The `path`  doesn't have its own class unlike the [name option](https://github.com/MetPX/sarracenia/blob/development/sarracenia/flowcb/nodupe/name.py). So by default it always has the checksum as the key.

***

We also talked about adding an option to override not checking the checksum. `nodupe_check_data` might be a good name? It would be nice to give the user the option.

So the current behaviour we have is

basis | look at checksum? | look at path? | look at name?
-- | -- | -- | --
path | yes | yes | (not directly, included in path)
data | yes | no | no
name | no | no | yes

Ideally we would want the following

basis | look at checksum? | look at path? | look at name?
-- | -- | -- | --
path | optional (no by default) | yes | (not directly, included in path)
data | yes | no | no
name | optional (no by default) | no | yes

petersilva:
close  #1382 

* Does what was discussed, including some unit tests (more than before) 
* add path_only, name_only, rename data to data_only (but accept data as synonym.)
* documentation updated and (hopefully) clarified to give more concise guidance in the reference man page, but more robust motivation in the Explanation *DuplicateSuppression.rst* guide.  Also added a more full discussion about the Identity field in that same document.
*  in unit_tests message formats corrected to they better exercise the nodupe algorithm. Before this work,
    messages had *sum* fields  instead of *integrity*) so the "key" stuff in the code was never exercised.

path_only is what the original issue was complaining about.  it was expecting *path* to mean *path_only*.

### DIFF
--- a/sarracenia/flowcb/nodupe/__init__.py
+++ b/sarracenia/flowcb/nodupe/__init__.py
@@ -28,24 +28,42 @@ class NoDupe(FlowCB):
 
 
     def deriveKey(self, msg) -> str:
+        """ The key is the first thing checked when doing duplicate suppression.
+            If the keys do not match, nodupe doesn't bother looking at the path/name/anything else.
+            i.e. if the keys for two messages don't match, they are not duplicates.
+                 if the keys *do* match, it doesn't necessarily mean that they are duplicates, nodupe
+                 then has to check if the path/name/something else matches (see _not_in_cache method)
+        """
 
         key=None
+
+        # 1st priority: use the key from nodupe_override in the msg
         if ('nodupe_override' in msg) and ('key' in msg['nodupe_override']):
             key = msg['nodupe_override']['key']
+
+        # 2nd: derive from fileOp if fileOp is link or is a non-remove directory op
         elif 'fileOp' in msg :
             if 'link' in msg['fileOp']:
                 key = msg['fileOp']['link']
             elif 'directory' in msg['fileOp']:
                 if 'remove' not in msg['fileOp']:
                     key = msg['relPath']
+
+        # 3rd: use identity (checksum) if available (cod = calculate on download, i.e. no checksum yet)
         elif ('identity' in msg) and not (msg['identity']['method'] in ['cod']):
             key = msg['identity']['method'] + ',' + msg['identity']['value'].replace('\n', '')
 
+        # 4th: use relPath and time (and size, if known)
+        #   This is usually the case for polls, where we don't have the file's 
+        #   checksum, but the size and modification time are known
         if not key:
+            # time comes from file modification time when available, or msg pubTime (required field in all msgs)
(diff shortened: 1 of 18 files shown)


andreleblanc11:
The code changes look good to me. I like what's added to the documentation, it makes it really clear how the duplicate suppression works now.

petersilva:
If you're happy... please approve the PR...

andreleblanc11 (approved):

reidsunderland:
I'm reviewing too, it looks awesome, thanks @petersilva !

So far I noticed a debug log line that I'll make a commit to remove

sarracenia/flowcb/nodupe/__init__.py
```diff
@@ -28,24 +28,42 @@ class NoDupe(FlowCB):
...
+            # time comes from file modification time when available, or msg pubTime (required field in all msgs)
             if 'mtime' in msg:
                 t = msg['mtime']
             else:
                 t = msg['pubTime']
+
+            # if file size is known, use it
             if 'size' in msg:
                 key = f"{msg['relPath']},{t},{msg['size']}"
             else:
```
reidsunderland, line 74: When I first started looking at this, @andreleblanc11 was concerned that when identity is not available in the message `name` wouldn't work properly if the key is derived from `relPath` here. I think he's correct.

Should we check if `m['nodupe_override']['path']` is set, and if it is, use it here instead of `msg['relPath']`? 

(My code block selection didn't quite work). This is the section I'm wondering about:

https://github.com/MetPX/sarracenia/blob/8f8e90d9f18fc71b615e988b6900a486daf2db2a/sarracenia/flowcb/nodupe/__init__.py#L67-L70

reidsunderland (commented):
I added one note

sarracenia/flowcb/nodupe/__init__.py
reidsunderland: This is what I think we might need?

```python3
 if 'nodupe_override' in msg and 'path' in msg['nodupe_override']:
     path = msg['nodupe_override']['path']
 else:
     path = msg['relPath']

 if 'size' in msg: 
     key = f"{path},{t},{msg['size']}" 
 else: 
     key = f"{path},{t}" 
```

PR #1400, opened 2025-03-14
Azure transfer tweaks

reidsunderland:
- Adds support for specifying an account name in the credentials. 
    - When only specifying a key, I was getting this error: `ValueError: Unable to determine account name for shared key credential.`
    - This was the critical thing that eventually got the connection working
- Fixes the feature detection for Azure
- Allows us to ignore certificate errors with `tlsRigour lax` (I was trying to use an internal domain instead of blob.core.windows.net)
- Changes how the URL and credentials work to support different URLs. 
    - Unfortunately, this means that we have to use `azure://your_storage_account.blob.core.windows.net/container_name` instead of just `azure://your_storage_account/container_name`
- Specified the user agent when creating the ContainerClient. If it's set there, it seems to be applied to all method calls


andreleblanc11 (approved):

gcglinton:
image

Jokes aside, this looks like a great set of fixes.

PR #1401, opened 2025-03-14
add plugin to filter based on contentType

reidsunderland:
For building WMO-00 format files, we need to be able to separate binary and alphanumeric bulletins.


petersilva (approved):
This looks great!

petersilva:
uh... I approved it... so last couple of days, if we have 1 approval (like when I worked) I would trigger the merge...
Do people want me to wait for more approvals? or should I still merge if anyone does.

I think it will auto-merge if everybody on the review list blesses it, but that almost never happens.

reidsunderland:
I think 1 approval is still okay to merge

PR #1404, opened 2025-03-17
doc update after going through pre-release process

reidsunderland:
I changed the formatting a bit and a few of the commands.

And since Ubuntu 20.04 won't be supported on GitHub soon, added steps to test on 20.04 manually


petersilva (approved):
Do we have any deployments on ubuntu 20?  Maybe we can skip it?
The changes look good though.

reidsunderland:
Thanks. Yeah, the DSRs are 20.04 :(

PR #1405, opened 2025-03-18
fix #1390 amend documentation to better describe mirror

### LINKED ISSUE #1390: mirror defaults to True despite documentation being misleading
The [mirror options' documentation](https://metpx.github.io/sarracenia/Reference/sr3_options.7.html#mirror-flag-default-off) states that it defaults to `off`.

We've discovered (the hard way) that this only seems to be true for a subscriber. 
This is further emphasized in the documentation in [this table](https://metpx.github.io/sarracenia/Explanation/Concepts.html#id2) where only the `subscriber` has `mirror False` in the `Key Option Settings` field.

The reason why this has been noticed recently is because v3.00.57 of sr3 has been added to operational servers. The change in https://github.com/MetPX/sarracenia/issues/1337 has broken some of our configurations.

We had sarra/sender configs that had `mirror False` set after/reject statements in a configuration. So, because the masks' now get applied correctly and mirror defaults to True in sarra/senders, this caused unintended behaviours.

i.e. 

```
...
directory /path/to/somewhere
accept .*something.*
mirror False                                      <- This used to get applied to all of the masks' in v3.00.56, but won't do anything in v3.00.57
...
```

## Options
- We could default mirror to `False` for all configs, but that would require an audit of all of our current sr3 configs and would likely be a lot of work.
- We can update the documentation to reflect that mirror defaults to True, but defaults to False for a subscriber.

petersilva:
close #1390 

mirror already works correctly... but the documentation is unclear.  Fix that in both languages.

### DIFF
--- a/docs/source/Reference/sr3_options.7.rst
+++ b/docs/source/Reference/sr3_options.7.rst
@@ -1226,7 +1226,7 @@ be rejected when consuming. Messages older than Max value are discarded
 by the subscriber. (0 means no maximum age) 
 
 
-mirror <flag> (default: off)
+mirror <flag> (default: on (except subscribe))
 ----------------------------
 
 The  **mirror**  option can be used to mirror the dd.weather.gc.ca tree of the files.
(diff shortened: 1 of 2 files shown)


andreleblanc11 (approved):
I added a missing line in the french docs.

petersilva:
> I added a missing line in the french docs.

oops! You needed to lengthen the underline also.  The underline length has to match.  that won't render properly.

andreleblanc11:
It looks fine to me https://metpx.github.io/sarracenia/fr/Reference/sr3_options.7.html#mirror-flag-defaut-on-sauf-subcribe

PR #1406, opened 2025-03-20
Add AddSMHeader for client adaptability

andreleblanc11:
We have a client that is requesting the option to disable adding the `AAXX ddhhmm4` string to the second line of SM/SI bulletins.

They've recently started sending us `SM` bulletins that have `BBXX` as the second line of the bulletin, which doesn't need to have the  `AAXX ddhhmm4` line added. 

In sundew, the option to remove the `AAXX ddhhmm4` string was available, however I didn't port it to sr3, because all of our operational clusters had `AddSMHeader True`, so I assumed the option wasn't needed anymore.

I added the changes in the unit test. The results pass so the option seems to work.


reidsunderland (approved):
I think the changes are good the way they are. I changed `bool` to `flag` in the add_option call. It either has to be `'flag'` or `bool` (without quotes). https://github.com/MetPX/sarracenia/blob/3662558f17f552cfbd1a51492702d4c44c7d7015/sarracenia/config/__init__.py#L1155-L1245

PR #1414, opened 2025-04-01
Fix Azure's renaming not working when `directory` isn't `/`

andreleblanc11:
With the patch, a transfer now succeeds when `inflight .tmp` is set in the configuration and `directory` isn't set to `/`.

```
2025-04-01 14:10:40,410 [DEBUG] sarracenia.flow send azure_transport send /NUMERIC/2025/04/01/14 202503281912~CAPPI_1.0KM_NUMERIC_AS_META~CAPPI,1.0,AGL,MPRATE:URP:CASLA_NUMERIC_AS_META:RADAR:NUMERIC
2025-04-01 14:10:40,410 [DEBUG] sarracenia.transfer.azure check_is_connected sr_azure check_is_connected
2025-04-01 14:10:40,410 [DEBUG] sarracenia.flow send azure_transport send cd to /NUMERIC/2025/04/01/14
2025-04-01 14:10:40,410 [DEBUG] sarracenia.transfer.azure cd_forced forcing into  /NUMERIC/2025/04/01/14
2025-04-01 14:10:40,410 [DEBUG] sarracenia.transfer.azure cd changing into /NUMERIC/2025/04/01/14
2025-04-01 14:10:40,410 [DEBUG] sarracenia.flow send hasattr=False, thresh=0, len=356066, remote_off=0, local_off=0
2025-04-01 14:10:40,411 [DEBUG] sarracenia.transfer.azure put 202503281912~CAPPI_1.0KM_NUMERIC_AS_META~CAPPI,1.0,AGL,MPRATE:URP:CASLA_NUMERIC_AS_META:RADAR:NUMERIC to https://the-destination/eccc/NUMERIC/2025/04/01/14/202503281912~CAPPI_1.0KM_NUMERIC_AS_META~CAPPI,1.0,AGL,MPRATE:URP:CASLA_NUMERIC_AS_META:RADAR:NUMERIC.tmp
[...]

PR #1416, opened 2025-04-03
Email sender bug when file type is None + logging improvements

andreleblanc11:
Couple of things in this PR.

1. I found a bug with the file type check logic in the email sender. It would fail when the file type is `None`. 

This was the error message encountered.
```
2025-04-03 17:58:23,802 [DEBUG] sarracenia.flowcb.send.email send Exception details:
  File "/local/home/sarra/sr3/sarracenia/flowcb/send/email.py", line 155, in send
    if self.o.email_attachment or 'image' in file_type[0]:
TypeError: argument of type 'NoneType' is not iterable
```

2. I ran into a case where I was confused when a config had an invalid credential set for `broker` and `post_broker` wasn't set. It wasn't really obvious what was the problem. When this happens, both `broker` and `post_broker` are set to `None`. There weren't any logs that seemed to give a clue to this, so I added one.
3. In regards to the recommendations in https://github.com/MetPX/sarracenia/pull/1414, I added a warning when `inflight` is used with the Azure transfer class.

### DIFF
--- a/sarracenia/flowcb/send/email.py
+++ b/sarracenia/flowcb/send/email.py
@@ -147,12 +147,13 @@ class Email(FlowCB):
         # Get list of recipients for this message, from the mask that matched the filename/path
         recipients = self.o.masks[msg['_mask_index']][-1]
 
+        # i.e. (type/subtype, encoding)
         file_type = mimetypes.guess_type(ipath)
 
         # Prepare the email message
         try:
             # Build a non-text email message for the attachment if specified or if the file type can be deemed to be an image.
-            if self.o.email_attachment or 'image' in file_type[0]:
+            if self.o.email_attachment or (len(file_type) > 0 and file_type[0] and 'image' in file_type[0]):
                 emsg = MIMEMultipart()
                 emsg_text = MIMEText(f"{self.o.email_attachment_text}")
                 # Add the attachment text that will be paired with the attachment data
--- a/sarracenia/config/__init__.py
+++ b/sarracenia/config/__init__.py
@@ -2169,6 +2169,9 @@ class Config:
             logger.critical( f"{component}/{config} vip feature requested, but missing library: {' '.join(features['vip']['modules_needed'])} " )
             sys.exit(1)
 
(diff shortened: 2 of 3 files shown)


sarracenia/flowcb/send/email.py
```diff
@@ -147,21 +147,23 @@ def send(self, msg):
...
+                if self.o.email_attachment or 'image' in file_type[0]:
+                    emsg = MIMEMultipart()
+                    emsg_text = MIMEText(f"{self.o.email_attachment_text}")
+                    # Add the attachment text that will be paired with the attachment data
+                    emsg.attach(emsg_text)
+                    with open(ipath, 'rb') as fp:
+                        attachment_data = fp.read()
+                    attachment = MIMEApplication(attachment_data, name=os.path.basename(ipath))
+                    # Add the attachment data to the email
+                    emsg.attach(attachment)
```
reidsunderland: i think there might be a missed case here. Before the else would get executed when self.o.email_attachment was false and image was not in file_type[0]. 

But now, if you enter into the `if file_type[0]`, there's no else for the if inside?

I think `if self.o.email_attachment or (len(file_type) > 0 and file_type[0] and 'image' in file_type[0]): ` might be better? 

Or I might be misunderstanding

andreleblanc11: Nice catch. I tested the change by sending txt, None and image type files and they all came out OK at the reception...

The else block below should've been indented.. I thought it would've failed with the txt file with my changes.

I'll add your recommendation. That seems to work well. I didn't know the order of execution in a `if` statement was independent between operators and sequential.

reidsunderland: Yes, the boolean operations in python "short-circuit": https://docs.python.org/3/library/stdtypes.html#boolean-operations-and-or-not

andreleblanc11: I've reverted the last commit and force pushed it with the changes.

reidsunderland (approved):

PR #1424, opened 2025-04-17
issue #1419 remove overwrite setting (now always on.)

### LINKED ISSUE #1419: overwrite/"copy files to one node" behaviour unexpected
On a cluster with multiple nodes, I want to have one node that contains all the files *and post them to a different exchange*.

a config like this:

```
vip 123

broker     amqp://feeder@shared_broker/
exchange   xpublic
queueName  q_${BROKER_USER}.${PROGRAM}.${CONFIG}.dev

subtopic *.MY-SOURCE.SOME.FILES.#

post_broker    amqp://feeder@shared_broker/
post_exchange  xs_singleton
post_baseUrl   http://${HOSTNAME}/
post_baseDir   /apps/sarra/public_data

acceptUnmatched False
mirror          True

# tried with and without overwrite true, doesn't seem to make a difference
overwrite True

directory ${PBD}
accept .*MY-SOURCE.SOME.FILES.*something.*
```

What I expect to happen:

1. Files not on the node with the vip are downloaded and posted (this works)
2. Files already on the node with the vip are not re-downloaded, but they should be posted (they're not posted)

I think it might be related to this code. In the second `if` statement, I think the messages should be added to `worklist.ok` to cause them to be posted, even if they don't need to be downloaded. 

https://github.com/MetPX/sarracenia/blob/9c44166eacadfc04423d76162929a831b9aa60d7/sarracenia/flow/__init__.py#L1953-L1961

I need to do some testing to confirm, but I'm also worried about the consequences of changing this behaviour.

petersilva:
close #1419 

replace overwrite=no with nodupe_basis name_only
sr3 convert does the replacement but v3 configs will need manual modification.

doesn't fix the problem but makes progress.

UPDATE 2025/05/15: now that multiple publishers is merged, this is the last PR that would be needed to fully address the issue.  so added the *close* tag at the top.

### DIFF
--- a/docs/source/Explanation/CommandLineGuide.rst
+++ b/docs/source/Explanation/CommandLineGuide.rst
@@ -342,7 +342,6 @@ View all configuration settings (the result of all parsing... what the flow comp
      'msg_total_interval': '0',
      'fileAgeMax': 0,
      'nodupe_ttl': 0,
-     'overwrite': True,
      'permCopy': True,
      'permDefault': 0,
      'permDirDefault': 509,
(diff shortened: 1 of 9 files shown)


reidsunderland:
I'm not 100% sure that nodupe is a completely adequate replacement. Overwrite works across multiple instances on the same node but nodupe is per-instance. Maybe that's a weird edge case, I don't know if there are any situations where we would need it, but I think we need to take a look at our configs to see where overwrite is being used (especially on v2) and figure out why it's needed.

petersilva:
no rush that I know about... can wait.

petersilva:
pretty sure @reidsunderland 's objection is founded.   If using multiple instances and overwrite=no will not overwrite in cases where the original version was received by a different instance.

PR #1425, opened 2025-04-17
Misc. AM/bulletin renamer fixes - April 2025

andreleblanc11:
- Found an interesting bug while running AM flow tests. `batch` needs to be at a high value when the bulletin renamer is being called. The AM server doesn't have a limit of files to gather (either that or it's not known). I documented this problem in the AM servers' example.

- American bulletins (SRCN40/SXCN40/etc._KWAL) aren't causing problems anymore with the bulletin renamer. This will be huge when we migrate our sundew receivers that gather NWS data.

- Have true random ints (like sundew did) when generating the bulletin sequence number.

Flow tests for AM pass on both U18.04 and U22.04. AM flow test PR coming soon too.

### DIFF
--- a/sarracenia/bulletin.py
+++ b/sarracenia/bulletin.py
@@ -1,5 +1,6 @@
 import logging
 import time 
+import random
 import re
 from base64 import b64decode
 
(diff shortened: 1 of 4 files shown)


andreleblanc11:
The explanation as to why we likely can attribute no station for the american bulletin headers : https://github.com/MetPX/sarracenia/issues/980#issuecomment-2813045243

petersilva (approved):
It's weird to keep the name seq ... implying a sequential number when it's actually randomly generated.

PR #1426, opened 2025-04-19
fix post_exchange parsing, it had a weirdness.

petersilva:
was adding the suffix multiple times. dunno why.  Noticed this symptom running
the static flow test... see poll/sftp_f62.conf:

```

fractal% sr3 show poll/sftp_f62 | grep post_broker
2025-04-19 09:00:55,377 736592 [INFO] sarracenia.flow.poll __init__ Good! post_exchange: xs_tsource_poll and exchange: xs_tsource_poll match so multiple instances to share a poll.
2025-04-19 09:00:55,377 736592 [WARNING] sarracenia.flow.poll __init__ nodupe_ttl < fileAgeMax means some files could age out of the cache and be re-ingested ( see : https://github.com/MetPX/sarracenia/issues/904
2025-04-19 09:00:55,381 736592 [INFO] sarracenia.flowcb.scheduled update_appointments for 2025-04-19 13:00:55.381841+00:00: [] 
 '_Config__post_broker': 'amqp://tsource@localhost/',
 'post_broker': 'amqp://tsource@localhost/',
fractal% sr3 show poll/sftp_f62 | grep post_exchange
2025-04-19 09:01:03,444 736598 [INFO] sarracenia.flow.poll __init__ Good! post_exchange: xs_tsource_poll and exchange: xs_tsource_poll match so multiple instances to share a poll.
2025-04-19 09:01:03,444 736598 [WARNING] sarracenia.flow.poll __init__ nodupe_ttl < fileAgeMax means some files could age out of the cache and be re-ingested ( see : https://github.com/MetPX/sarracenia/issues/904
2025-04-19 09:01:03,449 736598 [INFO] sarracenia.flowcb.scheduled update_appointments for 2025-04-19 13:01:03.449223+00:00: [] 
 'post_exchange': ['xs_tsource_poll', '_', 'p', 'o', 'l', 'l'],
 'post_exchangeSuffix': 'poll',
 'post_exchanges': [],
fractal% 

```

fixed by this patch.

PR #1418, opened 2025-04-08
Issue339 try6 implement multi-broker subscriptions

petersilva:
close #339

Multiple subscriptions from a single broker work now.

This builds on the subscriptions that were built over the last few months.
The flowcb/gather/message plugin implementes one moth instance per subscription.

there is the *subscribe/dual_amis.conf* example added to the included ones in the distribution:

(available by: *sr3 add subscribe/dual_amis.conf* )
```
broker amqps://dd.weather.gc.ca/
topicPrefix v02.post

logReject on

# instances: number of downloading processes to run at once.  defaults to 1. Not enough for this case
instances 5

# expire, in operational use, should be longer than longest expected interruption
expire 10m

# new topic... in 2025
subtopic *.WXO-DD.bulletins.alphanumeric.#

broker amqps://hpfx.collab.science.gc.ca/

subtopic *.WXO-DD.bulletins.alphanumeric.#

#just dump into the current directory.
mirror false
directory /tmp/hpfx_amis/

```

can see that messages about the same item are received from both brokers.
The second one received is rejected as a duplicate, while the first one received
is acted on as a download.  Log extract below:

```
 ...
2025-04-07 23:37:52,034 [INFO] 3048061 sarracenia.flowcb.log after_accept accepted: (lag: 155.76 )  exchange: xpublic subtopic: 20250408.WXO-DD.bulletins.alphanumeric.20250408.SR.KWAL.03 a file with baseUrl: https://dd2.weather.gc.ca/ relPath: 20250408/WXO -DD/bulletins/alphanumeric/20250408/SR/KWAL/03/SRCN40_KWAL_080334___39084 sundew_extension: cvt_nws_bulletins-sr:KWAL:SR:3:Direct:20250408033514 id: r9wR/s5 size: 180 
 ...

025-04-07 23:37:52,046 [INFO] 3048061 sarracenia.flowcb.log after_accept accepted: (lag: 155.76 )  exchange: xpublic subtopic: 20250408.WXO-DD.bulletins.alphanumeric.20250408.SR.KWAL.03 a file with baseUrl: https://hpfx.collab.science.gc.ca relPath: /20250408/WXO-DD/bulletins/alphanumeric/20250408/SR/KWAL/03/SRCN40_KWAL_080334___39084 sundew_extension: cvt_nws_bulletins-sr:KWAL:SR:3:Direct:20250408033514 id: r9wR/s5 size: 180 
...
[...]

PR #1427, opened 2025-04-22
Issue339 part7 multi publish

petersilva:
Make *post_broker* be a setting you can specify multiple times to be able to publish to multiple destinations.  

This is a continuation of work on #339. to add support for multiple post_brokers so after download, one goes to multiple brokers and/or exchanges. Also addressing the last bit of #1419 ... this comment: https://github.com/MetPX/sarracenia/issues/1419#issuecomment-2801776030

So this is kind of part 3  as a  completing the work of #1418, and #1419

This PR includes all the patches in #1418, and #1426.   Please merge those first,..
Then this PR will be only a few commits long.


petersilva:
please note, that this work does not include #1424 ... "removing the overwrite option"... that is still being evaluated, and this code is completely independent of those changes.  It's just that that  change and this change together implement everything needed to address #1419.

petersilva:
closing in favour of #1528 which is much easier to read.  After other stuff got merged, the commit count in this PR just kep on rising... but only a fraction are unique to this PR.  created a new branch by cherry picking the really unique stuff.

PR #1428, opened 2025-04-22
Issue339 part8 multiple publishers

petersilva:
replacing #1427 since subscribers are now merged, can make much smaller & cleaner PR.

This is a continuation of work on #339. to add support for multiple post_brokers so after download, one goes to multiple brokers and/or exchanges. Also addressing the last bit of #1419 ... this comment: #1419 (comment)

So this is kind of part 3 as a completing the work of #1418, and #1419


andreleblanc11:
I tried testing the multi-publisher logic, and it does look like it was able to publish to 2 brokers simultaneously (dd.weather and hpfx.collab). 

I just tried posting a couple of files from my test machine. Here is the config that I used for the test.

```
amqp_consumer True

broker amqp://tfeed@localhost/
exchange xs_tfeed
topicPrefix v02.post

# instances: number of downloading processes to run at once.  defaults to 1. Not enough for this case
instances 1

# expire, in operational use, should be longer than longest expected interruption
expire 10m
prefetch 25

# new topic... in 2025
queueName q_anonymous.${CONFIG}.${HOSTNAME}.ANL-TEST
subtopic #

#just dump into the current directory.
mirror True
directory /net/local/home/leblanca/tmp/hpfx_amis/
accept .*
acceptUnmatched False

post_broker amqps://feeder@dd.weather.gc.ca/
post_exchange xpublic2
post_baseDir  /net/local/home/leblanca/
post_baseUrl  ${HOSTNAME}


post_broker amqps://feeder@hpfx.collab.science.gc.ca/
post_exchange xpublic2
post_baseDir  /net/local/home/leblanca/
post_baseUrl  ${HOSTNAME}

debug True
logReject on
```

I noticed that one `publisher` entry didn't have the `routing_key` attribute in an `sr3 show`. I'm not sure if that's expected or not.

```
sr3 show sarra/issue339 | grep -A20 publisher
 'publishers': [{'auto_delete': False,
                 'broker': 'amqps://feeder@dd.weather.gc.ca/',
                 'durable': True,
                 'exchange': ['xpublic2'],
                 'exchangeDeclare': True,
                 'messageAgeMax': 0,
                 'messageDebugDump': False,
                 'persistent': True,
                 'timeout': 300},
                {'auto_delete': False,
                 'broker': 'amqps://feeder@hpfx.collab.science.gc.ca/',
                 'durable': True,
                 'exchange': ['xpublic2'],
                 'exchangeDeclare': True,
                 'messageAgeMax': 0,
                 'messageDebugDump': False,
                 'persistent': True,
                 'timeout': 300,
                 'topicPrefix': ['v02', 'post']}],
```

The logging output also doesn't look right. I think it'd be better to have an `after_post` entry for each broker that successfully got posted to.

```
2025-04-22 19:59:16,132 [INFO] sarracenia.flowcb.log after_post posted to amqps://feeder@dd.weather.gc.ca/,xpublic2,v02.post.tmp.hpfx_amis.net.local.home.leblanca.am_bulletins.20250327.AA.CYCX.23 amqps://feeder@hpfx.collab.science.gc.ca/,xpublic2,v02.post.tmp.hpfx_amis.net.local.home.leblanca.am_bulletins.20250327.AA.CYCX.23 a file with baseUrl: https://my-test-serve relPath: tmp/hpfx_amis/net/local/home/leblanca/am_bulletins/20250327/AA/CYCX/23/AACN98_CYCX_272308___00270 size: 35 id: t6xy9gR
```

I think it's also worth testing failure situations (if one broker is down, and one is up, what happens).

petersilva:
multi-publish failures are not fully tested, but the codes is supposed to work as follows:
the after_post log message prints each broker it succeeded in posting to.  if one post failed, isn't listed.
the ones that failed end up in the retry queue, and when retried, it should only try to post to the missing ones, and if it succeeds, it will say it posted to that one.   

so... yeah... it will only pass to after_post after the last publish has succeeded... until then it will stay in the retry list... hmm... getting it a separate message per  publish... can do that, but isn't how it is currently done.

petersilva:
about *routing_key* ... that is something set differently for every message.  The settings that contribute to a routing_key when publishing are:  

* post_topicPrefix (aka publisher[i]['topicPrefix'])
* post_exchange (aka publishers[i]['exchange'] 

the rest of the topic comes from the topic derived from the message content.

petersilva:
oh... I see what you mean... topicPrefix is missing in one of the publishers... I'll take a look.

petersilva:
So the discussion about one log publish message reporting on all post_brokers, vs. separate log messages for publishing to each broker is actually quite interesting.  What this PR currently does is arrange for the same message to be sent to multiple destinations, so the settings that are different per post broker are the ones relevant to sending the message:

* post_broker
* everything related to setting exchanges (aka post_exchange, post_exchangeSuffix, post_exchangeSplit)
* topic_prefix
* durable/timeouts etc...

but everything related to message content is currently shared. There are settings like:

* post_baseUrl
* post_baseDir
* post_documentRoot
* post_format

So, if you wanted to:

* send to an old broker with v02 messages, and a new broker with v03... nope. pick one.
* send to one exchange with an http url, and another with file urls ... nope. pick one.

With the current implementation, the same message is sent to all destinations. I don't think that is something that should be changed as part of this PR, but it can be further discussed. It would require some deep re-factoring... that would create per destination messages.   I think the scope of this PR should be restricted to one message being sent to multiple post brokers.

petersilva:
ok... so fixed the lack of topicPrefix in some publishers problem, with one commit, and make it obey post_format with a second one.  I'm now thinking about post_baseDir, post_documentRoot, and post_baseURl... hmm...
will take a day or two, but folks can play with what is there for now.

petersilva:
so... changing the post_baseUrl would be easy, since it is it's own field, but the problem is interactions with post_baseDir, post_documentRoot.

These substitutions result in setting *relPath* appropriately.  The problem is that different baseUrl's can result in different document roots, and therefore different relPath per post_broker.  So that field needs to be calculated separately per post_broker, and so message differences would potentially be:

* post_baseUrl
* relPath
* (sub)topic

so the problem here is running substitutions of the new_dir field vs. the documentRoot and baseUrl setings to build a new relPath.

Also... on receipt... currently using a single *baseDir* for all messages... should that be different for different subscriptions? similarly for *documentRoot*

The changes are going to be hard to straighten out.  I'm making a new branch for it.
Probably consider merging this PR as it currently is.  There are no known bugs in it, it just just has the behaviour that it republished the same message to all destinations.

Publishing per destination tuned messages is further work.

petersilva:
OK, so I merged the other branch back into this PR.  I think it now behaves like @andreleblanc11  was asking for. So now, after a download is done, one can specify two sets of publication parameters (broker, baseUrl, baseDir, relPath) by splitting each download into a separate message for each publisher. So now if a publication fails, it will get requeued, but the one that succeeds will be marked as such.  Now each publication will be logged (so each log message will be for one broker.)

Yeah, I think it completely works now.

petersilva:
fwiw... I just started work on a test where I have one mqtt consumer, and one amqp consumer...  in the same subscriber, and mixing the two causes problems. working on it.  All amqp, or all MQTT looks fine, but multi-consumer where the consumers aren't the same protocol is broken.

andreleblanc11:
I think I may have found a bug while running a cleanup with sr3_tools. I guess this is more of a bug for https://github.com/MetPX/sarracenia/pull/1418

```
sr3d --dangerWillRobinson=7 cleanup 'sender/my-confs*'

# On the first queue it tries to remove
 cleanup: deleting: sender/wxo-dev-be01-dd is: q_feeder.sender.wxo-dev-be01-dd.server.cmc.ec.gc.ca.test @ broker.cmc.ec.gc.ca

   File "/local/home/sarra/sr3/sarracenia/moth/amqp.py", line 353, in getSetup
     if not self.__connect(broker):
   File "/local/home/sarra/sr3/sarracenia/moth/amqp.py", line 245, in __connect
     self.connection.connect()
   File "/usr/lib/python3/dist-packages/amqp/connection.py", line 288, in connect
     self.drain_events(timeout=self.connect_timeout)
   File "/usr/lib/python3/dist-packages/amqp/connection.py", line 471, in drain_events
     while not self.blocking_read(timeout):
   File "/usr/lib/python3/dist-packages/amqp/connection.py", line 477, in blocking_read
     return self.on_inbound_frame(frame)
   File "/usr/lib/python3/dist-packages/amqp/method_framing.py", line 55, in on_frame
     callback(channel, method_sig, buf, None)
   File "/usr/lib/python3/dist-packages/amqp/connection.py", line 481, in on_inbound_method
     method_sig, payload, content,
   File "/usr/lib/python3/dist-packages/amqp/abstract_channel.py", line 128, in dispatch_method
     listener(*args)
   File "/usr/lib/python3/dist-packages/amqp/connection.py", line 603, in _on_close
     (class_id, method_id), ConnectionError)
 amqp.exceptions.AccessRefused: (0, 0): (403) ACCESS_REFUSED - Login was refused using authentication mechanism AMQPLAIN. For details see the broker logfile.

 During handling of the above exception, another exception occurred:

   File "/local/home/sarra/.local/bin/sr3", line 33, in <module>
     sys.exit(load_entry_point('metpx-sr3', 'console_scripts', 'sr3')())
   File "/local/home/sarra/sr3/sarracenia/sr.py", line 3220, in main
     gs.cleanup()
   File "/local/home/sarra/sr3/sarracenia/sr.py", line 1872, in cleanup
     qdc.getSetup()
   File "/local/home/sarra/sr3/sarracenia/moth/amqp.py", line 404, in getSetup
     f'connecting to: {queue["name"]}, durable: {queue["durable"]}, expire: {queue["expire"]}, auto_delete={queue["auto_delete"]}'
 KeyError: 'auto_delete'
```

andreleblanc11:
I just tried re-adding the configs I had just removed when I discovered the bug. 

I restarted them and retried to stop and perform a `sr3d --dangerWillRobinson=7 cleanup 'sender/my-confs*'` but wasn't able to recreate the error.. strange.

petersilva:
I tried reproducing your issue, and couldn't.  I think you maybe had one node that had an old revision when you did it initially (just a guess?) I edited subscriptions.json to remove the 'autodelete' from it... and it dealt with it fine, no error.
I fixed this other problem I found making mult_static tests, but afaict there is no known problem.

petersilva:
found another bobo in multi testing... hang on.

petersilva:
I got the new sr_insects tests to work (it uncovered the issues fixed by the last few commits) and afaict... Everything is good now...

petersilva:
except... I found some unit test fails... hang on...

petersilva:
except... I found some unit test fails... hang on...

petersilva:
If  you want to try out some tests, you can try the *multi_tests* branch of sr_insects.  there is a matching PR for that https://github.com/MetPX/sr_insects/pull/72

petersilva:
so... fwiw... all the unit tests should be fine now.  The thing I am looking at is trying to update python API examples... I'm trying to figure out if the old stuff works unchanged... but it doesn't look like it, so python API change... need to document it by updating the examples.  Still what is here could be merged.  It is complete and working.

On the other hand, I'm wondering if we should bump the version number a bit... maybe go from 3.0.xx  to 3.10.xx ?  It's pretty different with multi subscribers and publishers.

reidsunderland:
@andreleblanc11 and I were just talking about that yesterday! We agree, this should probably be v3.1

petersilva:
OK, glad you guys thought the same thing... while it's upwardly compatible (people don't *need* to change anything.) it feels like the "advice" we give will be quite different, and examples we provide should probably change over time.

Regarding specifics of the version numbering... with older versions, the second set of digits referred to the year, so I think we two digits is a better starting point to avoid ambiguity of 3.1 vs. 3.10   ... we could do 3.01, or  3.10  ... but I think 3.1 is ambiguous.  but whatever the idea of changing the second set of digits is the thing.

andreleblanc11:
I think we should keep our 2 digit system. So the next version would be `3.01.00` with that logic.

petersilva:
OK I finished updating examples and things... should be good to go now.

petersilva:
fwiw... I feel like 3.01 sounds *too small* for the change... I would prefer 3.10 ...  I was actually musing about 4... but I think that would spook people... so no... anyways... the 3.01 is fine too.

petersilva:
fwiw... no further patches likely for this branch.  It's done now.

reidsunderland (approved):

andreleblanc11:
I've found another problem. It seems to only affect configs that have `debug on` or `logLevel debug`

```
2025-05-12 18:37:15,827 [DEBUG] sarracenia.flow loadCallbacks complete
2025-05-12 18:37:15,827 [DEBUG] sarracenia.flow run working directory: 46505
2025-05-12 18:37:15,827 [DEBUG] sarracenia.flow run options:
  File "/local/home/sarra/sr3/sarracenia/instance.py", line 276, in <module>
    i.start()
  File "/local/home/sarra/sr3/sarracenia/instance.py", line 267, in start
    self.running_instance.run()
  File "/local/home/sarra/sr3/sarracenia/flow/__init__.py", line 554, in run
    self.o.dump()
  File "/local/home/sarra/sr3/sarracenia/config/__init__.py", line 1258, in dump
    c = copy.deepcopy(self.dictify())
  File "/usr/lib/python3.6/copy.py", line 150, in deepcopy
    y = copier(x, memo)
  File "/usr/lib/python3.6/copy.py", line 240, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
  File "/usr/lib/python3.6/copy.py", line 150, in deepcopy
    y = copier(x, memo)
  File "/usr/lib/python3.6/copy.py", line 215, in _deepcopy_list
    append(deepcopy(a, memo))
  File "/usr/lib/python3.6/copy.py", line 150, in deepcopy
    y = copier(x, memo)
  File "/usr/lib/python3.6/copy.py", line 220, in _deepcopy_tuple
    y = [deepcopy(a, memo) for a in x]
  File "/usr/lib/python3.6/copy.py", line 220, in <listcomp>
    y = [deepcopy(a, memo) for a in x]
  File "/usr/lib/python3.6/copy.py", line 161, in deepcopy
    y = copier(memo)
TypeError: cannot deepcopy this pattern object
```

We probably need to revert this change https://github.com/MetPX/sarracenia/commit/47ef926677204e58ff844737a7fae6e429d0ff33#diff-f3a0c25e66d07e41ce748b41d16e2507811b6e819e85edb90fdb69fd127654fcL1258?

petersilva:
Is that on 18.04/python 3.6 ?  It passes all flow tests on 24.04, and lots of those have logLevel debug in them.

andreleblanc11:
Yes, sorry I should've specified. I've added the merged code to both ddsr-dev systems. Both servers are hosting U18.04 and have python3.6 installed.

petersilva:
OK running it on 18.04 trying to reproduce.

petersilva:
uh... to avoid another PR, the fix is two lines... I just pushed directly to development... so git pull should fix it.
commit: bbaeb2ef31ab09af8be6295925dedadc5e26f8ad

PR #1429, opened 2025-04-24
Lots of wiski improvements

andreleblanc11:
- Give the option to customize return fields in the csv file
- Instead of specifying a parameter name, use a rejected list, as we want to accept most parameters.
- Improve code in error situations.


petersilva (approved):

petersilva:
It says draft... so is it ready for merging? I was ok with the existing patch...

andreleblanc11:
I'm going to start feeding the data to the client soon, and just want to wait in case he has some special requests for the data acquisition.

andreleblanc11:
The client seems happy. @petersilva requesting another review.

I just added another commit since your last review.

Issue #1411, opened 2025-03-29
Look into AMQP/RabbitMQ heartbeat


reidsunderland:
It seems like the AMQP library we are using defaults to setting the heartbeat value to 0.

This means that the RabbitMQ broker doesn't detect when a connection has been broken. The sarra client reconnects itself on a new connection, but the old one seems to get left behind on the server.

https://www.rabbitmq.com/docs/heartbeats

https://docs.celeryq.dev/projects/amqp/en/latest/#:~:text=Support%20for%20heartbeats,Connection.send_heartbeat.

![Image](https://github.com/user-attachments/assets/16897dd1-c280-4e27-8462-0c580e1ce273)

![Image](https://github.com/user-attachments/assets/14a2b0c2-af10-4fc7-a2f3-9d6be131ecc1)

![Image](https://github.com/user-attachments/assets/dd9f012e-5c9a-4937-835f-1907b4ba8e6d)

reidsunderland:
If we set heartbeat to a low value, it could cause problems during slow downloads. The py-amqp library documentation recommends calling `heartbeat_tick` "on the order of once per second" https://docs.celeryq.dev/projects/amqp/en/latest/reference/amqp.connection.html#:~:text=heartbeat_tick(rate,timeout

petersilva:
I think you already know this, but just clarifying... there are two settings:
```
heartbeat = None
Final heartbeat interval value (in float seconds) after negotiation

heartbeat_tick(rate=2)[source]
Send heartbeat packets if necessary.
```

I  dunno if we even need a setting.  I guess we could try to figure out a setting that works
for all cases and just hard code it.  otoh, if a connection is idle it's going to be exchanging packets  relatively often, I don't know if that's a concern (a thousand connections means 1000 packets/sec.) seems like a DOS attack on the broker.  I wonder why it's so frequent... my gut tells me 10 or 100 seconds is a better choice... but often the recommendations come from experience.

reidsunderland:
That's a good point, and I didn't understand why they recommended calling `heartbeat_tick` every ~1 second. 

But I just re-read the documentation and looked at the code and I think it makes sense now. https://docs.celeryq.dev/projects/amqp/en/latest/_modules/amqp/connection.html#Connection.heartbeat_tick 

[...]

PR #1430, opened 2025-04-28
ack messages earlier for improved performance and reliability

petersilva:
Somebody had a complaint about the AMQP session dropping during long transfers... because it was idle while the transfers were going on.

One such discussion is #1411 ... where it talks about "a file transfer to take >60 seconds" where the real problem is that it will only ack any message in a batch after the entire batch has been attempted, which might take quite a while.

So with this change, each message is acknowledged individually ASAP, in download and send.

### DIFF
--- a/sarracenia/flow/__init__.py
+++ b/sarracenia/flow/__init__.py
@@ -486,7 +486,6 @@ class Flow:
         )
 
     def ack(self, mlist) -> None:
-        logger.critical("acking {m}" )
         if "ack" in self.plugins:
             for p in self.plugins["ack"]:
                 if self.o.logLevel.lower() == 'debug' :
--- a/sarracenia/flow/__init__.py
+++ b/sarracenia/flow/__init__.py
@@ -458,6 +458,8 @@ class Flow:
         """
         self.worklist.rejected.append(m)
         m.setReport(code, reason)
+        self.ack([m])
+        
 
     def stop_request(self) -> None:
         """ called by the signal handler to tell self and FlowCB classes to stop. Without this,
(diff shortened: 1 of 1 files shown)


petersilva:
specific mention: https://github.com/MetPX/sarracenia/issues/1411#issuecomment-2764334359

petersilva:
rephrasing the purpose of this PR: currently, all transfers in a batch must be processed (either completed, or retried *attempts* times) before any message is acknowledged.

so it you have batch 50... all 50 transfers have to happen before the first messages is acked.

With this PR, as soon as each transfer succeeds or fails, send the ack for that one message.  This is done at the same time the message is put on a non-incoming worklist.

Does anybody have any worries/failure modes about this?

reidsunderland:
Yes, I don't think we want to make this change.

This is my understanding from memory, it might not be correct:

Right now we receive messages from the broker, they go into unacked state on the broker, and we only ack them at the end of a batch once they've either been downloaded successfully or failed to download and have gone into the retry queue. 

Currently if we crash in the middle of a batch, the unacked messages will go back into ready state in the queue and will get re-delivered once the instance comes back (or delivered to another instance). We might process a message twice, but shouldn't lose anything. Essentially we have two copies of the message: one in the worklist (in memory) and one on the broker. We only ack and lose the broker copy once we know that the download was successful or the message has been stored to disk.

<br>

I think by changing to ack in the middle of a batch, we risk losing messages. If we crash in the middle of a batch, after some messages are acked, they'll be lost because the only place they exist is the in-memory worklist.

---

For the broker heartbeat thing, right now it's not an issue because we have heartbeats disabled, so the broker will never terminate our connection. If/when we want to turn it on, we can call `send_heartbeat()` in the middle of a batch to tell the broker we're still alive.

petersilva:
fwiw:
Even if we only ack after the complete batch, nothing has been persisted at that point. if we want to eliminate the possibility of message loss, then we would need to postpone the ack until message processing is complete (aka republish is done.)   The current processing has the same weaknesses.

PR #1433, opened 2025-05-02
Bulletin renaming miscellaneous fixes - 2025/05

andreleblanc11:
Things that are fixed

- The weird `batch` problem identified in https://github.com/MetPX/sr_insects/pull/70 is triggered because the `after_gather` call location is likely wrong.
   - We now call `after_gather` plugins before the return for the batch condition.

- I think WMO bulletins will hold carriage returns at the end of the bulletin header. Current code doesn't take into account this and renames the file wrong (with the carriage return).

- DMS sometimes sends bulletins with only the header, and that makes the rename plugin crash. The bulletin is technically wrong, but the code should handle such errors better.

- Change `getSequence` to a more reflective method name.

- Some stations coming from NWS have 6 or 7 characters and seem to be correct (at least comparing to Sundew behaviour). 
   - We also don't want to fail when the station REGEX fails. Assign no station to the filename instead.

Part of the yielded results came from testing with the full suite of bulletins coming from NWS. It's been over a day now and there are much less errors then initially. I'm still running the test to try and catch more minor errors.

### DIFF
--- a/sarracenia/bulletin.py
+++ b/sarracenia/bulletin.py
@@ -247,7 +247,7 @@ class Bulletin:
                 station = ''
 
             # Added to SR3
-            # The station needs to be alphanumeric, between 3 and 6 characters. If not, don't assign a station
+            # The station needs to be alphanumeric, between 3 and 7 characters. If not, don't assign a station
             if re.search('^[a-zA-Z0-9]{3,7}$', station) == None:
                 station = ''
 
--- a/sarracenia/bulletin.py
+++ b/sarracenia/bulletin.py
@@ -248,8 +248,8 @@ class Bulletin:
 
             # Added to SR3
             # The station needs to be alphanumeric, between 3 and 6 characters. If not, don't assign a station
-            if re.search('^[a-zA-Z0-9]{3,6}$', station) == None:
-                station = None
+            if re.search('^[a-zA-Z0-9]{3,7}$', station) == None:
+                station = ''
 
         return station
 
--- a/sarracenia/bulletin.py
+++ b/sarracenia/bulletin.py
@@ -156,8 +156,8 @@ class Bulletin:
             logger.error(f"Could not fetch file data of from either message content or {path}. Error details: {e}")
             return None
 
-    def getSequence(self):
-        """ sequence number to make the file unique...
+    def getRandom(self):
+        """ Generate Random number to make the file unique...
         """
         return str(random.randint(0, 99999)).zfill(5)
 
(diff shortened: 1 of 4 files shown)


reidsunderland (approved):

andreleblanc11:
There might be more commits coming. I'm still testing on my test environment.

andreleblanc11:
I think this PR is ready.

Unfortunately, there is still a problem with the renamer when raw SA/SP* files come in from the DMS. 

There is missing information in the bulletin contents. We need to pass those files into here
https://github.com/MetPX/sarracenia/blob/4cc19201caf6a4c464f0070d19070c80d1433e44/sarracenia/flowcb/gather/am.py#L362-L394

This will modify the contents in the bulletins, so calling the method shouldn't be added in the renamer logic IMO. The renamer should only _rename the filename,_ and _shouldn't touch the file contents_

The best way to do this is probably through a custom plugin that calls the `correctContents` method.

petersilva:
One of the mottos we used to have for Sundew development was: 

## We don't want to know...

https://github.com/MetPX/Sundew/blob/main/doc/dev/DevGuide.rst#principle-try-not-to-decode-the-data-aka-we-don-t-want-to-know

For what it's worth... A lot of the logic in bulletin parsing in Sundew was just bug-for-bug compatibility with the Tandem system it was replacing.  This made old systems that worked with the Tandem work with Sundew as a replacement.  We were paying six figures in maintenance costs per annum for the Tandem and so we needed the old system removed asap.  So we simplified the transition as much as possible.

But this method of getting to a quick Tandem replacement ended up also locking in bugs.   Sundew corrupts some data with it's transformations.  That's why we tried NOT to do all the transformations that Sundew did in Sarracenia... especially in WMO (vs. MSC) bulletins.  All those character substitutions cause message corruptions, they are just a snapshot of what was needed 30 years ago to get something to work.)  We were hoping to leave that behind in Sundew... porting them all forward... is probably unhelpful.

It is NOT correct  to run stuff we get from NWS through the old bulletin formatting traditionally used to get MSC bulletins, and then make WMO bulletins from that.  Rather, the WMO bulletins should be accepted as-is, untouched.  and only modify the MSC bulletins.  Encourage clients to migrate to WMO bulletins... giving them time to deal with the strange obsolete breakage on their own schedule.

Is any of this worthwhile? dunno.  the met people are working on JSON and other formats, the aviation people are deep into XML.  Will these old WMO386 formats be relevant for much longer? They will probably hang around for decades, but just annoying people with every change as people have to touch long obsolete systems.  In that light making stuff just work might be the right idea.  otoh... last time I checked (many years ago) normal MSC processing was corrupting binary (BUFR and GRIB) stuff and some other specialized WMO386 encodings as well... no obvious answers.

andreleblanc11:
Yeah, the answer also doesn't seem obvious to me.

The main goal of this PR was to move forward with the retirement of sundew. I didn't try to take into account the relevancy of the format of the bulletins we disseminate.

The DMS should probably be handling this bulletin 'transformation' I've been referencing.  
I think part of the problem is that we're still relying on a lot of the bulletins to be transferred to the circuit (AMCIR).

The system is pretty finnicky and those workarounds that we've been doing need to stay in place partially because of this.

petersilva:
so @reidsunderland  has blessed this... so do you want it merged?

andreleblanc11:
Yes

PR #1440, opened 2025-05-16
Issue1436 random ebo maximum

### LINKED ISSUE #1436: Exponential Back Off for broker reconnects
https://github.com/MetPX/sarracenia/blob/6d9f938429d3f0d73c72f186315543cc60fb4004/sarracenia/moth/__init__.py#L421-L435

When the broker for some flow gets disconnected (Monthly patch reboots? network issues? ...etc) the EBO function will continuously double until [it hits 600 seconds](https://github.com/MetPX/sarracenia/blob/development/sarracenia/moth/__init__.py#L434C48-L434C51). 10 minutes is a long time for SPC folks and it would be nice to be able to configure their individual subscribers to be more aggressive in reconnecting to the regional brokers.

petersilva:
close #1436

Proposed resolution for #1436 ... The backoff logic:

* backs off too quickly to maximum interval (10 minutes.)
* the maximum interval is too long.

corrections in this branch:
* it now waits at least 1 second before trying to connect again.
* it increases the interval at a rate of 1.2**number_of_attempts (base use to be 2) so the interval increases more slowly.
* the reconnection maximum interval is a value decided at start up as a random time between 1 and 4 minutes.

so if the broker goes down, and the connection attempt fails immediately, it should wait for:

* 1.2, 1.44, 1.73, 2.07, 2.49, 2.99, 3.58, 4.8, 5.16, 6.19, 7.43, 8.92, 10.7, 12.84, 15.41, 18.49 .... it will eventually get to a few minutes...

### DIFF
--- a/sarracenia/moth/__init__.py
+++ b/sarracenia/moth/__init__.py
@@ -464,8 +464,8 @@ class Moth():
         attempt_duration = max(now - start,1)
         self.next_connect_failures += 1
 
-        # wait twice as long after each failure. 
-        ebo = 2**self.next_connect_failures
+        # wait a little longer after each failure. 
+        ebo = 1.2**self.next_connect_failures
 
         # eboIntervalMaximum is something random between 1 and 4 minutes.
         # it is the ceiling. Otherwise based on the number of failures to connect and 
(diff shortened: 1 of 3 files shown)


petersilva:
closing in favour of #1445

PR #1442, opened 2025-05-20
Geometry plugin behaviour changed for no clear reason.

petersilva:
the unit tests were failing...  root causes:

* config.subtopic_seen  had to exist... (new thing) ... added hasattr() tests to make it optional.
* geometry test behaviour changed, where documented bahaviour says a polygon doesn't intercept a given one.
   afaict, it IS close enough to be considered "in."  I just changed what the test expects.  I don't understand why this behaviour changed.

### DIFF
--- a/sarracenia/config/__init__.py
+++ b/sarracenia/config/__init__.py
@@ -1652,13 +1652,13 @@ class Config:
                 self.logEvents = self.logEvents | set(['nodupe'])
 
             if k in ['statehost' ]:
-                if self.subtopic_seen:
+                if hasattr(self,'subtopic_seen') and self.subtopic_seen:
                     logger.error( f"{','.join(self.files)}:{lineno} {k} statehost needs to be before subtopic." )
                 elif lineno > 5:
                     logger.warning( f"{','.join(self.files)}:{lineno} {k} needs to be near the start of the file." )
             return
 
-        if k in queue_options and self.subtopic_seen:
+        if k in queue_options and hasattr(self,'subtopic_seen') and self.subtopic_seen:
             logger.warning( f"{','.join(self.files)}:{lineno} {k} needs to appear before *subtopic*" \
                 " unless you need different queues to have different settings")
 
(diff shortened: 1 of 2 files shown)


gcglinton:
Hm... according to my testing, Poly3 *isn't* inside either Poly1 or Poly2, so it should absolutely be rejected, and on my local environment (against `development` branch), it passes as is.

*edit: I took some time to plot out all the features in [JSFiddle](https://jsfiddle.net/8dt4n365/1/), and Poly3 is in fact not inside (or even touching) anything else.*

Just changing the counts makes the tests pass, but... it's not actually the proper behaviour IMO.

gcglinton:
Looks like there was a change in the way the `intersect` method of turfpy works.

In v0.0.7 it would return `None` if there was no intersection, but in v0.0.8, it retuns a GeoJSON Feature, with no coordinates.

My local dev environment is using v0.0.7, but your tests are using v0.0.8.

You can either:
 - pin 0.0.7 of turfpy (`turfpy==0.0.7` in `requirements.txt`)
 - update the callback code to account for the behaviour of `intersect` in both versions of `turfpy`
   - v0.0.7 returns `None` if there's no intersection, or a dict if there is
   - v0.0.8 always returns a dict, but the list at `$RETURNVALUE['geometry']['coordinates'][0]` will have length 0 if there's no collision, or a list of coordinates if there is


v0.0.7:
```
Intersect Poly1 <> Poly2:  {"geometry": {"coordinates": [[[-75.332, 39.284], [-75.833, 39.284], [-75.833, 40.0], [-75.332, 40.0], [-75.332, 39.284]]], "type": "Polygon"}, "properties": {}, "type": "Feature"}
Intersect Poly1 <> Poly3:  None
```

v0.0.8:
```
Intersect Poly1 <> Poly2:  {"geometry": {"coordinates": [[[-75.332, 39.284], [-75.332, 40.0], [-75.833, 40.0], [-75.833, 39.284], [-75.332, 39.284]]], "type": "Polygon"}, "properties": {}, "type": "Feature"}
Intersect Poly1 <> Poly3:  {"geometry": {"coordinates": [[]], "type": "Polygon"}, "properties": {}, "type": "Feature"}
   
```

petersilva:
that explains it much better. this pr is obviously bogus.

PR #1443, opened 2025-05-21
fix geometry for 0.0.8

petersilva:
proper fix because what intersect returns changed with version turfpy 0.0.8  thanks to @gcglinton 

replaces #1442

### DIFF
--- a/sarracenia/config/__init__.py
+++ b/sarracenia/config/__init__.py
@@ -1652,13 +1652,13 @@ class Config:
                 self.logEvents = self.logEvents | set(['nodupe'])
 
             if k in ['statehost' ]:
-                if self.subtopic_seen:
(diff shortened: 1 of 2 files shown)


gcglinton:
I don't think I'd have done it this way (version check to account for the different behaviour), but instead baked it into the code, comparing `intersect`'s return against both possible values (and documenting both checks in a comment).

This accomplishes the same thing, a bit more explicitly even, and obviously works though so maybe it's the better solution.

gcglinton (approved):

petersilva:
If you don´t check the version, then you leave a  puzzle for future devs:  Why compare to None?  They wonder why it is there, and it needs to stay there forever.

With the explicit check. the future people understand why it is bi-furcated.
It lets them eventually phase it out.   Otherwise, the code monotonically grows as work-arounds for versions accumulate.

PR #1445, opened 2025-05-21
Issue1436 try2 random ebo maximum

### LINKED ISSUE #1436: Exponential Back Off for broker reconnects

petersilva:
close #1436

replacing #1440 because other merges created conflicts. just picked the 
same commits and adjusted to resolve them.

otherwise it is the same as #1440 

copied from there:

Proposed resolution for #1436 ... The backoff logic:

backs off too quickly to maximum interval (10 minutes.)
the maximum interval is too long.
corrections in this branch:

it now waits at least 1 second before trying to connect again.
it increases the interval at a rate of 1.2**number_of_attempts (base use to be 2) so the interval increases more slowly.
the reconnection maximum interval is a value decided at start up as a random time between 1 and 4 minutes.
so if the broker goes down, and the connection attempt fails immediately, it should wait for:

1.2, 1.44, 1.73, 2.07, 2.49, 2.99, 3.58, 4.8, 5.16, 6.19, 7.43, 8.92, 10.7, 12.84, 15.41, 18.49 .... it will eventually get to a few minutes...

### DIFF
--- a/sarracenia/moth/__init__.py
+++ b/sarracenia/moth/__init__.py
@@ -49,6 +49,10 @@ def default_options() -> dict:
             o = copy.deepcopy(__default_options)
 
         return o
+          
+import random
+
+eboIntervalMaximum = 60 + random.random()*180
 
 def ProtocolPresent(p) -> bool:
     if ( p[0:4] in ['amqp'] ) and sarracenia.features['amqp']['present']:
@@ -345,6 +349,7 @@ class Moth():
 
         logging.basicConfig(format=self.o['logFormat'],
                             level=getattr(logging, self.o['logLevel'].upper()))
+        logger.debug( f" Maximum interval exponential back off of connecting to broker: {eboIntervalMaximum} " )
 
     def ack(self, message: sarracenia.Message ) -> bool:
         """
@@ -455,10 +460,18 @@ class Moth():
              it should eventually settle down to a long period though.
         """
         now=time.time()
-        attempt_duration = now - start
+        # if the attempt takes a long time, do not want to try again quickly.
+        # but if it fails immediately, then wait at least 1 second.
+        attempt_duration = max(now - start,1)
         self.next_connect_failures += 1
+
+        # wait twice as long after each failure. 
         ebo = 2**self.next_connect_failures
-        next_try = min(attempt_duration * ebo, 600)
+
+        # eboIntervalMaximum is something random between 1 and 4 minutes.
+        # it is the ceiling. Otherwise based on the number of failures to connect and 
+        # how long each attempt takes to fail.
+        next_try = min(max(attempt_duration * ebo,0.1), eboIntervalMaximum)
         self.next_connect_time = now + next_try
 
     def splitPick(self,message) -> int:
--- a/sarracenia/moth/__init__.py
+++ b/sarracenia/moth/__init__.py
@@ -52,7 +52,7 @@ def default_options() -> dict:
           
 import random
 
-eboIntervalMaximum = 60 + random.random()*180
+eboIntervalMaximum = 60 + random.random()*60
 
 def ProtocolPresent(p) -> bool:
     if ( p[0:4] in ['amqp'] ) and sarracenia.features['amqp']['present']:
--- a/sarracenia/moth/__init__.py
+++ b/sarracenia/moth/__init__.py
@@ -465,8 +465,8 @@ class Moth():
         attempt_duration = max(now - start,1)
         self.next_connect_failures += 1
 
-        # wait twice as long after each failure. 
-        ebo = 2**self.next_connect_failures
+        # wait a little longer after each failure. 
+        ebo = 1.2**self.next_connect_failures
 
         # eboIntervalMaximum is something random between 1 and 4 minutes.
         # it is the ceiling. Otherwise based on the number of failures to connect and 
--- a/sarracenia/moth/amqp.py
+++ b/sarracenia/moth/amqp.py
@@ -587,7 +587,7 @@ class AMQP(Moth):
         time.sleep(1)
         return None
 
-    def ack(self, m: sarracenia.Message) -> None:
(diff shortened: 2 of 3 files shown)


petersilva:
the flakey broker failures are not relevant.  They were there before... should be looked into, but separate problem than the PR.

petersilva:
while I was testing... I noticed that when it calculated "attempt_duration" it would often take 0 time... that would result in very high speed attempts at re-connection, which I think is undesirable when connections are suffering due to load.  It was quickly trying hundreds of re-connections before the frequency dropped.

With these changes, it now waits at least 1 second before attempting again.

reidsunderland:
We just reviewed it in our meeting, we're okay with the change but are still worried that the maximum time is too long. 

@junhu3 was thinking it would be best to have a hardcoded minimum for `eboIntervalMaximum` and a configurable maximum, then randomly choose a value within that range.

e.g.

```python
# can't have max < 60
max = max(60, self.o.eboMaximum)
eboIntervalMaximum = random.randint(60, max)
```

I'm not totally sure that it's worth adding the option, I think it's unlikely to be changed by users but it might be nice for us to set in default.conf on our clusters.

I think the randomness is mainly helpful for clusters with lots of configs that were started at the same time? Each unique client already has randomness since it's unlikely that >= two users would have started their configs at the exact same time.

petersilva:
The  interval already has a minimum of 1 minute, and it maxes out at 4 minutes.
the existing code is...

```

eboIntervalMaximum = 60 + random.random()*180

```

The 60 provides that floor.  max is 4min (60+180)

Another reason for the randomizing is that things are naturally synchronized when a failure occurs.  the timer can start, worst case, when the failure happens, so it can be like a starting gun and without randomization, everybody can stay in formation.

petersilva:
1) I looked at some numbers and it turns out that, assuming the connection attempt fails instantaneously, it takes 270 seconds (almost five minutes) before it has backed off to a minute... 

justification:
```
defix% python3
Python 3.12.3 (main, Feb  4 2025, 14:48:35) [GCC 13.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import math
>>> math.log(60,1.2
... )
22.456722256102744
>>> tot=0
>>> for i in range(1,22):
...    tot += 1.2**i
... 
>>> tot
270.03071945621804
>>> 

```

So we will only hit one minute of backoff after being down nearly five minutes. If the server comes back faster than that, it will try sooner.  If we have an outage that has lasted five minutes... all deadlines are blown anyways... not sure what the point is of tuning the eboIntervalMax.

The other thing is, there are multiple ebo's in the code.  It isn't clear if this setting should apply to all of them, or just this one.  It's further unclear whether all ebo's should have the same boundaries... maybe we need a separate setting per ebo...

there is an ebo in moth/amqp/ack()   (ebo max = 60 ) 

there is the main one we are dealing with which is about publish/subscribe/opening connections.

in moth/mqtt/getSetup() there is an ebo while we wait for the
broker to accept the connection, because the protocol is async.
(this is after we are connected, but a broker can be slow to accept subscriptions and such.) we need to understand if the subscription 
is active before returning.  (ebo max = 512) 

same in putSetup()...

So if we want settings:  we probably want multiple ones for each situation.   eboAckIntervalMax,  eboMQTTConnectionCompletionMax,
eboConnectionIntervalMax (the normal one?)   

I can't think of a situation where one would vary the maximum ebo interval.  It is not something one would tune because in one case, it makes sense to have a longer one, versus some other case a shorter value makes sense.   
We should only be creating settings when there are legitimate reasons to pick different values.

We should not provide variables just because people disagree what the value should be. We should agree on a value, rather than complicating the lives of all who come after.  If we provide settings, we have to document it, and provide some examples for when one would want to raise or lower it.  I can't think of any advice other than "taste"

What value do people want to put in default.conf, and why in what situation is 4 minutes bad?  If we want to put some value in default.conf, why isn't that value good in every case? 

The previous behaviour was buggy and caused problems, but adding a setting because of a no longer existing bug doesn't seem right.

petersilva:
If four minutes is too long, we'll just change the 180 to 60... and then max would vary between one and two minutes?  Would that make people happy?

reidsunderland:
I didn't think about all the failures being synchronized to when the broker fails, that makes sense. And I agree with all the other points. 

It looks like the ebo in the AMQP ack method is never reached. I'm pretty sure all code paths return before hitting the ebo and sleep. 

4 minutes as a maximum still feels a little bit too long to me, even though it would take a long time to actually reach 4 minutes. I have no data to back up my opinion but I'd definitely be happy with a range between 60-120 seconds. @junhu3 @andreleblanc11 @kurt2x any objection to that?

andreleblanc11:
60-120 seconds works for me.

petersilva:
If the ack/ebo code path is bad, should we include a cleanup/removal in this branch?

reidsunderland:
I think so, yeah. I'm not missing anything, right? Every case here has a return:

https://github.com/MetPX/sarracenia/blob/07a0ae48e3f2656277198137cfe66c3ee9e1fc6a/sarracenia/moth/amqp.py#L586-L639

petersilva:
working on:

* moth/amqp/ack() -- removing loop and ebo references.
* moth/*/ack() making ack return bool.

run local flow tests now before commit & push.

petersilva:
OK the ack refactor is done. ready for review/approval.

reidsunderland (approved):

PR #1446, opened 2025-05-21
Fix #1444

### LINKED ISSUE #1444: post_baseUrl not properly assigned in a poll/flow component
This looks to be affected by #1428.

`post_baseUrl` doesn't correctly get assigned in a component that uses `pollUrl` (poll/flow) . 

Normally this isn't a problem in a normal poll, where we don't use `post_baseUrl` usually. 
However, if we have a scheduled flow that downloads, we need to change the `post_baseUrl` to advertise the local file.

Based on the output of my affected config

```
# My config - flow/opg_wiski.conf
...
pollUrl https://kiwis.opg.com
post_baseUrl http://${HOSTNAME}
...

# sr3 show flow/opg_wiski
...
 'publishers': [{'auto_delete': False,
                 'baseDir': '/apps/sarra/public_data/',
                 'baseUrl': 'https://kiwis.opg.com',
                 'broker': 'amqps://feeder@ddsr-dev.cmc.ec.gc.ca/',
                 'durable': True,
                 'exchange': ['xpublic'],
                 'exchangeDeclare': True,
                 'format': 'v02',
                 'messageAgeMax': 0,
                 'messageDebugDump': False,
                 'persistent': True,
                 'timeout': 300,
                 'topicPrefix': ['v02', 'post']}],
```

I think the problem is here

https://github.com/MetPX/sarracenia/blob/57b53a208a94b48343b431c6e9eae809f9205abd/sarracenia/config/publisher.py#L63-L69

I don't think `post_baseUrl` should ever show up as an attribute based on the logic in the for loop.

I think we need to check for `if not 'baseUrl' in a` instead

andreleblanc11:
Previously we would compare `post_baseUrl` with `a`, but there's a problems with this.

`a` will always be `topicPrefix` because its the last element of the list in the for loop, which doesnt make sense.

### DIFF
--- a/sarracenia/config/publisher.py
+++ b/sarracenia/config/publisher.py
@@ -65,7 +65,7 @@ class Publisher(dict):
             if hasattr(options, aa):
                 self[a] = getattr(options,aa)
 
-        if not 'post_baseUrl' in a and hasattr(options,'pollUrl') and options.pollUrl:
+        if not 'baseUrl' in self and hasattr(options,'pollUrl') and options.pollUrl:
             self['baseUrl'] = options.pollUrl
 
         if not 'baseDir' in self and not self.baseDir:


petersilva (approved):

petersilva:
I am a bit puzzled about this. I agree the existing logic was wrong, and blessed the change.
I am worried that... sometimes you actually need the baseUrl setting to fix paths when downloading somehow... I think it´s a very edge case, if it even exists... which is why I blessed it.

andreleblanc11:
Yeah it doesn't happen often. Sometimes it's required for a poll, sometimes also for a scheduled flow that needs to download (like the OPG wiski stuff).

PR #1448, opened 2025-05-23
message baseUrl from poll_baseUrl if avail... dir also

petersilva:
fix what was broken by original fix for #1444
all the flow tests failing since the last merge... this fixes it.


andreleblanc11:
Flow tests now pass.

There's still a problem with the flakey broker test for both MQTT and AMQP flow tests. But it seems to have been present before my patch.

andreleblanc11 (approved):

PR #1451, opened 2025-06-04
More multi config unit tests

petersilva:
close #1450 

* remove some dead code: config/get_source_from_exchange (isn't called anywhere.) 
* move exchange resolution into subscribtion.  There seem to be some cases where setting the global *exchange* config option pollutes other subscriptions (some configurations get exchanges resolved for different subscriptions mixed together.)
* make a publisher comparator __eq__ that compares more fields so that publishers no longer match when format, or topicPrefix don't.
* added some unit tests for multi subscriber/publisher cases.

PR #1452, opened 2025-06-05
retry failed fileOps

### LINKED ISSUE #1366: ln -sf a b ... leaves tmp files around in mirroring use case.
As discovered by client, reproduced in sarrac test cases... in an hpc mirroring configuration, when 
* you have an *ln -sf a b* command.
* *a* (the thing being linked to) does not exist.
*  you end up with two links in the destination... the normal one, an a second one with ln -sf temp name (typically starts with Cuxxxx)

### LINKED ISSUE #1377: rmdir processing broken
pairs with https://github.com/MetPX/sarrac/issues/198

when testing mirroring... I end up with a whole pile of empty directories.

reidsunderland:
Fix #1366 and fix #1377 by retrying fileOps that fail. 

The changed behaviour is described in the comments.

The main changes are:

- When a directory removal fails, it retries (previously we'd ignore the error and reject the message).
   - This fixes the problem where a directory removal fails because it's not empty yet, because other instances have not yet deleted the files/subdirectories inside it.
- When link renames fail, also retry them, assuming the link has not been created yet. Previously we'd ignore the error and fall through to "downloading" the link from the source. This could leave behind temporary links as described in #1366.

Flakey broker fails with 36/37, but I think that's because of other changes?

I'll also have a PR coming for sarrac that adds tests to simulate these cases.

### DIFF
--- a/sarracenia/flow/__init__.py
+++ b/sarracenia/flow/__init__.py
@@ -858,7 +858,7 @@ class Flow:
             if self.o.post_baseDir:
                 new_dir = self.o.variableExpansion(self.o.post_baseDir, msg)
         d=None
-       
+
         if 'subcription_index' in msg:
             old_baseDir = self.o.subscriptions[msg['subscription_index']]['baseDir']
         else:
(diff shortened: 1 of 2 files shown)


petersilva:
the fix for flakey is #1451 ... commit: 6394442c989f859e6aef62da995d06bf6895dde5

petersilva (approved):

PR #1455, opened 2025-06-09
Add JSON option + better convention for filenames for WISKI

andreleblanc11:
Some more client requests.
- Add an option to formulate JSON files.
- Have `_` be the field delimiter and have fields be separated with `-` instead of `_`


andreleblanc11:
I think the new `test_multi` unit test needs to have `expire 7h` defined so that the unit test doesn't fail.

petersilva (approved):

petersilva:
The expire failure in the unit test comes from default.conf ...  not sure how to fix it yet...  the default.conf in the static_flow ...  tests from sr_insects...

PR #1458, opened 2025-06-12
Add queueType logic for Quorum queues/streams

andreleblanc11:
Branch is not ready. Things to do
- [X] Validate U18.04 flow test works (quorum queues not supported on RMQ 3.7)
- [x] Add unit tests
- [x] Add documentation on new option
- [x] Validate with U22.04/U24.04 flow tests + unit tests


andreleblanc11:
I tried running the unit tests on U18.04 (using RabbitMQ 3.7 as the broker) and defined `queueType quorum` from the `default.conf`. I ran the static and flakey broker unit tests.

I wanted to test if adding the `x-queue-type` variable would break anything, since RabbitMQ 3.7 doesn't support the quorum queues yet. 

Thankfully both tests passed. The `x-queue-type` option was defined in each queue, but didn't seem to do anything.

petersilva:
I guess a unit test would just confirm that the setting entered shows up in the config? You already have a unit test for that... You can't test more than that without setting up a broker, which so far isn't done in the unit tests.  So I think you can check that box.

The flow tests run fine on ubuntu 22 and 24 on github... so you can check the other box also.

I wonder if there should be a check for illegal values (give an warning/error if the value isn't in the known set.)
maybe check it in the __init__ or getSetup functions?...   somebody puts "basic" in... the option...
it just gets ignored... which is fine, but should probably tell the user the option is bogus.

petersilva:
fwiw... Wouldn't people normally do this with broker-side policies?
no objection to the addition, but it's kind of a broker side decision (like queue mirroring was done with policy before.) so not sure would even use this.  Nice to have the option though.

Another thought... should we maybe just make a *queueProperty* setting...
```

queueProperty x-queue-type classic

```
That way if some other similar thing comes out, we don't need a new option.
could have done that with expiry also, but that's portable because other protocols have
the same concept.  Here, the concept... It feels proprietary... I'm wonder how portable (e.g. to mqtt) the "type" is... on MQTT, we could recast receiveMaximum and maybe some other stuff as Properties...
I think the idea that queues will have settings is portable, even if the settings themselves aren't.

It would be cleaner, but would require a moderate re-factor of the moth classes.  probably not worth it.

andreleblanc11:
> Wouldn't people normally do this with broker-side policies?

Yes that's right. We've discussed myself and @reidsunderland if we would still might want to have the option and we agreed it might not be a bad idea to add it.

I think my original thought behind adding the configurability was that we might have some operational use-cases where *streams* might be interesting to explore.

From the documentation : https://www.rabbitmq.com/docs/streams#use-cases
> RabbitMQ Streams shine for the following use cases:
>Large fan-outs
>
>When wanting to deliver the same message to multiple subscribers users currently have to bind a dedicated queue for each consumer. If the number of consumers is large this becomes potentially inefficient, especially when wanting persistence and/or replication. Streams will allow any number of consumers to consume the same messages from the same queue in a non-destructive manner, negating the need to bind multiple queues. Stream consumers will also be able to read from replicas allowing read load to be spread across the cluster.
>
>Replay (Time-travelling)
>
>As all current RabbitMQ queue types have destructive consume behaviour, i.e. messages are deleted from the queue when a consumer is finished with them, it is not possible to re-read messages that have been consumed. Streams will allow consumers to attach at any point in the log and read from there.
>
>Throughput Performance
>
>No persistent queue types are able to deliver throughput that can compete with any of the existing log based messaging systems. Streams have been designed with performance as a major goal.
>
>Large backlogs
>
>Most RabbitMQ queues are designed to converge towards the empty state and are optimised as such and can perform worse when there are millions of messages on a given queue. Streams are designed to store larger amounts of data in an efficient manner with minimal in-memory overhead.

The large backlog perfomance could be useful for some of our large traffic senders (PPP senders) in the event of an outage. The replay feature could also be interesting to use.

We'll likely be using quorum queues for replication once we move to RabbitMQ 4.0, but *streams* [also support replication](https://www.rabbitmq.com/docs/streams#member-management)

---

For the queueProperty stuff, I also agree it might be worth exploring, but it might be a bit of work to get what we want out of it.

andreleblanc11:
FWIW I haven't bothered with adding a `basic` entry for `queueType`.

The guard that's already in place in the `x-queue-type` logic should prevent erroneous entries from making their way in the queue definition. Also, I added a warning in the logs so the user should get alerted when an erroneous entry is placed in the config.

petersilva (approved):

PR #1460, opened 2025-06-14
co-ordinate start/stop with sanity, add *new* and *inter*

### LINKED ISSUE #1069: strays/missing on sr3 status during start or stop
there is a window of time while "stop" and "start" operations are in progress, where the status will be marked as *missing* because either instance processes have been killed but the pid files not removed (stopping) or the instance processes have been launched, but not yet written their pid files so they can be claimed.

so *sr3 sanity* if runing while a flow is being stopped, will restart it, rather than allowing the *stop* to complete.  When in start up, *sr3 sanity* can kill some instance processes that are marked stray because they have not fully initialized yet, and then restart those instances.

sub-optimal.

### LINKED ISSUE #1439: How to handle `sr3 sanity` running at the same time as a `sr3 stop/restart/start`
`service metpx-sr3 start` will perform an `ExecStop` and stop all processes if 
   1. Sr3 is already running
   2. Sr3 was not originally started by the service

## How to reproduce

```
# From your user that runs sr3
sr3 stop
sr3 start

sudo service metpx-sr3 start
# hangs for a little bit
service metpx-sr3 status
  
    ● metpx-sr3.service - Sarracenia File Copy Service
       Loaded: loaded (/lib/systemd/system/metpx-sr3.service; enabled; vendor preset: enabled)
       Active: inactive (dead) since Thu 2025-05-15 15:01:32 UTC; 1min 8s ago
      Process: 8796 ExecStop=/usr/bin/sr3 stop (code=exited, status=0/SUCCESS)
      Process: 8568 ExecStart=/usr/bin/sr3 start (code=exited, status=0/SUCCESS)
    ...
    May 15 15:00:46.894 test-server sr3[8568]: starting:( 0 ) Done
    May 15 15:00:49.114  test-server sr3[8796]: Stopping: sending SIGTERM ............................................................................................................................................................................................................................................................................... ( 271 ) Done
    May 15 15:01:32  test-server sr3[8796]: Done
    May 15 15:01:32  test-server sr3[8796]: Waiting again...
    May 15 15:01:32  test-server sr3[8796]: All stopped after KILL
    May 15 15:01:32  test-server systemd[1]: Started Sarracenia File Copy Service
```

The `ExecStop` gets executed after the `ExecStart` and I think that's because the `service metpx-sr3 start` doesn't fork anything despite having `Type=forking` so systemd assumes something went wrong and does a `ExecStop`.

https://github.com/MetPX/sarracenia/blob/8e50c2043346bd1d8d60ee1a240da2a992254026/debian/metpx-sr3.service#L13-L17

petersilva:
close #1069 
close #1439 

This PR makes progress on #1439, but does not address the originally reported issue of interaction between systemd initiated flows and manually initiated ones, as well as those initiated by sr3 sanity.
It completely ignores that issue, and I opened #1461 to deal with that problem.

The other half of #1439 is a general problem made obvious by this report that if *sr3 sanity* is running at the same time as maintenance activities occur, like *sr3 start*, or *sr3 stop*, then race conditions can result in undefined results where sr3 sanity kills or starts the wrong processes, or *start* or *stop* operations have unexpected or incomplete results.   That's what this PR addresses.

The improvements in this PR:

* there were some delay timers (now called *instance_gap* in the code) that would wait a long time while starting things up.  Those have been shortened, so startup should be much quicker.
* new flow states: *starting* and *stopping* are introduced.  *sr3 sanity* will not make any changes when any flow is in one of those states.
* new flow state: *new* introduced to identify a flow that has no state files, as would be the case when it is *added*, or after a *clean* operation.
* new flow state: *interactive* introduced to identify flows that don't normally run as daemons.
* global flag *sanitizing* introduced to the root of the state directory so that any *start* or *stop* operations will block before starting.  They will loop waiting for the sanitizing to be complete.
* when executing *sr3 start*, it used to launch the processes and exit.  Now instead it looks in the state file for the requisite number of pid files to indicate successful start up, and only exits after that is complete (to prevent a santizing run from happenning before the pid files are created.)
* Before, when all instances of a flow had crashed, sr3 decides the flow is stopped, so *status* reported it as *stopped* and *sanity* would not restart it.  In this branch, added a *running* statefile to the state directory, to differentiate between *stopped* versus running, but all instances have crashed.
* during development, encountered orphan *sr3_cpump* processes (exit status not read by parent pid.) that cause sr3 to crash, because lots of attributes (e.g. 'cmdline' ) are None.  So now filter out defunct processes from flow management.
* the *statehost* variable was not being used to understand where the state files of a given configuration were.  This would result in *sanity* starting up processes to replace existing functional ones.

## Initial unit testing: make everything slow...

Guaranteed overlap in between maintenance activities ( *start*, *stop*, and *sanity*) by adding sleeps so that each one takes a really long time.  Saw all the ways they could interfere with each other and results.  started by testing
with a single flow (sr3 add subscribe/dual_amis) then using the *static_flow*
suite of flows.  Once that seemed functional, moved on to:

## ran *sr3 sanity* every 3 seconds throughout the flow_tests.

ran static, flakey, dynamic, flow tests on local PC.  Also had
*watch -n 0.5 sr3 status* running to see what it was reporting 
at all times.

## Results:

* prior to this work:
   * running sr3 sanity would sometimes break things (timing sensitivity.)
   * running sr3 sanity would sometimes start wrong things (statehost out of sync.)
   * running sr3 sanity would sometimes not start things it should (all instances crashed.)
   * flow_tests would fail when *sr3 sanity* was run in the background.
   
* with this branch:
   * there is no known, reproducible condition where sr3 sanity does not do the right thing.
   * start, stop, sanitize actions co-ordinate to avoid undefined behaviour.
   * all flow_tests succeed regardless or sr3 sanity being run continuously throughout.
   * sr3 status now shows additional flow states:  new, interactive, starting, stopping.

### DIFF
--- a/docs/source/Explanation/CommandLineGuide.rst
+++ b/docs/source/Explanation/CommandLineGuide.rst
@@ -627,7 +627,9 @@ will be:
 * reje:  all processes running, but too high percent of messages being rejected (runStateThreshold_reject )
 * rtry:  all processes running, but too large number of transfers failed and retrying (runStateThreshold_retry )
 * run:   all processes are running (and transferring, and not behind, and not slow... normal state.)
+* shut:  in the process of shutting down (stop or restart in progress.)
 * slow:  transfering less than minimum bytes/second ( runStateThreshold_slow )
+* star:  in the process of starting up (start or restart in progress.)
 * stop:  no processes are running. 
 * stby:  Standby mode: all processes running, but messages are being stored in the local download_retry queue.
 * wVip:  process doesn't have the vip (only applies when the vip option is specified in the config)
--- a/docs/source/Explanation/CommandLineGuide.rst
+++ b/docs/source/Explanation/CommandLineGuide.rst
@@ -622,10 +622,12 @@ will be:
 * down:  cannot connect or exchange data with remote data source or sink.
 * hung:  processes appear hung, not writing anything to logs.
 * idle:  all processes running, but no data or message transfers for too long (runStateThreshold_idle)
+* inte:  Interactive configuration, calls from CLI or scripts (not a daemon.)
 * lag:   all processes running, but messages being processed are too old ( runStateThreshold_lag )
 * part:  some processes are running, others are missing.
 * reje:  all processes running, but too high percent of messages being rejected (runStateThreshold_reject )
 * rtry:  all processes running, but too large number of transfers failed and retrying (runStateThreshold_retry )
+* new:   no state files exist for this configuration (fresh after *add* or *cleanup* )
 * run:   all processes are running (and transferring, and not behind, and not slow... normal state.)
 * shut:  in the process of shutting down (stop or restart in progress.)
 * slow:  transfering less than minimum bytes/second ( runStateThreshold_slow )
--- a/docs/source/fr/Explication/GuideLigneDeCommande.rst
+++ b/docs/source/fr/Explication/GuideLigneDeCommande.rst
@@ -621,10 +621,12 @@ sera :
 * down : impossible de se connecter ou d'échanger des données avec une source ou un récepteur de données distant.
 * hung : les processus semblent bloqués et n'écrivent rien dans les journaux.
 * idle : tous les processus en cours d'exécution, mais ne transfert pas depuis trop longtemps (runStateThreshold_idle.)
+* inte : configuration à usage intéractif, n'éxecute pas comme un daëmon (processus de service en arrière plan.)
 * lag : tous les processus en cours d'exécution, mais les messages en cours de traitement sont trop anciens ( runStateThreshold_lag )
 * part: certains processus sont en cours d'exécution, d'autres manquent à l'appel.
 * reje : tous les processus en cours d'exécution, mais un pourcentage trop élevé de messages rejetés (runStateThreshold_reject )
 * rtry : tous les processus en cours d'exécution, mais un grand nombre de transferts échouent, causant d'autres tentatives (runStateThreshold_retry )
+* new : une configuration neuve (nouvellement installéé/créé, ou après un opeations *clean* )
 * run : tous les processus sont en cours d'exécution (et en transfert, et pas en retard, et pas lents... état normal.)
 * shut : en cours d'arrêt (stop ou restart en cours.)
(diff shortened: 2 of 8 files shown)


petersilva:
sample result from *sr3 sanity* run when there is a conflict:

```

sanity: 2025-06-13 18:32:25,186 601407 [WARNING] root sanity Not interfering with operations in progress: {'subscribe/q_f71': 'starting', 'subscribe/amqp_f30': 'starting', 'subscribe/u_sftp_f60': 'starting', 'subscribe/rabbitmqtt_f31': 'starting', 'subscribe/cfile_f44': 'starting', 'subscribe/cdnld_f21': 'starting', 'subscribe/cp_f61': 'starting', 'subscribe/ftp_f70': 'starting', 'watch/f40': 'starting'}

```
The sanitize aborts... whereas start|stop loop waiting (can be interrupted with ^C)

when sanity is in progress during a start, messages will be like:

```

2025-06-13 18:32:25,186 601407 [INFO] sanitizing in progress, please wait. 
2025-06-13 18:32:35,186 601407 [INFO] sanitizing in progress, please wait.
2025-06-13 18:32:36,186 601407 [INFO] sanitize complete, proceeding with start

```

petersilva:
fwiw... when I'm on the development branch, if there are no configs present, 
running *sr3 status*, I get this:
```
  File "/home/peter/.local/bin/sr3", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/peter/Sarracenia/sr3/sarracenia/sr.py", line 3132, in main
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/peter/Sarracenia/sr3/sarracenia/sr.py", line 1276, in __init__
    self._resolve()
  File "/home/peter/Sarracenia/sr3/sarracenia/sr.py", line 1093, in _resolve
    if not self.procs[pid]['claimed']:
           ~~~~~~~~~~~~~~~^^^^^^^^^^^
KeyError: 'claimed'
```

This error goes away on the PR branch.

reidsunderland (approved):
@andreleblanc11 and I reviewed it, it looks good. We'll start testing the development branch on dev.

PR #1465, opened 2025-06-17
Mqtt fixes mostly

petersilva:
On my machine, static, flakey, and dynamic all pass now on MQTT.
There is still a problem with restart_server, will fix later.  merge this to make incremental progress.

### DIFF
--- a/sarracenia/flow/__init__.py
+++ b/sarracenia/flow/__init__.py
@@ -1299,6 +1299,7 @@ class Flow:
                     self.work_message_adjust(new_m)
                     post_messages.append(new_m) 
                     i += 1
+            m['_deleteOnPost'] |= set(['publisher_index'])
                     
         if len(self.o.publishers) > 1:
             self.worklist.ok=post_messages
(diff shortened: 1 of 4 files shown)


petersilva:
so... on github, absolutely everything MQTT passes.  maybe can declare MQTT officially working now?
That's kind of the last mile stone.  Would like to get this merged.

otoh... on my laptop, restart_server never passes, it's always one failure with a few files missing in one copy.
So I think there is still something wrong, but don't have time to work on it, now, and these fixes definitely fix dynamic, and make everything better.

reidsunderland:
do you have `peristence true` in your mosquitto broker config? I think that should impact flakey_broker and not restart_server, but it's probably still worth double checking

https://github.com/MetPX/sarracenia/blob/c641837ffc628cb0832247eb90c909b202db6d12/travis/flow_autoconfig_add_mosquitto.sh#L72

reidsunderland:
I will test that the copy.deepcopy works on python 3.6/ubuntu 18

petersilva:
I think deepcopy only has a problem with complex data structures on older pythons... in this case, the removal of all the internal/working variables is done, so it should be safe... but yes... testing 18.04 is worthwhile...

petersilva:
@reidsunderland yes persistence is on on my mosquitto broker.

reidsunderland (approved):
I didn't find problems on ubuntu 18

PR #1470, opened 2025-06-28
No installsystemduser in 1804

petersilva:
check if dh_installsystemduser exists before using... so no failure on 18.04.

Noticed the deb tools package linter did not like executable permissions of python modules that aren't directly user invoked, so removed.


andreleblanc11:
Running from the cloned git repo still works.

File perms. from the source apt package already don't include the executable permissions.

andreleblanc11 (approved):

petersilva:
Since you guys are in release mode, I'm not going to merge it.
@reidsunderland might have some preferences. dunno if this results in rc2 or something.

reidsunderland:
I merged it, I haven't quite decided if it should be rc2 or if I should just re-tag rc1, since the code is the same and it's just a packaging change

PR #1471, opened 2025-07-07
fix typo runStateThreshold_slow --> runStateThreshold_idle

reidsunderland:


petersilva:
I think the doc is correct as-is.   Normally, low flow rates just indicate that there isn't much data going through.  For some flows, a low rate of transfer means there is something wrong.   
These two thresholds have different meanings:

runStateThreshold_slow --> if tranmitting slower than this rate... mark the state as *slow* meaning we expected a higher rate of transfer... aka *slow* is a warning of insufficient flow.

runStateThrehold_idle --> having a low transfer rate is not a problem, it is just tagged as idle when that happens.

So if low rates are not a problem, you set _idle, but if they ARE a problem, you set _slow.

then *sr3 status* will show those flows as being in *slow* state, rather than idle, meaning they are not running as fast as admins expect they should.

reidsunderland:
I might misunderstand the purpose of idle. I thought idle was equivalent to Columbo's "last transmission/reception" time.

i.e. if we expect at least 1 file every hour, and it's been >1 hour since we last transferred a file, then that flow should be marked idle?

image

reidsunderland:
runStateThreshold_slow is a rate, but runStateThreshold_idle is a period of time

petersilva:
idle is like "epsilon"... in a lot of numerical analysis... it's when does the flow rate fall to a low enough number that it might as well be zero...  

if you expect one file to go every hour, and it's been two hours, and that's a problem, then the threshold to set should be *slow* ... if not having a file after two hours is no cause for alarm... you can have the idle threshold set so that it appears as *idle*.    

*idle* should be the state when it's normal,  *slow* when it's a bad sign.

from the sr3_options man page:

> The runStateThreshold_idle option sets how long to consider too long before declaring no transfers are occurring. the sr3 status command will show such flows as idle ... This isn’t a problem in itself, unless one is expecting a continuous flow. If a continuous flow of a certain rate is expected, set the runStateThreshold_slow for the flow so that sr3 status flags it as a problem.


vs.

> The runStateThreshold_slow option sets how many messages bytes per second to expect this flow to transfer normally.  If the data RxData and TxData fields combined in the sr3 status command is lower than this, then the flow will be listed as slow.

petersilva:
I see what you are saying about *rate* vs. *interval* ...  what you are saying makes sense... have to look at the code, I guess.

petersilva:
looking at the code, I think it works documented, in that if there is a _slow threshold, the flow will never look at *_idle* thresholds, since being too slow is already an error state.  _idle is only considered if the _slow threshold is zero.

https://github.com/MetPX/sarracenia/blob/38322350a60e374c0ddcd06e99e71474b397a668/sarracenia/sr.py#L1113-L1135

petersilva:
still the weirdness about interval vs. rate stands. dunno what to do about that.

reidsunderland:
I think it's rare that we would know what the expected transfer rate is for a feed, so it's unlikely we'd see the slow state or set the threshold. 

But we do know how frequently files are supposed to come in for a lot of feeds, so idle will be useful. 

Slow is based on the data (bytes) transfer rate, and idle is based on file transfer times (although defined in an interval)

It should show an `idle` state when `Last` (transfer) is > `runStateThreshold_idle`.

```
Component/Config                            Processes                                         Rates
                                            State   Run Retry  Que     Lag      Last    %rej  messages      Data
                                            -----   --- -----  ---     ---      ----    ----  --------      ----
sarra/cvt_dms_to_wxo-dd_swob                idle    1/1    0    0    0.00s     2m38s    0.0%     0m/s       0B/s
sarra/cvt_fn_2bulletin                      idle    1/1    0    0    0.00s    49m22s    0.0%     0m/s       0B/s
sarra/cvt_rawca_to_bulletin_loggernet       idle    1/1    0    0    0.00s    10m15s    0.0%     0m/s       0B/s
sarra/cvt_rawca_to_bulletin_pxpaz           idle    1/1    0    0    0.00s       n/a    0.0%     0m/s       0B/s
sarra/cvt_sxvd02_to_sxvd62                  idle    1/1    0    0    0.00s       n/a    0.0%     0m/s       0B/s
sarra/cvt_uppercase_accents                 idle    1/1    0    0    0.00s     5m26s    0.0%     0m/s       0B/s
sarra/get-cmems-oceanobs                    idle    4/4    0    0    0.00s    41m47s    0.0%     0m/s       0B/s
sarra/get-lrgs-data2                        run     1/1    0    0    1.37s     2m30s    0.0%  0.00m/s    0.59m/s
sarra/get-msc-sat-ops-poes                  run     1/1    0    0    0.10s    15.41s    0.0%  0.02m/s 550.2KiB/s
sarra/get-msc-sci-cmcops-ppp-cmot           idle    1/1    0    0    0.00s    17h38m    0.0%     0m/s       0B/s
sarra/get_airnow_ab                         idle    1/1    0    0    0.00s      3h4m    0.0%     0m/s       0B/s
```

Both states function correctly.

But I think I understand what the documentation is saying now, so I realize that what I thought was a typo isn't. I'll see if I can clarify the documentation more.

reidsunderland:
I changed `runStateThreshold_idle` a bit more. What do you guys think of this:

runStateThreshold_idle <interval> (default: 900)
------------------------------------------------

The runStateThreshold_idle option sets how long to consider too long before declaring no transfers are occurring.
the *sr3 status* command will show such flows as *idle*

In a flow that posts data, the last activity will be based on when it posted last.
In a flow that transfers data, the last activity will be based on the last data transfer.
In a flow does neither of the above, then the last activity is based on the last message
received.

Idle isn't a problem in itself, unless one is expecting to receive data at regular intervals (e.g. at least
one file every 15 minutes).

If a continuous flow of a certain data rate (bytes per second) is expected, set
*runStateThreshold_slow* for the flow so that *sr3 status* flags it as a problem (too slow).

petersilva:
fwiw.. I'm ok with this PR, but I'm not on the list of reviewers... need @andreleblanc11 to bless.

andreleblanc11 (approved):

reidsunderland:
oops, sorry @petersilva 😄

reidsunderland:
I need to add the French translation now that the English is blessed

reidsunderland:
I finally added the French translation (thanks @mksmsn)

PR #1473, opened 2025-07-24
fix #1472 add missing _isRetry flag for retried messages

### LINKED ISSUE #1472: some retried messages are not tagged as retries
In #1206, we decided that retry messages would not be included in the calculation of `lag`.  This part works correctly, but the retry.py code doesn't always set `_isRetry` in the message. The `after_accept` method in retry.py needs to be updated to set it.

Related history: In #466 we decided to remove the `isRetry` field from messages, but it was added back later in #920. #527 was opened to note that a plugin was broken by the removal of isRetry in #466, so I guess now that `_isRetry` has been added back I can also fix that plugin.

reidsunderland:
When retry_refilter is False and the after_accept entry point of retry.py was used, `_isRetry` did not get set on the messages to be retried. 

When testing this to confirm it didn't break retry_refilter, I found an unrelated problem with retry_refilter. It would crash with this error:

```
2025-07-24 20:22:49,933 [DEBUG] 3038018 sarracenia.flow download https_transport remote cd to /20250724/WXO-DD/citypage_weather/mp3/SK
2025-07-24 20:22:49,933 [DEBUG] 3038018 sarracenia.transfer.https cd sr_http cd /20250724/WXO-DD/citypage_weather/mp3/SK
2025-07-24 20:22:49,933 [DEBUG] 3038018 sarracenia.flow download Exception details:
  File "/net/local/home/sunderlandr/sr3/sarracenia/flow/__init__.py", line 2303, in download
    (urlstr, remote_offset, block_length-1, new_inflight_path, msg['local_offset'],
KeyError: 'local_offset'
2025-07-24 20:22:49,933 [WARNING] 3038018 sarracenia.flow download failed to write s0000300_sa_f.mp3: 'local_offset'
2025-07-24 20:22:49,933 [DEBUG] 3038018 sarracenia.transfer.https close sr_http close
2025-07-24 20:22:49,933 [DEBUG] 3038018 sarracenia.transfer.https init sr_http init
2025-07-24 20:22:49,933 [INFO] 3038018 sarracenia.flow do_download attempt 3 failed to download https://dd1.weather.gc.ca//20250724/WXO-DD/citypage_weather/mp3/SK/s0000300_sa_f.mp3 to /tmp/citypage/s0000300_sa_f.mp3
```

local_offset is required for downloads to work properly and it is set in the message by the moth classes, e.g.: 
https://github.com/MetPX/sarracenia/blob/6bbd6904d4666bca4880ef87c2900abd873db8a7/sarracenia/moth/amqp.py#L151-L152

but retry.py in retry_refilter mode deletes all the delete_on_post values from the message, including local_offset, which causes the crash.

I fixed it by not deleting `_isRetry` in that case, but I wonder if it would be best to initially set local_offset = 0 somewhere else in the code? It feels weird having that happen in moth...

### DIFF
--- a/sarracenia/__init__.py
+++ b/sarracenia/__init__.py
@@ -1152,3 +1152,6 @@ class Message(dict):
             msg.computeIdentity(opath,self.o,data=data)
         except Exception as ex:
             logger.error( f"problem with {opath}: {ex}" )
+
+    def isRetry(msg):
+        return '_isRetry' in msg and msg['_isRetry']
--- a/sarracenia/__init__.py
+++ b/sarracenia/__init__.py
@@ -1155,3 +1155,15 @@ class Message(dict):
 
     def isRetry(msg):
         return '_isRetry' in msg and msg['_isRetry']
+
+    def retryCount(msg):
+        """ return the number of times the message has been retried.
+            0 if the message has never been retried.
+        """
+        rcount = 0
+        if '_isRetry' in msg:
+            if type(msg['_isRetry']) == int:
+                rcount = msg['_isRetry']
+            else:
+                rcount = 1 if msg['_isRetry'] else 0
+        return rcount
--- a/sarracenia/flow/__init__.py
+++ b/sarracenia/flow/__init__.py
@@ -611,7 +611,7 @@ class Flow:
                 #               worklist.ok by retry.py, even when last_poll_gather_len is 0
                 elif self.o.component == 'poll':
                     # get rid of non-retry messages from worklist.ok, we only want to post retries
-                    new_ok = [ msg for msg in self.worklist.ok if ('_isRetry' in msg and msg['_isRetry']) ]
+                    new_ok = [ msg for msg in self.worklist.ok if msg.isRetry() ]
                     self.worklist.ok = new_ok
                     self.post(now)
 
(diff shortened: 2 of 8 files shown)


andreleblanc11:
I think it might be best to set it just before executing the work? And check if it's `_isRetry` at the same time.

I do agree that we should reset the value to 0. We're restarting the work, so whatever value we had before should be irrelevant now.. unless there's some logic that I'm not aware of where we're starting the download from where we left off

petersilva:
I'm wondering if, as long as we're touching this, maybe change it to a count instead, so it increments every time retry is done?  ie... change *_isRetry* to *_RetryCount* so that instead of setting to True, we set it to a count.  We could then
add a "retryCountLimit" to give another way to limit how hard to deliver things... and be able to report on how many 
failures a given message has sufferred.  If the variable is either missing or 0 then it is the initial attempt.  Could even set it (to 0) for the initial attempt.

reidsunderland:
I started writing a reply to Andre and didn't click post:

I'm not sure about setting local_offset to 0 before work. I thought local_offset was related to partitioned transfers (a particular message might want to start downloading at a non-zero offset), and that's why I didn't set it to 0. 

But that doesn't seem to be true (anymore?) after taking a quick look at the code.

---

I like the idea of making isRetry a count, it makes sense to do it now while it's still relatively unused. I'll work on changing it

petersilva:
yeah... so the "partitioned_flow" is in the main flow tests, so yeah, I guess the offset !=0 is not needed anymore?
I don't know how that works... I would run partitioned flow locally to confirm/understand.

reidsunderland:
I switched isRetry to a count, but didn't rename it. I think leaving it with the current name should be backwards compatible with plugins that use it (there's one internal one).

petersilva (approved):

Issue #1478, opened 2025-07-28
Place data `inline` from other components


andreleblanc11:
With the current sr3 code, we're only able to place data *in* the sarracenia message from the gather.file entry point. This is being done while calling the [inline](https://metpx.github.io/sarracenia/Reference/sr3_options.7.html#inline-flag-default-false) option.

Myself and @reidsunderland think it would be good if we could get the file contents in the message at other points in the sr3 flow as well. There's a use case where a client wants to have the file contents inside the AMQP message, but the source doesn't put the contents inline of the message.

It's easy to write a flow callback to do add this from an `after_accept` or from an `after_work` entry point, but it would probably be better if we could just have it built in the code.

---

One approach to do this would be to have a new message method called maybe `putContentInline` that would place the data inside the message if found.

reidsunderland:
I was a bit confused when I first read this so I'll add a little bit more. There's kind of two problems here:

1. I agree with Andre that the existing implementation of `inline` is limited - we would like to be able to use it in other components like a sarra/subscriber or sender config, where we already have the data on the local system and we are publishing messages to re-advertise it.

An after_work plugin would probably be good for this. We could automatically load it when inline is enabled and we are **not** using gather.file. 

--- 

2. The other part is a utility method that plugins could used to load the content into the message itself. We have `Message.getContent()` but it reads the content of the file and returns it in a variable.

`putContentInline` would take this a little further by loading the content from file, then store it in `msg['content']` and calculate the checksum and set identity fields (if necessary), set `msg['size']` if it's not already there.

With this, the theoretical plugin would be really easy, something like:

```python
def after_work(self, worklist):
    for msg in worklist.ok:
        # TODO size is optional, but we have the file on disk so we can still check the size if it's not in the message
        # ideally putContentInline should deal with the size check
        if 'content' not in msg and msg['size'] <= self.o.inlineByteSizeMax :
            msg.putContentInline()
```

andreleblanc11:
IMO, an `after_work` plugin is a good work around. I added a custom plugin in our internal plugins repo when I ran into this, but I think it makes sense to have the source code evaluate the value of `inline` to determine if we want to place the file contents inside of the sarracenia message.

What I mean by this is we could perhaps call the `putContentInline` at the end of the following for loop.

https://github.com/MetPX/sarracenia/blob/dc3b0f539b2f6741a01b90c494eece9a7ab86c01/sarracenia/flow/__init__.py#L1288-L1303

[...]

Issue #1392, opened 2025-03-03
When using a download plugin, the size doesn't get updated/checked


reidsunderland:
I'm not sure that this behavior is wrong. It does seem to cause v2 compatibility issues though.

When a download plugin is used, the part of the code that compares the downloaded file size with the expected size in the message is not executed.

This means that there's no check to see if the downloaded size is wrong, and if there's no size already set in the message from the upstream source, the sarra/subscriber will not set it. This causes issues for downstream v2 subscribers due to #1391.

There's a lot of ways we could change this, if we want to.

Some options that come to mind are: 
- Change the return value of `FlowCB.download` to return a status code *and downloaded size* (could be optional), then modify the main flow code to look at the returned size. 
- Add a separate after_work plugin that re-implements most of the size checking code from flow. (easy, but redundant, repeated code)
    - More involved: This could be a built in plugin that is added by default when download is enabled, and then we could remove the size check code from `Flow.download`. (fixes the redundancy issue)
    - This might make download retry handling more difficult, I haven't looked at the code in enough detail to check.

https://github.com/MetPX/sarracenia/blob/372e77b633c958f185d9c7f264b3e53376b3a1bf/sarracenia/flow/__init__.py#L2236-L2307

Edit: I think this might be the same for identity, it doesn't appear to get set when the download plugin is used, but I need to adjust the options in the config and look at the code to be 100% sure

petersilva:
As we found out with checksums, this is a case for Postel's law: "be liberal in what you accept, conservative in what you send" ... meaning with checksums, when we got a mismatch, we eventually got to the decision that incorrect checksums should be overridden by correct ones on download, because to do otherwise would result in download loops. Similarly, if the size header doesn't match what is actually downloaded, then the size header should be corrected to correspond with what was actually downloaded to prevent errors on every downstream recipient.

reidsunderland:
This was partially fixed by #1546, except it only sets the size and/or checksum if it is not already present.

petersilva:
There are lots of if statements around checksum values.  It feels like much simpler logic is possible.
Just always overwrite the checksum with the onfly_ one.  When using binary downloads,
always calculate the onfly_ one after download, which should always override the one in the message.
That costs an extra file read on every download, which sounds very expensive to me... 

I have the strong feeling that we should be comparing onfly checksums with what was expected and
squawking about it, but that logic seems to have all been commented out. I have the feeling not
having a notification of content mismatch will be perceived as a bug at some future date.
[...]

PR #1479, opened 2025-07-29
#1478 Add support for inlining data to a sarracenia message from othe…

andreleblanc11:
…r components.

* I've only tested the functionality with a sarra component, and it seems to have worked well.
* I'm not sure what we should do if we can't find the size of the message? I just put in the code that we wouldn't add the file contents inside of the sarracenia message.
   * There might be a simple way (a method that already exists?) to get extract the file size.
* I know @petersilva mentionned using `putContent` as the method name, but I think `putContentInline` is better because the scope of the work is limited to the data only being inline.

### DIFF
--- a/sarracenia/__init__.py
+++ b/sarracenia/__init__.py
@@ -1112,22 +1112,33 @@ class Message(dict):
         Does not return any value.
         """
 
-        if 'size' in msg and msg['size'] >= options.inlineByteMax:
-            logger.warning(f"Not placing file contents in message due to file size being to big. File size {msg['size']}, inlineByteMax: {options.inlineByteMax}")
-            return
-        elif 'size' not in msg:
-            logger.warning(f"Cannot get file size. Won't write file contents in message as a consequence")
+        # Don't try to add data inline if it's already present.
+        if 'content' in msg:
             return
 
         try:
             content = msg.getContent()
+            sz = len(content)
+
+            # If there's not already a size in the message, give it an arbitrary value.
+            if 'size' not in msg:
+                msg['size'] = 0
+            # We want to update the message size with the recently fetched content.
+            if sz != msg['size']:
+                logger.warning(f"Size from getContent doesn't match previously assigned size. Reassigning size to {sz}")
+                msg['size'] = sz
+
+            if msg['size'] >= options.inlineByteMax:
+                logger.warning(f"Not placing file contents in message due to file size being to big. File size {msg['size']}, inlineByteMax: {options.inlineByteMax}")
+                return
+
         except Exception as e:
             logger.error(f"Couldn't fetch file contents with getContent. Error: {e}")
             logger.debug("Exception details:", exc_info=True)
+            return
 
         # We may want to call this method even if the file contents is already in the message to rewrite the entries.
-        if 'content' not in msg:
-            msg['content'] = {'value' : '' , 'encoding' : ''}
+        msg['content'] = {'value' : '' , 'encoding' : ''}
 
         try:
             msg['content']['value'] = content.decode('utf-8')
--- a/sarracenia/__init__.py
+++ b/sarracenia/__init__.py
@@ -1102,6 +1102,47 @@ class Message(dict):
         with urllib.request.urlopen(retUrl) as response:
             return response.read()
 
+
+    def putContentInline(msg,options=None):
+        """
+        Embed file data inside a sarracenia message. Leverages the
+        getContent method to acquire the file data, then inserts it
+        in the sarracenia message when possible.
+
+        Does not return any value.
+        """
+
+        if 'size' in msg and msg['size'] >= options.inlineByteMax:
+            logger.warning(f"Not placing file contents in message due to file size being to big. File size {msg['size']}, inlineByteMax: {options.inlineByteMax}")
+            return
+        elif 'size' not in msg:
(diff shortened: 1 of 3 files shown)


sarracenia/__init__.py
```diff
@@ -1102,6 +1102,47 @@ def getContent(msg,options=None):
...
+            return
+        elif 'size' not in msg:
+            logger.warning(f"Cannot get file size. Won't write file contents in message as a consequence")
+            return
+
+        try:
+            content = msg.getContent()
+        except Exception as e:
+            logger.error(f"Couldn't fetch file contents with getContent. Error: {e}")
+            logger.debug("Exception details:", exc_info=True)
```
reidsunderland: I think it might help to change the order a bit.

First check if content is already in the message, if it is there, don't do anything and return right away. 

Next try getContent, if it works and msg['size'] isn't already there, you can set `msg['size'] = len(content)`.  (i.e. if we have the contents, we will always have the size too). If getContent doesn't work then putContentInline can't do anything anyways.

Otherwise it looks good

reidsunderland (changes requested):
I added a comment about re-ordering some of the code, but it looks good

sarracenia/__init__.py
petersilva: I think the size of the message should be:

```python

    content = msg.getContent()
    sz = len(content)
    if sz != msg['size']:
         logger.warning( f"data size doesn't match message, overriding to actual size: {sz}" )
         msg['size'] = sz

```

and sz should override msg['size'] if it doesn't match.

petersilva (changes requested):
size processing should be based on content actually obtained.

petersilva:
I was thinking that "setContent" is another potential name.   A later refactor might create a setter for the message class that does the right thing depending on which field is being set.

sarracenia/__init__.py
```diff
@@ -1102,6 +1102,58 @@ def getContent(msg,options=None):
...
+            content = msg.getContent()
+            sz = len(content)
+
+            # If there's not already a size in the message, give it an arbitrary value.
+            if 'size' not in msg:
+                msg['size'] = 0
+            # We want to update the message size with the recently fetched content.
+            if sz != msg['size']:
+                logger.warning(f"Size from getContent doesn't match previously assigned size. Reassigning size to {sz}")
+                msg['size'] = sz
```
reidsunderland: the changes look great to me but i'm a bit worried that we might get a lot of warning messages because [size is optional in sr3 messages](https://metpx.github.io/sarracenia/Reference/sr_post.7.html#body-headers-json-encoding) 

maybe change to an if+elif?

```python
if 'size' not in msg:
    logger.debug?
    msg['size'] = sz
elif sz != msg['size']:
    logger.warning... 
    msg['size'] = sz
# else size is already in the message and it matches the size of the content
```

petersilva: hmm... I think size is optional because some messages are about links, directories, or renames where size isn't relevant...aka "fileOp".  I imagine that this code path should never see those. For files, the size is always supposed to be there.

andreleblanc11: That's a fair assumption

petersilva: oh yeah.,.. also polls... when polling, sometimes we can't get the real size, so just leave it blank and get it filled in on download...

petersilva: I don't think it should be a warning/error unless the size was there before and it does not match.  the current will warn on mismatch, ok, but also whenever it's zero... I think warning on zero is unhelpful.

reidsunderland: Yeah I agree, I just want to eliminate the warning for cases where size is missing. Polls are the cases I had in mind

reidsunderland: This actually reminds me, I'm not convinced that we do fill in the size after downloading. 

For example:
```
# no size in the incoming messages from the poll
2025-07-30 08:35:07,326 [INFO] sarracenia.flowcb.log after_accept accepted: (lag: 1.89 )  exchange: xs_SMAP_NSIDC subtopic: pub.DATASETS.SMAP.SPL1BTB_NRT.105 a file with baseUrl: https://daacdata.apps.nsidc.org/ relPath: pub/DATASETS/SMAP/SPL1BTB_NRT.105/SMAP_L1B_TB_NRT_56054_A_20250730T064817_N19241_001.qa id: sha512

# no size and no checksum in the message posted by the sarra after downloading
2025-07-30 08:35:07,677 [INFO] sarracenia.flowcb.log after_post posted to exchange: xpublic topic: v02.post.20250730.SMAP.NSIDC.08 a file with baseUrl: http://ddsr relPath: 20250730/SMAP/NSIDC/08/SMAP_L1B_TB_NRT_56054_A_20250730T064817_N19241_001.qa id: sha512
```

I think I've seen this before and wasn't sure if it was correct behaviour or not, but it's always been in the process of investigating other problems so I didn't think about it at the time.

reidsunderland: Oh, I remember now. It's only when using a custom downloader plugin that the sarra doesn't fill in the size and checksum. I already created an issue and forgot about it: https://github.com/MetPX/sarracenia/issues/1392

reidsunderland (approved):

petersilva (approved):

PR #1483, opened 2025-08-18
More updates to the DD examples

reidsunderland:
Based on Sandrine's suggestions


petersilva:
Maybe these examples should now all be updated to use both upstream data pumps, instead of just dd, so people get automatically highest reliability.

reidsunderland:
I agree it might be a good idea, but I think we should get @junhu3 and @habilinour's input first.

It's unlikely that existing users would update their configs, so it wouldn't be an issue right away, but I'm not sure if dd and hpfx can handle the load of ~all users connecting to both servers.

andreleblanc11 (approved):

PR #1502, opened 2025-10-02
Issue1189 try2 - add  nofsetattr option

petersilva:
As per discussions in #1189,  and based on the PR from that #1190.   That PR was refused, and we said we should just take one patch from that to implement *nofsetattr* option, but by the time it was ready, the problem we had had been addressed by other means.  Now it has come up again, and the old patch doesn´t apply cleanly to the current development tree.  So cherry-picked and merged just the *nofsetattr* parts of the old PR, and @andreleblanc11  has been tested it (and found some gaps.) 

I´ve run the flow tests, on my laptop, and using github actions:

* AMQP flow test: https://github.com/MetPX/sarracenia/actions/runs/18167097725
* AMQP consumer test: https://github.com/MetPX/sarracenia/actions/runs/18179597277

so it seems to work.


andreleblanc11:
Ran flakey_broker, static and dynamic flow on Ubuntu18.04 and it looks good.

andreleblanc11 (approved):

Issue #1504, opened 2025-10-09
Poll binds to wrong exchange in v3.01


andreleblanc11:
We've installed v3.01.00 on the DDSR operational nodes this week and we've noticed a new problem in poll configs. The binding to the queue are now being set wrong.

This is one of the configurations where we've noticed the problem.

```
# The config : poll/airnow_http_poll.conf

  vip my-vip
  
  post_broker amqp://AIRNOW@my-broker/
  post_exchange xs_AIRNOW_http_poll
  
  queueName q_${POST_BROKER_USER}_${PROGRAM}.${CONFIG}.${HOSTNAME}
  
  fileEvents +create,modify
  # Download station files to user_cache directory
  # PROBLEM: unknown do_poll poll_airnow_http, manual conversion required.
  callback poll.airnow
  
  nodupe_fileAgeMax 6h
  nodupe_ttl        7h
  
  #destination http://files.airnowtech.org/?prefix=airnow/today/
  pollUrl https://s3-us-west-1.amazonaws.com//files.airnowtech.org/airnow/today
  source AIRNOW
  
  # ACDC 22203 , ANL - Increase sleep time as we're getting max retry errors
  scheduled_interval 900
  
  #to DDSR.CMC,DDI.CMC,DDSR.SCIENCE,DDI.SCIENCE
  
  logLevel debug
  accept .*.dat
  acceptUnmatched False

# When we look at the `subscriptions.json` file, it shows the wrong exchange being a part of the subscriptions.

  [
      {
          "baseDir": null,
          "bindings": [
              {
                  "exchange": "xs_AIRNOW",
                  "prefix": [
                      "v02",
                      "post"
                  ],
                  "sub": [
                      "#"
                  ]
              }
          ],
          "broker": "amqp://AIRNOW@my-broker",
          "queue": {
              "auto_delete": false,
              "bind": true,
              "cleanup_needed": null,
              "declare": true,
              "durable": true,
              "expire": 604800.0,
              "name": "q_AIRNOW_poll.airnow_http_poll.my-server",
              "prefetch": 25,
              "template": "q_${POST_BROKER_USER}_${PROGRAM}.${CONFIG}.${HOSTNAME}",
              "tlsRigour": "normal"
          }
      }
  ]

# The exchange should be xs_AIRNOW_http_poll and not xs_AIRNOW
```

[...]

PR #1505, opened 2025-10-12
#1504 honor post_exchange in posting components as binding for subscr…

petersilva:
The new *Subscription* class was missing code to honour the "post_exchange" argument when running poll,post, or watch (the posting components.)


andreleblanc11:
I tested the changes on dev and the patch seems to fix the problem. 

Half of the cluster has the patch and their bindings got fixed after a stop/cleanup/start. The other half didn't have the patch and kept the same exchanges.

andreleblanc11 (approved):

PR #1508, opened 2025-10-16
Syntax refactoring

belowsky26:
This is a Python script that processes products based on routing tables and configuration files. Here's what it does step by step:

1. It loads routing patterns from command line arguments using the `Swallow_Config` class.
2. Loads sender configurations from the last argument in command line.
3. Iterates over product lines, matching each one against a predefined pattern and a sender configuration.
4. If a product matches both the routing table and the sender configuration, it adds an entry to the `resulting_dict` with the subtopic (directory path transformed into a subtopic format) as key, along with details of the matched patterns and options.
5. After processing all products, the script outputs the matching results in various formats (stdout or file).

The script uses regular expressions to match product names against routing patterns and sender configurations. If a product matches both criteria, it determines whether to accept or reject it based on the configuration and adds appropriate entries to the output.


petersilva:
Hi Ilya,

Your commit contains changes to 311 files. It is very difficult to review.  Can you reformulate into multiple commits that separate changes made for different reasons?  It looks like you ran a reformatting tool on the entire code base.  If you do that, that alone should be a PR without any other functional changes.

If you have a functional change, then it would be ideal for those changes to be in isolated commits that do only that.

Issue #1509, opened 2025-10-17
cod identity broken in watches


andreleblanc11:
Discovered while working on #1495.

The identity is not defined properly when the `cod,` field is specified in a watch. 
In the example below, `identity cod,md5` is specified in the config.

```
2025-10-17 18:45:19,253 [DEBUG] sarracenia.flow do processing 1 messages worked!
2025-10-17 18:45:19,253 [INFO] sarracenia.moth.amqp putNewMessage raw message body: version: v03 type: <class 'str'> {"pubTime": "20251017T184519.247205973", "relPath": "net/local/home/leblanca/test_issue1495/t
est234", "baseUrl": "file:/", "from_cluster": "localhost", "source": "tsource", "identity": "cod,md5", "mode": "644", "size": 37, "mtime": "20251017T184517.540113926", "atime": "20251017T184517.540113926", "con
tentType": "text/plain"}
2025-10-17 18:45:19,253 [INFO] sarracenia.moth.amqp putNewMessage raw message headers: type: <class 'dict'> value: {'topic': ['v03', 'post', 'net', 'local', 'home', 'leblanca', 'test_issue1495']}
2025-10-17 18:45:19,255 [INFO] sarracenia.flowcb.log after_post posted to amqp://tsource@localhost/,xs_tsource,v03.post.net.local.home.leblanca.test_issue1495 a file with baseUrl: file:/ relPath: net/local/home
/leblanca/test_issue1495/test234 size: 37
```

`"identity": "cod,md5"` should be `"identity": "{ 'method' : 'cod' , 'value' : md5}"`

I double checked that this problem happens on both v02 and v03 post formats and it does.

This will also likely eventually lead to `TypeError: string indices must be integers` python3 errors.

```
Traceback (most recent call last):
  File "/net/local/home/leblanca/.local/lib/python3.10/site-packages/sarracenia/instance.py", line 278, in <module>
    i.start()
  File "/net/local/home/leblanca/.local/lib/python3.10/site-packages/sarracenia/instance.py", line 269, in start
    self.running_instance.run()
  File "/net/local/home/leblanca/.local/lib/python3.10/site-packages/sarracenia/flow/__init__.py", line 602, in run
    self.filter()
  File "/net/local/home/leblanca/.local/lib/python3.10/site-packages/sarracenia/flow/__init__.py", line 1157, in filter
    self._runCallbacksWorklist('after_accept')
[...]

Issue #1495, opened 2025-09-10
sr3 does not set checksum after downloading files


reidsunderland:
I think this is a bug, but I've always been a bit unclear on how it's supposed to work.

Situation: 

- poll config polls a server where checksums are not available (e.g. SFTP, FTP, HTTP, etc.), config has identity method set to `cod something` ([e.g. the default value, `cod,sha512`](https://github.com/MetPX/sarracenia/blob/c0fc15632ba8ff5bb68f504ecde8955172cef409/sarracenia/flow/poll.py#L26).)
- sarra config downloads the files and re-advertises. Config has identity method set to something (e.g. the [default value `sha512`](https://github.com/MetPX/sarracenia/blob/c0fc15632ba8ff5bb68f504ecde8955172cef409/sarracenia/config/__init__.py#L97)).

Current behavior:

- Source (e.g. sr3 sftp poll) posts a message with identity set to `cod,sha512` (correct)
- Sarra downloads the message and posts a message with identity still set to `cod,sha512`

I think the sarra should calculate the checksum after it downloads the file, then publish the message with the checksum it calculates.

There is a little bit related to https://github.com/MetPX/sarracenia/issues/1322, at least there's some useful information in there. Identity is optional in a message. But in this case, all the messages have the identity field set to cod, but the message isn't getting updated with the calculated checksum. Otherwise the cod just gets passed along forever and it never gets calculated.

Maybe this is a different topic, but I'm also not sure what's supposed to happen if the method specified in the message is different than the setting in the downloading config. e.g. if the poll posts with `cod,sha512` but the sarra has identity set to md5sum.

andreleblanc11:
I'm surprised that the `cod` would get re-appended to a posted message from a download when `sha512` is the specified `identity`.

I think this might be the area of code that sets the `cod,` inside of the message : https://github.com/MetPX/sarracenia/blob/c0fc15632ba8ff5bb68f504ecde8955172cef409/sarracenia/flow/__init__.py#L2436-L2441

`set_sumalgo` sets the checksum algorithm -> https://github.com/MetPX/sarracenia/blob/c0fc15632ba8ff5bb68f504ecde8955172cef409/sarracenia/flow/__init__.py#L2312-L2323C41 but because `identity_method` is set to `sha512` it will get the value from the incoming message.

`set_sumstr` gets that same value and then re-adds it back to the message, so it doesn't change.

*** 

[...]

PR #1510, opened 2025-10-20
Issue1495 - Changes in checksum algorithm on download

andreleblanc11:
This PR addresses the use case initially presented in #1495 and proposes the following checksum algorithm upon download (this was tested with a watch + sarra on my test environment)

Identity in incoming message | Sarra component `identity_method`   | Behaviour during posting
-- | -- | --
cod,method | None | method checksum gets posted in the message. COD does its job.
cod,method1 | method2 | method1 checksum gets posted in the message. COD does its job.
cod,method1 | cod,method2 | method1 checksum gets posted in the message. identity_method in subscriber is ignored.
None | None | Advertise no checksum.
None | method | method checksum gets posted in the message.
None | cod,method | cod,method checksum gets posted in the message.
method1 | cod,method2 | method1 gets posted in the message. identity_method in subscriber is ignored.
method1 | method2 | method1 gets posted in the message
method | None | method gets posted in the message.

This behaviour gives priority in the checksum **from the incoming message**. If the identity is not provided in the message, we fallback to the method that is provided in the configuration.

This algorithm covers most of what @reidsunderland proposed in this comment , https://github.com/MetPX/sarracenia/issues/1495#issuecomment-3276667330. It follows the defaults he proposed for the `identityFromMessage` option. 

This PR doesn't give the flexibility with the configuration option `identityFromMessage`, but I think it should be easier to implement this option if ever we would want to later.

I also provided a patch for #1509 because I needed it for my tests.

### DIFF
--- a/sarracenia/__init__.py
+++ b/sarracenia/__init__.py
@@ -554,7 +554,10 @@ class Message(dict):
         logger.debug( f"mtime persisted, calc_method: {calc_method}" )
 
         if calc_method[:4] == 'cod,' and len(calc_method) > 2:
-            sumstr = calc_method
+            sumstr = {
+                    'method' : 'cod',
+                    'value': calc_method[4:]
+                    }
         elif calc_method in [ 'md5name', 'invalid' ]:
             xattr.persist()  # persist the mtime, at least...
             return  # no checksum needed for md5name. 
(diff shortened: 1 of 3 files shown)


petersilva (approved):
It looks like you found other occurrences where it was wrong.
Good stuff.

petersilva:
weird thing going on, is flakey_broker test is failing on 22.04 only on this branch. I don't think it's real, but it is entirely consistent, which is worrisome. same test on 24.04 passes most of the time.

PR #1512, opened 2025-10-30
Be able to rename the attachment

andreleblanc11:
If we don't use msg['new_file'] we can't rename the attached file. The source file is always used.

I tested with `filename WHATFN` with a file that had sundew headers and it worked.


petersilva (approved):

petersilva:
I think that using msg['new_file'] for renaming is the correct/expected behaviour, so this is correcting the plugin to do the normal thing.

PR #1518, opened 2025-11-12
Create plugin that can automatically compress with zstd (in Python 3.14).

endlisnis:
The zstd algorithm is WAY faster than gzip (and compresses better).

I'd like to switch to using it instead of gzip, so here's my plugin for using zstd.

It relies on compression.zstd, which only ships with Python v3.14.


reidsunderland (approved):
This looks good, thanks for your contribution :)

Issue #1523, opened 2025-11-26
FTP Idle timeout generating an error


andreleblanc11:
When we connect through an FTP client, and we have long periods where we don't re-attempt a connection, we will sometimes get a `421 Idle timeout` response from the FTP server. The reported error is harmless and the FTP connections connect as expected.

```
2025-11-26 06:10:12,093 [DEBUG] sarracenia.transfer.ftp check_is_connected sr_ftp check_is_connected
2025-11-26 06:10:12,093 [DEBUG] sarracenia.transfer.ftp close sr_ftp close
2025-11-26 06:10:12,093 [DEBUG] sarracenia.transfer.ftp close Exception details:
Traceback (most recent call last):
  File "/local/home/sarra/.local/lib/python3.6/site-packages/sarracenia/transfer/ftp.py", line 192, in close
    old_ftp.quit()
  File "/usr/lib/python3.6/ftplib.py", line 672, in quit
    resp = self.voidcmd('QUIT')
  File "/usr/lib/python3.6/ftplib.py", line 280, in voidcmd
    return self.voidresp()
  File "/usr/lib/python3.6/ftplib.py", line 253, in voidresp
    resp = self.getresp()
  File "/usr/lib/python3.6/ftplib.py", line 246, in getresp
    raise error_temp(resp)
ftplib.error_temp: 421 Idle timeout (600 seconds): closing control connection
```

This seems to be happening because on the last message of a batch, **we are not attempting a `close` of the FTP connection**.

`check_is_connected` is the only method we try to `close` a connection through the FTP transfer protocol and this only gets executed before a download

https://github.com/MetPX/sarracenia/blob/b9063d8349648d6ae53ac97713bc2478548a00fe/sarracenia/flow/__init__.py#L2576-L2577

***

I have a patch ready and have tested it. The error message stops popping up.

https://github.com/MetPX/sarracenia/commit/bcf64d78763f876b0e8ec9ee2153055faf8f2c7f

I think part of the problem though is that we don't try to close the connection after the batch is complete? Maybe I'm not understanding something correctly though. I've added the `refactor` tag for this.

reidsunderland:
I think the fix is good, but it comes across something that's a bit weird with the way the batch reconnection is implemented. 

The way it is now, we always have a connection open to the server, even if it's sitting there not being used for a long time. And if the server terminates the connection, we don't notice until the next time we try to use it. 

[...]

PR #1524, opened 2025-11-27
Add fix for #1523

andreleblanc11:
The patch complies with ftplib's best practices. Execute a `close()`  after an exception is raised from `quit()`.

https://docs.python.org/3/library/ftplib.html#ftplib.FTP.quit


andreleblanc11:
I'm just realising I put the wrong issue number in the branch and the commit...

reidsunderland (approved):

PR #1526, opened 2025-12-05
Issue1525 fix AMQP not obeying tlsRigour

petersilva:
The functionality to have message queuing protocols was already present in the MQTT driver.
The AMQP protocol code, which is over 10 years old, was from a time when the *ssl* argument to the routine creating a connection accepted only a boolean, I think... or maybe it was just ok to give a boolean, and I (or Michel?) didn't bother at the time.

Anyways, nowadays, the AMQP routine accepts a TLS Context as an argument, so we can provide that similarly to how it is done for MQTT.  It is so similar, that we just move the code from mqtt.py to the parent moth class, and call it from both.

There's a weirdness where in AMQP they want to call the SSL routine *tls_set_context(s...* on the context, so if it is done before, AMQP barfs.  MQTT needed a little extra to keep the call in for it (hence the override in the child class.)

Anyways, it should do what is expected.

### DIFF
--- a/sarracenia/diskqueue.py
+++ b/sarracenia/diskqueue.py
@@ -140,10 +140,11 @@ class DiskQueue():
             self.new_fp = open(self.new_path, 'a')
 
         for message in message_list:
-            logger.debug("DEBUG add to new file %s %s" %
-                         (os.path.basename(self.new_path), message))
+            #logger.debug("DEBUG add to new file %s %s" %
+            #             (os.path.basename(self.new_path), message))
             self.new_fp.write(self.msgToJSON(message))
             self.msg_count_new += 1
+            logger.debug( f"DEBUG new file {os.path.basename(self.new_path)} +1 = {self.msg_count_new}" )
         self.new_fp.flush()
 
     def cleanup(self):
@@ -433,7 +434,7 @@ class DiskQueue():
                 fp, message = self.msg_get_from_file(fp, self.new_path)
                 if not message: break
                 i = i + 1
-                logger.debug("DEBUG message %s" % message)
+                #logger.debug("DEBUG message %s" % message)
                 if not self.needs_requeuing(message): continue
 
                 #logger.debug("MG DEBUG flush retry to state %s" % message)
--- a/sarracenia/moth/__init__.py
+++ b/sarracenia/moth/__init__.py
@@ -1,6 +1,7 @@
 import copy
 import json
 import logging
+import ssl
 import sys
 import time
 from sarracenia.featuredetection import features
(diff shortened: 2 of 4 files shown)


reidsunderland:
I tried testing it and couldn't get it to work. I set up a self-signed cert on a broker and:

On Ubuntu 18.04, sr3 connects with no complaints and `tlsRigour strict/normal/lax` makes no difference.

On Ubuntu 22.04, sr3 refuses to connect due to the self-signed certificate and `tlsRigout strict/normal/lax` also makes no difference.

I read the docs and code a bit:

- https://docs.celeryq.dev/projects/amqp/en/latest/reference/amqp.connection.html#module-amqp.connection
- https://docs.celeryq.dev/projects/amqp/en/latest/reference/amqp.transport.html#amqp.transport.SSLTransport
- https://github.com/celery/py-amqp/blob/8ad060140b5a337d83d12f57dd9f7cef43344cad/amqp/transport.py#L359

I couldn't find a way to pass in our own context object. It seems like the AMQP library always creates a context itself, and we can customize it by passing a dictionary, but I'm confused by the documentation. I think I'd need to experiment with some code to figure out how it actually works...

reidsunderland:
I think the reason it still works when given an SSLContext object is because as long as the `ssl` argument evaluates to True, it will use an SSLTransport with default options. 

```python
transport = SSLTransport if ssl else TCPTransport
```

https://github.com/celery/py-amqp/blob/8ad060140b5a337d83d12f57dd9f7cef43344cad/amqp/transport.py#L678

petersilva:
I cloned the py-amqp source and found this:

> class SSLTransport(_AbstractTransport):
>     """Transport that works over SSL.
> 
>     PARAMETERS:
>         host: str
> 
>             Broker address in format ``HOSTNAME:PORT``.
> 
>         connect_timeout: int
> 
>             Timeout of creating new connection.
> 
>         ssl: bool|dict
> 
>             parameters of TLS subsystem.
>                 - when ``ssl`` is not dictionary, defaults of TLS are used
>                 - otherwise:
>                     - if ``ssl`` dictionary contains ``context`` key,
>                       :attr:`~SSLTransport._wrap_context` is used for wrapping
>                       socket. ``context`` is a dictionary passed to
>                       :attr:`~SSLTransport._wrap_context` as context parameter.
>                       All others items from ``ssl`` argument are passed as
>                       ``sslopts``.
>                     - if ``ssl`` dictionary does not contain ``context`` key,
>                       :attr:`~SSLTransport._wrap_socket_sni` is used for
>                       wrapping socket. All items in ``ssl`` argument are
>                       passed to :attr:`~SSLTransport._wrap_socket_sni` as
>                       parameters.

So it looks like we should be passing::

     ssl={ 'context': something }

but the something is another dict of some kind? unclear.

petersilva:
This branch does not and cannot work.  Withdrawing.

Issue #1528, opened 2025-12-10
Set a CPU threshold option (similar to memory threshold)


andreleblanc11:
It would be nice to have an option to set a CPU threshold. Similar to what is already set up for the memory threshold.

https://github.com/MetPX/sarracenia/blob/dae320370e69999d5dbd9fea180d2cd366195c8c/sarracenia/flowcb/housekeeping/resources.py#L72-L99

This would help us mitigate the current impacts of [Issue 1035 - cpu time used in sr3 is higher than v2 for equivalent configuration.](https://github.com/MetPX/sarracenia/issues/1035). This gives us an alternative option to having a temporary cron that periodically restarts sr3 due to this issue.

We would likely want this new option to be turned off by default. 

The implementation also looks pretty easy. @reidsunderland pointed out the CPU usage metric might be infinitely cumulative? That would need to be tested.

petersilva:
After the first PR was removed, because it didn't restart... a mechanism that could work would be to have the instance exit.   The next run of *sr3 sanity* would then restart it.   It would probably need something to indicate an abnormal exit, like leaving the pid file around... so it's still not just a normal exit... but could work.

PR #1529, opened 2025-12-12
Add CPU time threshold for #1528

andreleblanc11:
* Also add most recent add_option option types in some documentation pages.
( I didn't bother with the french translation. It was already un-translated before)

***

I added the CPU threshold. It tracks the system CPU time of the process, and will restart it if `CpuTimeMax` is reached.
If `CpuTimeMax` is omitted, the restart logic gets also omitted.

I tried to take leverage of the psutils library and use [cpu_percent](https://psutil.readthedocs.io/en/latest/#psutil.Process.cpu_percent) during my initial testings. However, I couldn't get it to output any value other then 0.0%.

```
>>> while True:
...     print(p.cpu_percent(interval=0.1))
...     for S in range(0,10000000):
...             S += 0.1
...
0.0
0.0
0.0
0.0
0.0
0.0
0.0
0.0
0.0
0.0
0.0
0.0
0.0
0.0
0.0
0.0
^C
```

I think my tests were valid. The CPU usage of the process was logging ~90% from a `ps aux` run. 

Tracking the CPU times I think is still good. The processes that are more CPU intensive will get restarted quicker. The tricky part might be to get the right value to set as a maximum on our affected clusters.

### DIFF
--- a/docs/source/Explanation/SarraPluginDev.rst
+++ b/docs/source/Explanation/SarraPluginDev.rst
@@ -177,14 +177,23 @@ on their __init__() routines::
       """
            options can be declared in any plugin. There are various *kind* of options, where the declared type modifies the parsing.
            
-           'count'      integer count type. 
+           'count'      integer count type.
+           'octal'      base-8 (octal) integer type.
            'duration'   a floating point number indicating a quantity of seconds (0.001 is 1 milisecond)
-                        modified by a unit suffix ( m-minute, h-hour, w-week ) 
+                        modified by a unit suffix ( m-minute, h-hour, w-week )
            'flag'       boolean (True/False) option.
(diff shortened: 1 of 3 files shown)


petersilva:
the cpu time calculation is wrong... you need:

cpu_time = ost.system + ost.user

ost.system is when the process is running in the kernel (a system call.)
os.user is when it is running application code (python aka outside the kernel.)
both together are the cpu time of the process.

andreleblanc11:
I realized through further testing that performing an `os.execl` won't reset the CPU time metrics. I think it's because the CPU usage metrics only get reset when the process gets restarted with a new PID. `os.execl` only replaces the exisiting program with a new one and doesn't generate a new PID.

PR #1530, opened 2025-12-12
MessageRate should be measured post filtering.

### LINKED ISSUE #1181: messageRateMax - I think we might need a separate post_messageRateMax
I tried using messageRateMax to slow down publishing in a poll, but it appears to also slow down ingestion of messages into the nodupe cache.

In a poll, I think we'd mostly want to only limit the rate messages are posted at, not ingested. So maybe we need `post_messageRateMax` and make the "plain" `messageRateMax` only for incoming messages?

petersilva:
close #1181 

As per discussion in #1181 about messageRateMax.
It was concluded that some of the difficulty was because we were applying rate to the raw input, where that rate is kind of irrelevant when dealing with very high rejection rates (which are quite common for polls in particular, or any case where there is redundant sources for the same information.)

### DIFF
--- a/docs/source/How2Guides/UPGRADING.rst
+++ b/docs/source/How2Guides/UPGRADING.rst
@@ -39,15 +39,21 @@ Installation Instructions
 git
 ---
 
+3.02.00
+-------
+
+*CHANGE*: ``messageRateMax`` now applies *after* accept/reject filtering. In previous versions, it limited the message
+rate *before* filtering.
+
 3.01.00
 -------
 
 *NOTICE*: can now subscribe to multiple brokers, and publish to multiple post_brokers with a single configuration.
 
 *CHANGE*: python API *bindings* being replaced by *subscriptions.* API is quite different, see examples.
-*Subscriptions* enables multi-queue and multi-broker support for subscribing. 
+*Subscriptions* enables multi-queue and multi-broker support for subscribing.
 
-*CHANGE*: python API post_ settings being replaced by *publishers.* need to build 
+*CHANGE*: python API ``post_`` settings being replaced by *publishers.* need to build
 sarracenia.config.publisher.Publisher structure now to publish messages.  It enables publishing
 to multiple destinations.
 
--- a/docs/source/Reference/sr3_options.7.rst
+++ b/docs/source/Reference/sr3_options.7.rst
@@ -1227,8 +1227,8 @@ messageRateMax <float> (default: 0)
 -----------------------------------
 
 if **messageRateMax** is greater than zero, the flow attempts to respect this delivery
-speed in terms of messages per second. Note that the throttle is on messages obtained or generated
-per second, prior to accept/reject filtering. the flow will sleep to limit the processing rate.
+speed in terms of messages per second. Note that the throttle is applied to the message rate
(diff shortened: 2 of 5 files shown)


reidsunderland:
I tested this and it works, but @junhu3 was hesitant about losing the pre-filtering rate limiting, and since it's pretty easy to have both I agree.

I'm thinking we keep `messageRateMax` unchanged from the current 3.01.00 behaviour (i.e. pre-filtering) because then we don't have to worry about updating existing documentation (which is very clear that it applies pre-filtering) or breaking configs. And we add a new option, `messageRateMax_postFilter`, (I'm open to other name suggestions) which would be the changed code from this PR.

I don't mind making the changes myself.

petersilva:
I guess you want to add something like messageRateMaxAccept   to differentiate it from messageRateMax ?
Do you want to adjust this branch, or make a new PR? 

UPDATE: never mind... postFilter is find... I somehow missed it on first reading... go ahead.

petersilva:
If you haven't done this... I was thinking about it... I think the _postfilter case is going to be the 99+% use case.  To me, it should be the shorter, simpler defaultish option. The current behaviour is really has only niche uses, in fact, none of them come to mind easily.

I really think messageRateMax's behaviour should change.  If you want to keep the existing behaviour available,
I would make that the new option,  ... maybe calling it messageRateMax_gather.

reidsunderland:
I haven't worked on it yet. 

My plan was to rename the existing option to `messageRateMax_preFilter` (with `messageRateMax` as a synonym) and call the new option `messageRateMax_postFilter`. So both options would be similarly long. :)

I agree that I can't really think of any use cases for pre-filtering rate limiting. Maybe to limit processes' message ingest from queues for backpressure/load balancing, but `prefetch` is probably the better way to accomplish that.

I'm okay with naming the options `messageRateMax_gather` (pre-filter, current behaviour of `messageRateMax`) and changing `messageRateMax` to post-filtering.

I'm also okay with abandoning the idea of having both and accepting the PR as-is. It would just need to be clear in the upgrading doc that the behaviour has changed, and we would have to update our configs internally. There are a few that have specific rate limits that assume limiting is pre-filtering. @junhu3 didn't like the idea of losing functionality we already had, and I agreed. But it's also pointless to have a feature that will never be used.

@junhu3 or @andreleblanc11 do you have opinions?

andreleblanc11:
I think I prefer keeping the original behavior as an option. There might be a use case in the future where it might be useful but we can't think of yet.

petersilva:
bug with seniority == holy relic == feature!   We provide options for use cases.  If we can't even think of a use case, generally, we would not provide an option, it's just a bug fix.  That's why the original PR is formulated the way it is... we were "fixing" the existing behaviour, getting it to work the way people expect.  The original problem was a violation of PoLA, that is people setting a rate, and the implementation suprising them by working differently than expected.

It might be that we need to update the documentation, but principle of least astonishment (PoLA) would lead the shorter version to be changed to reflect the most likely intent of someone who hasn't read the docs.  On that basis, the shorter option would ideally refer to the _afterFilter rate.

reidsunderland (approved):
I added documentation in the latest commits

PR #1533, opened 2025-12-21
implement tlsRigour for AMQP broker connections

### LINKED ISSUE #1525: tlsRigour is not used for AMQP broker connections
When trying to connect to a broker with a new certificate, from a server that did not have the root cert in its trusted root CA store, sr3 refused to connect:

```
2025-12-02 00:00:06,619 [ERROR] sarracenia.moth.amqp getSetup connecting to: qname, durable: True, expire: None, auto_delete=False
2025-12-02 00:00:06,620 [ERROR] sarracenia.moth.amqp getSetup failed connection to amqps://user@broker/: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self-signed certificate in certificate chain (_ssl.c:1007)
```

I added `tlsRigour lax` to the config, and it made no difference. It looks like `tlsRigour` isn't used in the AMQP code.

petersilva:
close #1525 

To honour tlsRigour,  need to construct an ssl argument for amqp Connection routine.

Second try at this problem.  Worth testing.  I tested a bit, and it worked when I faked hostnames, but now sure if it is enough to solve the entire problem.

### DIFF
--- a/sarracenia/moth/amqp.py
+++ b/sarracenia/moth/amqp.py
@@ -36,6 +36,7 @@ from sarracenia.postformat import PostFormat
 from sarracenia.moth import Moth
 from sarracenia.interruptible_sleep import interruptible_sleep
 import os
+import ssl
 
 import time
 from urllib.parse import unquote
@@ -67,6 +68,7 @@ default_options = {
     'topicPrefix': ['v03'],
     'vhost': '/',
 }
+
 
 
 class AMQP(Moth):
@@ -228,13 +230,32 @@ class AMQP(Moth):
         if broker.url.path != '/' and broker.url.path != '':
             vhost = broker.url.path.strip('/')
 
+        if broker.url.scheme[-1] == 's':
+            if self.o['tlsRigour'] == 'lax':
+                sslarg = {
+                    'cert_reqs' : ssl.CERT_NONE,
+                    'ssl_version':  ssl.PROTOCOL_TLS_CLIENT,
+                    'server_hostname': None
+                }
+            elif self.o['tlsRigour'] == 'strict':
+                sslarg = {
+                      'cert_reqs' : ssl.CERT_REQUIRED,  # aka ssl.VerifyMode
+                      'ssl_version':  ssl.PROTOCOL_TLS_CLIENT,
+                      #'ciphers': set(ssl.OP_NO_TLSv1|ssl.OP_NO_TLSv1_1),
+                      'server_hostname': broker.url.hostname
+                }
+            else:
+                sslarg = True 
+        else:
+            sslarg=False
+        
         self.connection = amqp.Connection(host=host,
                                           userid=broker.url.username,
                                           password=unquote(
                                               broker.url.password),
                                           login_method=broker.login_method,
                                           virtual_host=vhost,
-                                          ssl=(broker.url.scheme[-1] == 's'),
+                                          ssl=sslarg,
                                           client_properties={'product':'MetPX Sarracenia (sr3)',
                                                              'product_version':sarracenia.__version__,
                                                             }
--- a/sarracenia/moth/amqp.py
+++ b/sarracenia/moth/amqp.py
@@ -234,7 +234,7 @@ class AMQP(Moth):
             if self.o['tlsRigour'] == 'lax':
                 sslarg = {
                     'cert_reqs' : ssl.CERT_NONE,
-                    'ssl_version':  ssl.PROTOCOL_TLS_CLIENT,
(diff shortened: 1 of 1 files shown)


reidsunderland:
I tested with a self-signed cert. It works perfectly on Ubuntu 22 (Python 3.10, amqp 5.0.9):

- strict and normal (default) complain about the self-signed cert and refuse to connect to the broker
- lax connects successfully without any complaint

 On Ubuntu 18 (Python 3.6, amqp 2.2.2):

- ✅ `strict` refuses to connect: `sarracenia.moth.amqp getSetup failed connection to amqps://bunnymaster@edcm-dirt22-4.edc-mtl.ec.gc.ca/: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:852)`
- ❓`normal` connects without any complaint
- _`lax` refuses to connect_: `sarracenia.moth.amqp getSetup failed connection to amqps://bunnymaster@edcm-dirt......ca/: Cannot set verify_mode to CERT_NONE when check_hostname is enabled.`

I tried adding `'check_hostname': False` to the args but that caused a different error, `_wrap_socket_sni() got an unexpected keyword argument 'check_hostname'`.

I don't really think it's a big deal, Python 3.6 is obsolete and it seems like the `normal` mode works fine even with certificates that would be rejected by `strict` mode. So I'm okay with approving and merging this if nobody else has concerns.

reidsunderland:
I also tried commenting out the `'server_hostname': None` line but it didn't make any difference either

petersilva:
I don't have access to 18.04... if someone has some time here are some things somebody can check:

in the lax scheme:

```
+            if self.o['tlsRigour'] == 'lax':
+                sslarg = {
+                    'cert_reqs' : ssl.CERT_NONE,
+                    'ssl_version':  ssl.PROTOCOL_TLS_CLIENT,
+                    'server_hostname': None
+               }

```
try adding:
```
                         'check_hostname': False,
```

It might work... If not then we need some idea of what they were expecting back then...
to get some documentation, maybe go on an 18.04 machine, and do:
```

python
import amqp
help(amqp.Connection)   # in the long text it should say look for help via the following:
help(amqp.transport.SSLTransport)

```

Maybe it will say what the fields are in the old version of py-amqp?

reidsunderland:
I tried adding `'check_hostname': False` but it gave a different error: 
`_wrap_socket_sni() got an unexpected keyword argument 'check_hostname'`

The only relevant info I see in `help(amqp.Connection)` is:

```
 |  The 'ssl' parameter may be simply True/False, or for Python >= 2.6
 |  a dictionary of options to pass to ssl.wrap_socket() such as
 |  requiring certain certificates.
```

`help(amqp.transport.SSLTransport)` doesn't address the ssl parameter, it just show that it exists in the signature and is None by default.

```
Help on function wrap_socket in module ssl:

wrap_socket(sock, keyfile=None, certfile=None, server_side=False, cert_reqs=<VerifyMode.CERT_NONE: 0>, ssl_version=<_SSLMethod.PROTOCOL_TLS: 2>, ca_certs=None, do_handshake_on_connect=True, suppress_ragged_eofs=True, ciphers=None)
```

So `cert_reqs`, `ssl_version` are both supported, but `server_hostname` and `check_hostname` aren't present.

petersilva:
I read a bit... can you replace PROTOCOL_TLS_CLIENT, with just PROTOCOL_TLS   on ubuntu 18.04? 
```
+                sslarg = {
+                    'cert_reqs' : ssl.CERT_NONE,
+                    'ssl_version':  ssl.PROTOCOL_TLS_CLIENT,
+                    'server_hostname': None
+                }
```
I was reading that PROTOCOL_TLS_CLIENT *enables CERT_REQUIRED and check_hostname by default.*  ... the older PROTOCOL_TLS doesn't (but is deprecated in newer OS.)

reidsunderland:
PROTOCOL_TLS works!

petersilva:
so if you have PROTOCOL_TLS on modern OS, does it still work? do we just change that for everyone?
Or do we need that for older versions and _CLIENT for the newer ones?

reidsunderland:
I tried PROTOCOL_TLS on Ubuntu 22.04/Python 3.10 and it worked fine there as well. I guess it would be okay to use it everywhere, and hopefully by the time it gets removed from Python we won't be using/supporting Python 3.6 anymore

petersilva:
okey dokey...  I added the patch, please bless if happy.

reidsunderland (approved):
The latest changes are confirmed to work on old Python (3.6) and new.

PR #1534, opened 2025-12-21
Issue1532 add rate to transfer logs for tools and humans

### LINKED ISSUE #1532: Add transfer speed to log messages
Sundew logs the file transfer speed for every delivered file. It would be nice to have the same thing in sarra. Currently we log the number of bytes sent, so it should be relatively easy to measure how long the transfer took and the byte rate.

It would be nice to have this for file downloads too, but I think it would be slightly harder to implement. The log message for sent files is here: https://github.com/MetPX/sarracenia/blob/dabd5ed3cccfbb14c90c9eabfe9228ef095ac3fe/sarracenia/flow/__init__.py#L2918-L2920

```
# Example Sundew log 
[INFO] (1305 Bytes) File /var/spool/px/txq/client/1/file delivered to ftp://destination/file (lat=1.240935,speed=4134.048267)

# Current sr3 Log
[INFO] sarracenia.flow send Sent: /apps/sarra/public_data/source_file  into /dest/file 0-13246380

# Desired Log
[INFO] sarracenia.flow send Sent: /apps/sarra/public_data/source_file  into /dest/file 0-13246380 time=1000 speed=13246.38
```

We are currently troubleshooting network issues, and it was super easy to generate a graph from the Sundew logs:

Image

petersilva:
close #1532
add rate to transfer logging.  See issue for discussion.

### DIFF
--- a/sarracenia/flow/__init__.py
+++ b/sarracenia/flow/__init__.py
@@ -1937,6 +1937,7 @@ class Flow:
 
         for msg in self.worklist.incoming:
 
+            start_time=time.perf_counter()
             if 'newname' in msg:
                 """
                   revamped rename algorithm requires only 1 message, ignore newname.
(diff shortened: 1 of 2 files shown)


petersilva:
note that the raw rate number is in the report field.  If people start routing reports, the rates should make it back to the publisher for analytics.  Can see which clients have problems.

reidsunderland (approved):
I tested and it works. I found and fixed a small bug, `relPath` was used instead of `msg['relPath']`.

I also added the size to the after_work messages

Issue #1537, opened 2025-12-30
sr3 reset? Have a command that does a sr3 stop/cleanup/start


andreleblanc11:
I find myself often doing a `sr3 stop component/config && sr3 cleanup component/config && sr3 start component/config` when either migrating configurations from v2 to sr3 or when debugging code/testing configurations.

It would be nice to have a command that would include all of these steps into one. Maybe we could call it `sr3 reset`? Am I the only one that would find this useful?

`--dangerWillRobinson` would need to apply for multi-configuration operations, like for `sr3 cleanup` individually.

petersilva:
*reset* sounds good initially, but  is probably too generic a choice... people will get confused between cleanup and reset... and wonder why it's running after having done a reset. probably want a less generic word.  *clean_restart* is much clearer.

PR #1540, opened 2026-01-06
Add new CLI option: clean-restart #1537

andreleblanc11:
* Add a new CLI option to do a `sr3 stop/cleanup/start` called `clean-restart`.
* I separated the `dangerWillRobinson` checks into its own method. `dangerWillRobinson` should be validated at the beginning of the `clean-restart` execution, otherwise the `stop` and `start` portion of the code will still execute.
* I added the reference to the new CLI option in the documentation.


petersilva (approved):
I ran it and it works.

PR #1547, opened 2026-01-13
documentation only, reflect change in versioning scheme

### LINKED ISSUE #1544: Are we there yet? Should the stable version convention be used again for sr3 ?
So the other projects in MetPX, and also in v2, there is a versioning convention:

*  v0  refers to MetPX Sundew  (current release:  v0.22.11 )
* v2  refers to Sarracenia prior to the refactoring started in 2020.  (current release: v2.24.08post2 )
* v3  post refactor.   (current release 3.01 )

In all the stable development trees, the versioning scheme for stable releases has been here:

https://metpx.github.io/sarracenia/Contribution/Release.html#versioning-scheme

Since 2016, when sarra v2 initially became stable, it was called  2.16.01a01.  After the first decimal, the  set of digits refers to the year of publication, and the third set of numbers is the month.  
It's a similar versioning scheme to that used by ubuntu.
3.01 introduced the multi-subscriber/multi-posting functionality which brought in some incompatibilities with what came before.  This should be the last time that happens for
a long time.  It looks like future changes are going to build on what's already done. There should not be any more compatibility breakage within the sr3. 

 Sr3 has now reached a bit of stability, and perhaps it should use the stable versioning scheme.   

To do that, the next release of sr3 should likely be:  3.26.01  or 3.26.02 

relevant: #139

petersilva:
close #1544

### DIFF
--- a/docs/source/Contribution/Release.rst
+++ b/docs/source/Contribution/Release.rst
@@ -368,7 +368,32 @@ Once the above are done, the pre-release process can proceed.
 Versioning Scheme
 ~~~~~~~~~~~~~~~~~
 
-Each release will be versioned as ``<version>.<YY>.<MM> <segment>``
+After Sr3 v3.01, the MetPX project has adopted semantic versioning (based on ideas like
+https://packaging.python.org/en/latest/specifications/version-specifiers/ )
+Versions are described there as having major.minor.micro components.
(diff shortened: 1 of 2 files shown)


reidsunderland (approved):
that looks good to me, thanks

Issue #1501, opened 2025-09-26
Ack timeout vulnerability


andreleblanc11:
This is somewhat of a duplicate to this PR : https://github.com/MetPX/sarracenia/pull/1430

I've ran into this issue the same way Peter described in the pull request above.

> Somebody had a complaint about the AMQP session dropping during long transfers... because it was idle while the transfers were going on.

We've configured a sr3 sender that is transferring multi-gigabyte files over SFTP. The batch is set to the default value, 25. When enough messages arrive in this queue, and the total transfer times exceed 30 minutes, the queue stops ack'ing messages due to the **Acknowledgment timeout value being exceeded**. When the queue reaches 25 messages, this causes an infinite loop and no messages get ack'ed.

```
2025-09-26 14:19:58,722 [WARNING] sarracenia.moth.amqp getNewMessage failed q_feeder_sender.config-name.host: (0, 0): (406) PRECONDITION_FAILED - delivery acknowledgement on channel
2 timed out. Timeout value used: 1800000 ms. This timeout value can be configured, see consumers doc guide to learn more
```

The default value for the ack. timeout is set to 30 minutes.

We can work around this by setting the `batch` value to 1. However, there's still the low risk that transferring a big enough file will exceed the ack timeout. 

We can work around this by :
- Increasing the global broker policy for the ack (consumer) timeout -> https://www.rabbitmq.com/docs/consumers#per-node-configuration
- Adding a per queue policy -> https://www.rabbitmq.com/docs/consumers#per-queue-configuration. This is only supported in RMQ 3.12 and higher however.

***

Myself and @reidsunderland still think that the ack behaviour is correct, even though it introduces this vulnerability.

A potential improvement could be to add a sarracenia option that enables ack'ing during the batch processing, and then add @petersilva's logic with the condition that the option needs to be enabled.. we're not sure if this is worth while or if this just complicates an already complicated logic though.

petersilva:
maybe and *ackEarly* ... or *acknowledgeQuickly* with an explanation of the trade off... a paragraph like:

*acknowledgeQuickly*, when set, will acknowledge messages to upstream brokers as soon as a single file's download has completed. This setting involves more frequent message protocol connection active during large batches of transfers, making timeouts less likely,  may slightly reduce message broker load (fewer messages awaiting ack), and slightly increase performance.  There is a risk of processing node failure between reception and retransmission (before the rest of the batch is processed.) so recovery from failure may have more complex failure modes.

[...]

PR #1548, opened 2026-01-14
Implement `amqp_queue_args`

reidsunderland:
We needed a way to increase the consumer timeout. This was discussed in #1501 and we decided the best way was to implement a generic way to set any AMQP queue argument. This should be more future proof than adding extra sr3 options each time we need to use a new AMQP option

### DIFF
--- a/docs/source/Reference/sr3_options.7.rst
+++ b/docs/source/Reference/sr3_options.7.rst
@@ -427,6 +427,44 @@ If the check of the upstream server fails, or the retrieve itself has failed,
 then it puts the resource on the retry queue for later attempts.
 
 
+amqp_queue_args <list> (default: not set)
+-----------------------------------------
+
+*Only applies to subscriptions to AMQP brokers.*
+
+This option is used to define optional queue arguments. The arguments that can be
+defined depend on the broker software and version used. For
+example, `RabbitMQ <https://www.rabbitmq.com/docs/queues#optional-arguments>`_.
+
+Human-friendly values (e.g. ``1m`` for 1-minute) are **not supported**. You must
+use the unit expected by the broker.
+
+Note: any queue arguments that are already defined by other sr3 options that are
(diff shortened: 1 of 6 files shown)


andreleblanc11:
I tested on 2 Ubuntu systems (U18.04 and U22.04) and on 2 RabbiMQ brokers (3.13.7 and 3.9.27) and it seems to work. 

I tested with 2 added options and both were inherited by the queue when I created them.

```
amqp_queue_args x-consumer-timeout=3600000
amqp_queue_args x-expires=5400000


rabbitmqctl list_queues name messages consumers arguments | grep x-consumer-timeout
    q_anonymous_subscribe.dd_radar.my-host.AndreLeBlancTest  6       0       [{"x-expires",5400000},{"x-consumer-timeout",3600000}]



cat ~/.cache/sr3/subscribe/dd_radar/subscriptions.json

        "broker": "amqps://anonymous@dd.weather.gc.ca/",
        "queue": {
            "amqp_queue_args": {
                "x-consumer-timeout": 3600000,
                "x-expires": 5400000
            },
            "auto_delete": false,
            "bind": true,
            "cleanup_needed": null,
            "declare": true,
            "durable": true,
            "expire": 600.0,
            "name": "q_anonymous_subscribe.dd_radar.my-host.AndreLeBlancTest",
            "prefetch": 25,
            "template": "q_${BROKER_USER}_${PROGRAM}.${CONFIG}.${HOSTNAME}.AndreLeBlancTest",
            "tlsRigour": "normal"
        }

```

On the `3.9.27` RabbitMQ broker, I don't think `x-consumer-timeout` did anything as it was only added as an option in RabbitMQ 3.12.

andreleblanc11 (approved):

reidsunderland:
Yeah, I noticed that I could define any argument and RabbitMQ wouldn't complain. I put `x-fake` and it showed up in the RabbitMQ GUI, but obviously didn't do anything.

I don't really think there's anything we can do about that though.

PR #1551, opened 2026-01-16
stop faster by queuing rest of batch for retry.

### LINKED ISSUE #1531: Option to stop a config without killing (or extend amount of time before SIGKILL is sent)
In some cases, for example when shutting down a cluster for patching or when changing the settings of a config, it would be nice to have the option to stop configs without killing processes. i.e. send SIGTERM and wait forever, rather than waiting a while then sending SIGKILL.

I think the easiest implementation would be adding a command line argument that lets us increase the number of [attempts ](https://github.com/MetPX/sarracenia/blob/dabd5ed3cccfbb14c90c9eabfe9228ef095ac3fe/sarracenia/sr.py#L2475) before SIGKILL is sent.

In most cases, this would allow the process to finish sending its current batch of files before shutting down. 

Semi-related: https://github.com/MetPX/sarracenia/issues/618 - that issue is mainly talking about not abandoning *retries* when scaling down instances.

petersilva:
For #1531, after a single transfer has completed, check if the flow has been stopped. In that case, put the rest of the messages in the retry queue.

Prior to this patch, if you have batch=50, and you are transferring the 30th file in the batch, the flow would only stop after completing the other 20 file transfers.  If the other end is down, then it will need to wait for 20 more connection attempts to timeout, which could take many, many minutes.

With this patch, the file actually in transit is completed, but the remaining 20 files get appended to the retry queue for transfer after restart.

This should make stop take effect much more quickly.

Does not completely fix #1531, but should help sr3 stop more quickly.

### DIFF
--- a/sarracenia/flow/__init__.py
+++ b/sarracenia/flow/__init__.py
@@ -1911,6 +1911,11 @@ class Flow:
                     logger.info( f"since hard link failed, fall back to copying from source" )
                     return False # fall through to download code
 
+            if self._stop_requested:
+                logger.info( f"stop requested deferring fileOp {msg['fileOp']} -> {new_path}" )
+                self.worklist.failed.append(msg)
+                return True
+
             attempt += 1
 
         # once attempts is exceeded, queue for retry later by putting the msg into worklist.failed
@@ -1955,6 +1960,11 @@ class Flow:
 
             new_path = msg['new_dir'] + os.path.sep + msg['new_file']
             new_file = msg['new_file']
+
+            if self._stop_requested:
+                logger.info(f"stop requested, deferring transfer of {new_file}")
+                self.worklist.failed.append(msg)
+                continue
 
             if not os.path.isdir(msg['new_dir']):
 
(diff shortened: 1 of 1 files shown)


reidsunderland:
We talked about this one in the meeting and we think there are too many drawbacks.

1. when removing a node from the cluster: messages are stuck in the retry queue until the instance is started again, potentially causing delays.
    - Currently, a process will be SIGKILLed if it takes too long to stop. When the process gets killed, RabbitMQ detects that the messages have not been acked yet and re-queues them, allowing another node to process them instead. This avoids the risk of delays.
2. You have to check the logs to see if transfers were deferred, which would be confusing for people who don't know this function exists. (even if it's documented, it's unlikely that people will find that information before they encounter this behavior)
3. If we were to update our node-removal procedures to check the retry queues, we wouldn't know which messages are in the retry queue because they actually failed and need to be retried, or if they were just put in there when shutting down. (we could find a way to deal with that, but I'm not sure that the extra complexity in both the code and procedures would be worth it)
4. It will likely cause problems when using the vip setting. If the node has the VIP when the process is shut down, but doesn't have the VIP when it starts back up again, the messages will be stuck in the retry queue until the node gets the VIP again.

PR #1553, opened 2026-01-20
Add new statefile for OOM restarting use case #1439

andreleblanc11:
We previously discussed in issue #1439 that we could potentially use the already `starting` statefile to prevent sanity from interfering with the OOM restart. However, I decided to create a new statefile : `oom_restarting` because,

- `os.execl` will only re-execute the command executed from the command line (i.e. `/usr/bin/python3 /path/to/lib/python3.10/site-packages/sarracenia/instance.py --no 1 start component/config`). When we create a new statefile, we need to remove it after the process has started and the only way to do that with the command line executable is from `instance.py`. 
- My thought is that it will be easier for future analysts to understand why we are removing a statefile from `instance.py` when it's referencing directly to the OOM restart. Otherwise, it might cause confusion.

### DIFF
--- a/sarracenia/flowcb/housekeeping/resources.py
+++ b/sarracenia/flowcb/housekeeping/resources.py
@@ -34,7 +34,7 @@ Returns:
 
 import logging
 
-import os, socket
+import os, socket, time, random
 from sarracenia.flowcb import FlowCB
 from sarracenia import naturalSize, naturalTime, user_cache_dir, nowstr
 from sarracenia.featuredetection import features
@@ -58,6 +58,7 @@ class Resources(FlowCB):
         ''' Per-process maximum memory footprint that is considered too large, forcing a process restart.'''
         self.transferCount = 0
         self.msgCount = 0
+        self.randomSleep = 1 + round(random.random(), 2)
 
     def on_housekeeping(self):
         if features['process']['present']:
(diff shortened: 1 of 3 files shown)


reidsunderland:
I haven't tested it yet, but I have a couple suggestions for changes that might be worth making:

- I think `self.o.cfg_run_dir` can be used to get the correct `~/.cache/...` dir
- Maybe it would be better to use a more generic filename like `resources_restart` or something? Other thresholds (open files?) might be added to the plugin in the future. I also can't remember if we've completely given up on CPU usage monitoring through the plugin or not

andreleblanc11:
Those suggestions sound good to me. 

> I also can't remember if we've completely given up on CPU usage monitoring through the plugin or not

I haven't quite given up yet on #1528. I'm a bit hesitant on having the CPU restart being dependent on `sr3 sanity`, but I can't think of a better alternative. It'd probably be good to add some `WARNING` logLevel too just before the process shuts down, so at least then people that aren't aware of the sanity dependency will have an indication of what to do to restart the process.

andreleblanc11:
Re-requesting a review.

reidsunderland (approved):
It looks good to me :)

reidsunderland (changes requested):
actually, sorry, on second thought, there might be race conditions when there are multiple instances. The statefile probably needs to use the instance number as well (`self.o.no` iirc)

andreleblanc11:
> actually, sorry, on second thought, there might be race conditions when there are multiple instances. The statefile probably needs to use the instance number as well (`self.o.no` iirc)

We've discussed this and determined that a good solution to avoid the race condition would be to add a guard and check if the state file doesn't exist before creating it, and then wait a certain amount of time if it does exist.

reidsunderland (approved):

PR #1560, opened 2026-02-20
Add documentation improvements to release process post 3.02.00rc1

andreleblanc11:
I removed the `| head -3` from the sr3_cpost command because it doesn't change anything when I tried it out on Ubuntu systems.

More `git pulls` and `git branch` are always good

Some typo corrections

### DIFF
--- a/docs/source/Contribution/Release.rst
+++ b/docs/source/Contribution/Release.rst
@@ -34,7 +34,7 @@ To publish a pre-release one needs to:
 - validate that the correct version of C stack will be running when running flow tests.
   on each server::
 
-      sr3_cpost -h
+      sr3_cpost -h |& head -3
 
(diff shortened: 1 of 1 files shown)


petersilva:
fwiw ...  

sr3_cpost |& head -3 

It was just missing the & after the sr3_cpost... this makes it print the first 3 lines of (stderr) output, so you only get the version number, as opposed some multi-page help text. but changing it does not hurt.

petersilva (changes requested):
suggest changing the first line so it adds the & instead of deleting the head -3.
It's easier to read when the user actually runs it.

andreleblanc11:
Thanks. I knew the command was probably just missing something, but didn't bother checking.

petersilva (approved):

PR #1564, opened 2026-03-06
When message is missing the identity field, calculate onfly based on setting.

### LINKED ISSUE #1563: sr3 crashes when msg does not have identity field, and identity in config is not cod, something
https://github.com/MetPX/sarracenia/blob/ca63ba80f3f84e3ed780b69d33da1be777845f3b/sarracenia/flow/__init__.py#L1406-L1411

```
  File "/net/local/home/sunderlandr/sr3/sarracenia/instance.py", line 283, in <module>
    i.start()
  File "/net/local/home/sunderlandr/sr3/sarracenia/instance.py", line 274, in start
    self.running_instance.run()
  File "/net/local/home/sunderlandr/sr3/sarracenia/flow/__init__.py", line 606, in run
    self.work()
  File "/net/local/home/sunderlandr/sr3/sarracenia/flow/__init__.py", line 1283, in work
    self.do()
  File "/net/local/home/sunderlandr/sr3/sarracenia/flow/__init__.py", line 1223, in do
    self.do_download()
  File "/net/local/home/sunderlandr/sr3/sarracenia/flow/__init__.py", line 2075, in do_download
    if self.write_inline_file(msg):
  File "/net/local/home/sunderlandr/sr3/sarracenia/flow/__init__.py", line 1408, in write_inline_file
    elif msg['identity']['method'] == 'cod':
KeyError: 'identity'
```

According to the docs, identity is optional in the message, so it really shouldn't crash when this is encountered. There should be a check to see if identity exists in the message first.

Workaround: set `identity cod,sha512` (or md5, etc.) in the config.

petersilva:
close #1563

default algo_method when identity field missing in message

### DIFF
--- a/sarracenia/flow/__init__.py
+++ b/sarracenia/flow/__init__.py
@@ -1405,6 +1405,8 @@ class Flow:
 
         if self.o.identity_method.startswith('cod,'):
             algo_method = self.o.identity_method[4:]
+        elif 'identity' not in msg:
+            algo_method = self.o.identity_method
         elif msg['identity']['method'] == 'cod':
             algo_method = msg['identity']['value']
         else:


reidsunderland (approved):
Thanks :)

Issue #1573, opened 2026-03-12
When filtering messages with sundew extensions that come from sources with a port in the URL, the sundew extension is not added to the urlToFilter


reidsunderland:
https://github.com/MetPX/sarracenia/blob/a60f3636a520a5ba9d23ef2d58d60a9f76c82dfd/sarracenia/flow/__init__.py#L1042-L1048

https://github.com/MetPX/sarracenia/blob/a60f3636a520a5ba9d23ef2d58d60a9f76c82dfd/sarracenia/flow/__init__.py#L1074-L1077

If the source URL has a port specified, e.g.: `[DEBUG] sarracenia.flow filter  urlToMatch: http://dms-dev1.domain:8180/data/msc/forecast/atmospheric/aviation/`

The count of `:` in the URL is 2, and the Sundew extension doesn't get added. 

This isn't a critical issue, as far as I know nobody has ever noticed it. I only came across it while reading the code trying to figure out if `rename` fields were supposed to get added to the `urlToMatch` or not.

reidsunderland:
Actually, this happens all the time, because if you have a scheme:// in the URL, `url.count(':')` will always be at least one

reidsunderland:
It seems like v2 had the same issue: 

https://github.com/MetPX/sarracenia/blob/4b19896921460971049d43c8674d85ab78a89c7e/sarra/sr_subscribe.py#L981-L985

I think `if 'sundew_extension' in m and url.count(":") < 1: ` could just be changed to: `if 'sundew_extension' in m and m['sundew_extension'] not in url: `

I'll test that later...

(EDIT: v2 only looks at the URL **path** so it's okay)

reidsunderland:
I do think it's worth fixing, because the behaviour now could be surprising. Sometimes the Sundew extension is part of the relPath and not a separate field/header in the message, and in that case, the filename **with** sundew extension is used for filtering.

reidsunderland:
I noted this in: https://github.com/MetPX/sarracenia/pull/1574

But changing this will likely break existing accept statements.

Example:

```
accept .*\.txt
```

The user would expect that that would accept all .txt files.

But if a file comes in with:
```
relPath: path/to/file.txt , sundew_extension: some:sundew:extension:example:test
```

The urlToMatch becomes: `path/to/file.txt:some:sundew:extension:example:test` and the accept doesn't work.

So now I'm thinking it might be best to remove this code, mark this wontfix and document it.

[...]

PR #1574, opened 2026-03-13
tests: regression tests for sundew_extension with port in URL (#1573) ;)

robjarawan:
I noticed there were no tests for this, so I tried it out and added some:) bug: when a source URL has a port like http://someserver:8180/..., the sundew extension doesn't get appended to urlToMatch bc the port's colon was being counted. The fix is already in the code -  these tests just make sure it stays that way.

### DIFF
--- a/tests/sarracenia/flow/__flow___test.py
+++ b/tests/sarracenia/flow/__flow___test.py
@@ -1,6 +1,85 @@
 import pytest
 from tests.conftest import *
-#from unittest.mock import Mock
+
+import urllib.parse
 
 import sarracenia.config
 import sarracenia.flow
+import copy
+
+
+def _build_urlToMatch(m):
+    """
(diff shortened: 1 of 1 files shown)


reidsunderland:
This is great :) I'll wait to merge this until the actual code is changed though, and I'm starting to have second thoughts about that. If the way the sundew extension is processed changes, it could break existing configs. 

e.g. message has `relPath: path/to/file.txt` and `sundew_extension: some:extension:abc:123:idk`, and if the config has `accept .*txt` with this change, the urlToMatch becomes `path/to/file.txt:some:extension:abc:123:idk`. This would break the accept statement.

reidsunderland:
The filtering fix got merged.  I came back to look at this, but it re-implemented the logic inside the test code, so it doesn't reflect the filtering code in `flow/__init__.py`.

This is a PR from a fork, so I had to use a different branch. Closing, replaced by #1664.

Issue #1567, opened 2026-03-06
Compute Identity fails when `mtime` is missing in a message


andreleblanc11:
I think this is a bug. `mtime` is stated to be an optional parameter in the documentation : https://metpx.github.io/sarracenia/Reference/sr_post.7.html#:~:text=%22mtime%22%20%3A%20date%20string%20%2D%20last%20modification%20time%20of%20a%20file%20(optional).

When computeIdentity is called and `mtime` is missing in a message, the file gets retried.

```
2026-03-06 05:55:37,169 [INFO] convert.imv6_to_png download converting file.imv6 to file.png
2026-03-06 05:55:37,445 [INFO] convert.imv6_to_png download Exception details:
Traceback (most recent call last):
  File "/local/home/sarra/.config/sr3/plugins/convert/imv6_to_png.py", line 121, in download
    msg.computeIdentity(opath, self.o)
  File "/usr/lib/python3/dist-packages/sarracenia/__init__.py", line 542, in computeIdentity
    if xattr.get('mtime') >= msg['mtime']:
KeyError: 'mtime'
...
2026-03-06 05:55:38,033 [ERROR] sarracenia.flow do_download gave up downloading for now, appending to retry queu
```

I think this should be a pretty easy fix in the `computeIdentity` code.

PR #1575, opened 2026-03-13
tests: regression test for computeIdentity with missing mtime (#1567)

robjarawan:
No tests for this one either, so I can play some more:) bug: computeIdentity() crashed with `KeyError: 'mtime'` when a msg didnt have a `mtime` field even tho mtime is documented as optional. Code is already fixed this is just a reg test..

### DIFF
--- a/tests/sarracenia/__init___test.py
+++ b/tests/sarracenia/__init___test.py
@@ -252,6 +252,31 @@ class Test_Message():
         except:
             pass
 
+    def test_computeIdentity_without_mtime(self, tmp_path):
+        """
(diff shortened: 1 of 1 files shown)


andreleblanc11 (changes requested):
I think the added unit test is good. However, I see the git diffs shows the whole file has changed (possibly tabs replaced by spaces? possibly also identation corrections)?

Can you resubmit the changes but just with your changes and without the space/tabs changes? :)

andreleblanc11 (approved):

PR #1577, opened 2026-03-14
handle pid file race condition during sr3 restart

### LINKED ISSUE #1571: Potential PID file race condition causes `sr3 restart` to crash
I first noticed this about 2 weeks ago when I was running my auto-restart script on our stage servers. I wasn't able to catch the error from the stderr of my script output, but I was able to catch it today when I did a manual restart from the same stage server. 

What seems to happen:
1. `sr3 restart` gets invoked
2. At least one process is detected in a `missing` state. 
3. `_clean_missing_proc_state_dir` gets invoked to cleanup the directory of the missing process. Somehow, there's some kind of a race condition where the `*pid` file goes missing after doing the `os.listdir()` in the code.

Error message 
```
sr3 restart
stopping: sending SIGTERM ................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................... ( 515 ) Done
Waiting 1 sec. to check if 515 processes stopped (try: 0)
  File "/usr/bin/sr3", line 11, in <module>
	load_entry_point('metpx-sr3==3.2.0rc2', 'console_scripts', 'sr3')()
  File "/usr/lib/python3/dist-packages/sarracenia/sr.py", line 3447, in main
	gs.stop()
  File "/usr/lib/python3/dist-packages/sarracenia/sr.py", line 2495, in stop
	self._clean_missing_proc_state()
  File "/usr/lib/python3/dist-packages/sarracenia/sr.py", line 711, in _clean_missing_proc_state
	self._clean_missing_proc_state_dir(self.user_cache_dir)
  File "/usr/lib/python3/dist-packages/sarracenia/sr.py", line 695, in _clean_missing_proc_state_dir
	t = p.read_text().strip()
  File "/usr/lib/python3.6/pathlib.py", line 1196, in read_text
	with self.open(mode='r', encoding=encoding, errors=errors) as f:
  File "/usr/lib/python3.6/pathlib.py", line 1183, in open
	opener=self._opener)
  File "/usr/lib/python3.6/pathlib.py", line 1037, in _opener
	return self._accessor.open(self, flags, mode)
  File "/usr/lib/python3.6/pathlib.py", line 387, in wrapped
	return strfunc(str(pathobj), *args)
FileNotFoundError: [Errno 2] No such file or directory: 'sarra_get_copernicus_01.pid'
```

After the `sr3 restart` script fails, **all of the processes are in a `stop` state** ☹️ 
Doing a `sr3 start` afterwards resumes the processes to their normal operations.

I tried a bunch of ways to recreate the problem, all of which did not work.
1. Kill a process manually from its PID number, i.e. `kill $PID#` and doing a `sr3 restart` afterwards.
2. Remove the `*.pid` file from a process from `~/.cache/sr3/component/config/` and doing a `sr3 restart` afterwards.
3. Rebooting the server and doing a `sr3 restart` after it has come back up. (I only tried this, because this is the same scenario where the problem got triggered today).

robjarawan:
close #1571

pid file can disappear between os.listdir() and read_text() causing a FileNotFoundError crash. Added try/except in both _find_missing_instances_dir and _clean_missing_proc_state_dir. Added regression tests.

### DIFF
--- a/sarracenia/sr.py
+++ b/sarracenia/sr.py
@@ -641,12 +641,17 @@ class sr_GlobalState:
                                 i_found.append(i)
                                 if i != 0:
                                     p = pathlib.Path(filename)
-                                    if sys.version_info[0] > 3 or sys.version_info[
-                                            1] > 4:
-                                        t = p.read_text().strip()
-                                    else:
-                                        with p.open() as f:
-                                            t = f.read().strip()
+                                    try:
+                                        if sys.version_info[0] > 3 or sys.version_info[
+                                                1] > 4:
+                                            t = p.read_text().strip()
+                                        else:
+                                            with p.open() as f:
+                                                t = f.read().strip()
+                                    except FileNotFoundError:
+                                        logger.debug("pid file %s disappeared (race condition), treating as missing", filename)
+                                        missing.append([c, cfg, i])
+                                        continue
                                     if t.isdigit():
                                         pid = int(t)
                                         if pid not in self.procs:
@@ -690,12 +695,16 @@ class sr_GlobalState:
                         for filename in os.listdir():
                             if filename[-4:] == '.pid':
                                 p = pathlib.Path(filename)
-                                if sys.version_info[0] > 3 or sys.version_info[
-                                        1] > 4:
-                                    t = p.read_text().strip()
-                                else:
-                                    with p.open() as f:
-                                        t = f.read().strip()
+                                try:
+                                    if sys.version_info[0] > 3 or sys.version_info[
+                                            1] > 4:
+                                        t = p.read_text().strip()
+                                    else:
+                                        with p.open() as f:
+                                            t = f.read().strip()
+                                except FileNotFoundError:
+                                    logger.debug("pid file %s disappeared (race condition), skipping", filename)
+                                    continue
                                 if t.isdigit():
                                     pid = int(t)
                                     if pid not in self.procs:
--- a/tests/sarracenia/sr_test.py
+++ b/tests/sarracenia/sr_test.py
@@ -1,6 +1,120 @@
 import pytest
 from tests.conftest import *
-#from unittest.mock import Mock
+from unittest.mock import Mock, patch
 
+import os
+import pathlib
 import sarracenia.config
 import sarracenia.sr
+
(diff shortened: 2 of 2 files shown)


reidsunderland (changes requested):
A missing instance is an instance where the PID file exists and the process is not running. So if the PID file disappears, the instance can't be missing. So in the except case, the `missing.append([c, cfg, i])` can be removed

Unrelated, but while you're working on this code, the pidfile reading code with the if statement is obsolete. The `if`/`else` can be removed and `t = p.read_text().strip()` can be used without an if since that function was added in Python 3.5 and we don't support any Python versions older than 3.6.

sarracenia/sr.py
```diff
@@ -641,12 +641,17 @@ def _find_missing_instances_dir(self, dir, statehost=False):
...
-                                            t = f.read().strip()
+                                    try:
+                                        if sys.version_info[0] > 3 or sys.version_info[
+                                                1] > 4:
+                                            t = p.read_text().strip()
+                                        else:
+                                            with p.open() as f:
+                                                t = f.read().strip()
+                                    except FileNotFoundError:
+                                        logger.debug("pid file %s disappeared (race condition), treating as missing", filename)
```
andreleblanc11, line 652: I think these logger entries should be an ERROR message. Would also be good to reference the GH issue as well.

andreleblanc11 (changes requested):

andreleblanc11:
I agree with what Reid said. I also think the fix is good and probably the best we can achieve now given how rare and irreproducable the problem is.

andreleblanc11:
I can't checkout your branch on our test environment because your PR references your forked repo:  [robjarawan:fix/1571-pid-race](https://github.com/robjarawan/sarracenia/tree/fix/1571-pid-race). Can you create a new branch on the sarracenia repository and add the same changes, but with our recommendations? :smile:

robjarawan:
> I can't checkout your branch on our test environment because your PR references your forked repo: [robjarawan:fix/1571-pid-race](https://github.com/robjarawan/sarracenia/tree/fix/1571-pid-race). Can you create a new branch on the sarracenia repository and add the same changes, but with our recommendations? 😄

🫡 I will tonight

reidsunderland:
Superseded by #1595 , closed

PR #1579, opened 2026-03-14
Issue1572 *exchange* optional for non amqp 0.9, topicPrefix None  

### LINKED ISSUE #1572: Only one `topic` is allowed in a subscription
Because MQTT doesn't have the same exchange+binding model as AMQP0.9.1, we build topic bindings that replicate this: `exchange_name/topic_prefix+subtopic`. 

But when subscribing to non-Sarracenia MQTT brokers that use different formats (e.g. WIS2 brokers), this can be problematic. sr3 currently doesn't allow None/empty for exchange or topicPrefix.

But we have a `topic` option that will override exchange/topicPrefix/subtopic:

https://github.com/MetPX/sarracenia/blob/a60f3636a520a5ba9d23ef2d58d60a9f76c82dfd/sarracenia/moth/mqtt.py#L243-L252

This is very useful, but it's a string option so only one topic can be defined. I think we should change it to a list option. I'm mainly creating this issue to ask if there's any reason it's a string variable and shouldn't be a list?

I want to re-use `topic` for AMQP1.0, which is why I'm bringing it up.

---

Not exactly related to this issue, but there's a public WIS2 MQTT broker that we can use for testing: https://training.wis2box.wis.wmo.int/practical-sessions/connecting-to-wis2-over-mqtt/

```
broker mqtts://everyone:everyone@globalbroker.meteo.fr:8883/

tlsRigour lax

#exchange     origin
#topicPrefix  a
#subtopic     #

topic origin/a/wis2/#

set sarracenia.moth.mqtt.MQTT.logLevel debug
messageDebugDump True

directory /tmp/wis2/
```

petersilva:
close #1572 

## Update bindings to accept *topic* in addition to *subtopic*.

parsing of *topic*, as described in the issue, is added to the config/subscription.py parsing,
so now multiple topics can be specified.  *topic* is now explicitly an override that avoids
sr3 topic construction using many settings.  It can be used to debug problems with the 
topic construction built into config/subscription.py logic.

UPDATE adding this, is from the a commit added to the PR.
UPDATE topic is now explicitly documented as an override, and users are pointed at subtopic
for most usage.

## make *exchange None* work

* *Exchanges* only exist in amqp < 1.0.  They are a protocol idiosyncratic concept.  but they can help organize topic hierarchies in other protocols by providing a common root.
* make *exchange* a thing only used by the moth/amqp driver (for rabbitmq aka amqp 0.9)
* the *exchange* setting is now not used for mqtt or any protocol other than amqp 0.9
* Make *exchange None* a valid setting.

UPDATE: later commit changed name of *mqttExchangeBeforeTopicPrefix* to *topicExchangePrepend*
UPDATE: later commit replaced the new option with *exchange None* (and *post_exchange* None)

This makes mqtt subscription logic put the exchange in the topic hierarchy using the mapping used
until now. Use this logic for all non AMQP 0.9 (aka url != amqp(s)://.... ) brokers. If the option is turned off, then exchange is ignored, and people more versant in MQTT can set the topic more explicitly.

UPDATE: apply MQTT logic to all non AMQP0.9 brokers.

## make *topicPrefix None* work

If you specify topicPrefix None (or post_topicPrefix None) it does what people expect.

So when working with non sr3 interop, and different brokers, *exchange* is not needed,
and topicPrefix can be set to None.


petersilva:
Oh, I confirmed it basically works, but not thorough at all.

petersilva:
I'm wondering if the *mqttExchangeBeforeTopicPrefix* option should have a different name, so that it can be used in all protocols that do not have a native concept of *exchange* ... maybe it should be named. *topicExchangePrepend* instead, and used to change topic calculation in more cases.  It is likely a good concept for any protocol that supports hierarchical topics, but needs to be optional for cases where constraints make it unusable.

petersilva:
OK. I found some issues doing some custom testing with *topicExchangePrepend* off.  I had found so cases where it produced some exceptions while running.  those are all fixed, and it seems to do the right thing.   I'm done now, so it's worth reviewing.

reidsunderland:
So far this is looking good. 

It seems like it's possible to use both topics and subtopics in a config:

```
'subscriptions': [{'baseDir': None,
                    'bindings': [{'prefix': ['origin', 'my_topic_prefix'], 'sub': ['#']}, {'topic': ['origin', 'origin/a/wis2/#']}, {'topic': ['origin', 'fdsafdsa/fdsa/fdsa/#']}],
```

```
exchange     origin
topicPrefix  my_topic_prefix
subtopic     #

topic origin/a/wis2/#
topic fdsafdsa/fdsa/fdsa/#
```

That seems reasonable to me, I can't think of a good reason to not allow it.

petersilva:
fwiw... I am still working on this.  I have a patch that removes the topicExchangePrepend option, and everything seems to work for *exchange None*  but post_exchange is still getting assigned wrong... wait a bit.

petersilva:
The thing I committed passes all flow tests (amqp and mqtt), and *exchange None* seems to work.  Still testing posting... I found one issue... but still testing.

petersilva:
rebased from development... but it makes it harder to review.

petersilva:
OK, all done.  passes all the tests the old ones used to and now accepts:

* exchange None
* post_exchange None
* topicPrefix None
* post_topicPrefix None
* topics now work seamlessly with subtopics.

This way of doing it, as suggested by @reidsunderland, seem much more intuitive than the extra flag.

reidsunderland:
I'm getting a crash running any sr3 commands:

```
  File "/net/local/home/sunderlandr/.local/bin/sr3", line 8, in <module>
    sys.exit(main())
  File "/net/local/home/sunderlandr/sr3/sarracenia/sr.py", line 3388, in main
  File "/net/local/home/sunderlandr/sr3/sarracenia/sr.py", line 1350, in __init__
    self._resolve()
  File "/net/local/home/sunderlandr/sr3/sarracenia/sr.py", line 868, in _resolve
    self._resolve_brokers()
  File "/net/local/home/sunderlandr/sr3/sarracenia/sr.py", line 819, in _resolve_brokers
    for b in s['bindings']:
KeyError: 'bindings'
```

I'm taking a look to see why

reidsunderland:
Also seeing:

```
  File "/net/local/home/sunderlandr/.local/bin/sr3", line 8, in <module>
    sys.exit(main())
  File "/net/local/home/sunderlandr/sr3/sarracenia/sr.py", line 3392, in main
  File "/net/local/home/sunderlandr/sr3/sarracenia/sr.py", line 1348, in __init__
    self._read_configs()
  File "/net/local/home/sunderlandr/sr3/sarracenia/sr.py", line 339, in _read_configs
    cfgbody.parse_file(cfg,c)
  File "/net/local/home/sunderlandr/sr3/sarracenia/config/__init__.py", line 1545, in parse_file
    self.parse_line( component, cfg, cfname, lineno, l.strip() )
  File "/net/local/home/sunderlandr/sr3/sarracenia/config/__init__.py", line 1672, in parse_line
    self._parse_binding(v, k in ['topic'] )
  File "/net/local/home/sunderlandr/sr3/sarracenia/config/__init__.py", line 1393, in _parse_binding
    self.subscriptions.add(Subscription(self, self.queueName, resolved_queueName, subtopic, topicOverride))
  File "/net/local/home/sunderlandr/sr3/sarracenia/config/subscription.py", line 136, in add
    newb = new_subscription['bindings'][0]
KeyError: 'bindings'
```

It seems to be something with configs where topic/subtopic are missing. But it may be isolated to `amq1(s)` URLs.

petersilva:
what does the config look like... ie settings: exchange, prefix, subtopic or topic...

petersilva:
if there is no subtopic or topic... then there will be no bindings... .... I think that's the problem... why aren't your specifying that? oh... if everything is specified on the other side, so you don't bind... hmm...

petersilva:
ok I found it.

reidsunderland:
I think in AMQP and MQTT there are default values, which is why I'm only having problems with AMQP1.0

For AMQP1.0 we need at least 1 (sub)topic (i.e. address) for a subscriber to work, but I can't think of a sensible default.

petersilva:
I found it, here: 

https://github.com/MetPX/sarracenia/blob/5927eabfa9045489a07be8521ab38c13d34c57bb/sarracenia/config/subscription.py#L43-L55
need an else... for amq... hmm...

petersilva:
do you have an exchange setting in your config? I think exchange != None combined with amq1 broker might be a case not handled.  Perhaps remove the 'p' on line 51 to use the exchange in amqp 1.0 even though it isn't defined?  or don't insist on mqtt in the following clause... or have an additional else clause.

What are the settings in your config... in order, for setting up the binding?

reidsunderland:
### No Exchange

```
broker amq1://anonymous:anonymous@dd.weather.gc.ca
topic hello
```

works, gives `'bindings': [{'topic': ['xpublic', 'hello']}]`

```
broker amq1://anonymous:anonymous@dd.weather.gc.ca
subtopic hello
```

crashes in `_resolve_brokers`

(exchange defaults to xpublic, which is fine)

### Explicit Actual Exchange

```
broker amq1://anonymous:anonymous@dd.weather.gc.ca
exchange hi
topic hello
```

works, gives `'bindings': [{'topic': ['hi', 'hello']}]`

```
broker amq1://anonymous:anonymous@dd.weather.gc.ca
exchange hi
subtopic hello
```

crashes in ` _resolve_brokers`

### Explicit `None` Exchange

```
broker amq1://anonymous:anonymous@dd.weather.gc.ca
exchange None
topic hello
```

works, gives `'bindings': [{'topic': ['hello']}]`

```
broker amq1://anonymous:anonymous@dd.weather.gc.ca
exchange None
subtopic hello
```

works, gives `'bindings': [{'prefix': ['v02', 'post'], 'sub': ['hello']}]`

---

So I think the combination that is breaking is **non-** None exchange used in combination with `subtopic`

petersilva:
> I think in AMQP and MQTT there are default values, which is why I'm only having problems with AMQP1.0
> 
> For AMQP1.0 we need at least 1 (sub)topic (i.e. address) for a subscriber to work, but I can't think of a sensible default.

If it's a mandatory setting that's ok, we just should not crash incomprehensibly.  it sounds like you would have something like:
```

topicPrefix None
exchange None
topic mandatory_address

```

If you have something like that... topicOverride=True... 
so the binding should be set to [ { 'topic': mandatory_address }]

reidsunderland:
I think you found the right section of code in [this](https://github.com/MetPX/sarracenia/pull/1579#issuecomment-4100568783) comment above and adding an elif for AMQP1 (or just a general else) would be the right fix

petersilva:
> So I think the combination that is breaking is **non-** None exchange used in combination with `subtopic`

OK... but I don't know what to do with an exchange outside amqp 0.9... what would be a correct binding?
```
a)   [ { 'exchange': 'hi' , 'subtopic': 'hello' } ] 
```
or 
```
b) [ [  {  'subtopic': [ 'hi', 'hello' ] } ]  or equivalently 'hi.hello' in a string.
```
or somthing else?

petersilva:
After the last commits:

* *topic* is now explicitly an override for rare use, and doesn´t interact at all with exchange or topicPrefix.
* the config/subscription module now builds the entire subscription topic, so that you can see exactly what it will do when you do *sr3 show*
* As a consequence of the above, the field in the binding is now always *topic* because the queueName, prefix and subtopics are all resolved when the subscription config object is inited.
* because of the change in subscription formats, always resolving to a complete topic, could remove the logic in the message queue protocols, because now we just hand them the literal calculated topics.
* The subscriptions.json files stored by older versions will have different fields, so added some logic to convert the old ones to the current format.  the bindings were converted, and the tlsRigour is now added to queue settings when reading.

I think it should be a lot easier to understand what it is doing now, because *sr3 show* will show the real text that will be fed to the message protocol.  It also helps as much as reasonable, but gives flexibility to work around it when you need it.

petersilva:
something is broken, I found some problems in the flow tests.  Don't see anyway to revert to draft.

petersilva:
All better now.  should be mergeable.

reidsunderland:
I rebased to resolve merge conflicts

reidsunderland:
I think there is something wrong with reading the cached subscription.

If you have a config, start it, stop it, edit the (sub)topics, then run sr3 show, it will show both the bindings in the config and in the cache (even if the cached one has been removed from the config).

It's not just an issue with sr3 show, it happens when running the config too.

petersilva:
I think that's the same behaviour now in sr3 without any mods... unless you do a cleanup, it doesn't clean out the old bindings...

petersilva:
The stuff to compare bindings, and do deltas so that editing of bindings would work properly is future work.
There is some place holders in the code for it, but don't expect editing of bindings to work.  It did not before either, this is not a regression.

but maybe I don't understand what your are doing.

for clarity:
* there is still no way to cleanly modify bindings... To get that to work, we need code that does a fancy "diff" of bindings and figures out which bindings to remove on the broker  and which ones to add.
* the only way to do that cleanly is to *cleanup* the existing bindings, then edit, then declare new ones.  or... you can add the new ones with sr3, and go into the broker management interface and remove the offending old ones (obviously not useful in many, many cases.)


here's the issue about that: #35

reidsunderland:
This is the expected behaviour, that I see on the dev branch:

1. any config can be used (I will test with example `subscribe/dd_aqhi.conf`)
2. config has `subtopic *.WXO-DD.air_quality.aqhi.#`
3. run `sr3 show subscribe/dd_aqhi`:
    - correct binding is shown: `'bindings': [{'exchange': 'xpublic', 'prefix': ['v02', 'post'], 'sub': ['*.WXO-DD.air_quality.aqhi.#']}],`
4. start the process `sr3 start subscribe/dd_aqhi`, log shows that the correct binding is created:
    - `2026-03-25 16:27:15,313 [INFO] 2840383 sarracenia.moth.amqp getSetup binding q_anonymous.subscribe.dd_aqhi.sunderlandr_edcm-dirt_35176279 with v02.post.*.WXO-DD.air_quality.aqhi.# to xpublic (as: amqps://anonymous@dd.weather.gc.ca/)`
5. stop the subscriber (do not cleanup)
6. edit config, change binding to `subtopic *.HELLO`
7. run sr3 show again, as expected, only **one** binding is shown, and it reflects what is in the config:
    - `'bindings': [{'exchange': 'xpublic', 'prefix': ['v02', 'post'], 'sub': ['*.HELLO']}],`
8. start the config again, it will create only the binding that is in the config file:
    - `2026-03-25 16:30:21,366 [INFO] 2841060 sarracenia.moth.amqp getSetup binding q_anonymous.subscribe.dd_aqhi.sunderlandr_edcm-dirt_35176279 with v02.post.*.HELLO to xpublic (as: amqps://anonymous@dd.weather.gc.ca/)`
    - yes, the queue will retain the old binding, but that's a known different issue

<br><br>

now run `sr3 cleanup subscribe/dd_aqhi` to reset for the next test (double check the subscription file in the cache is actually deleted)

<br><br>

Repeating the test with branch `issue1572_pas1`:

1. config has `subtopic *.WXO-DD.air_quality.aqhi.#`
2. run `sr3 show subscribe/dd_aqhi`:
    - correct binding is shown: `'bindings': [{'exchange': 'xpublic', 'prefix': ['v02', 'post'], 'sub': ['*.WXO-DD.air_quality.aqhi.#']}],`
3. start the process `sr3 start subscribe/dd_aqhi`, log shows that the correct binding is created:
    - `2026-03-25 16:36:11,547 [INFO] 2841483 sarracenia.moth.amqp getSetup binding q_anonymous.subscribe.dd_aqhi.sunderlandr_edcm-dirt_43358769 with v02.post.*.WXO-DD.air_quality.aqhi.# to xpublic (as: amqps://anonymous@dd.weather.gc.ca/)`
4. stop the subscriber (do not cleanup)
5. edit config, change binding to `subtopic *.HELLO`
6. ❌ run sr3 show again, this time, **two** bindings are shown. The binding from the subscriptions cache file is re-used and the correct binding from the config is also present:
    - `'bindings': [{'exchange': 'xpublic', 'topic': 'v02.post.*.WXO-DD.air_quality.aqhi.#'}, {'exchange': 'xpublic', 'topic': 'v02.post.*.HELLO'}],`
7. ❌ if you start the config, it will (re-)create **both** bindings, not just the one from the config as expected (i.e. this problem is not just isolated to sr3 show):
    ```
    2026-03-25 16:38:17,394 [INFO] 2841589 sarracenia.moth.amqp getSetup binding q_anonymous.subscribe.dd_aqhi.sunderlandr_edcm-dirt_43358769 with v02.post.*.WXO-DD.air_quality.aqhi.# to xpublic (as: amqps://anonymous@dd.weather.gc.ca/)
    2026-03-25 16:38:17,396 [INFO] 2841589 sarracenia.moth.amqp getSetup binding q_anonymous.subscribe.dd_aqhi.sunderlandr_edcm-dirt_43358769 with v02.post.*.HELLO to xpublic (as: amqps://anonymous@dd.weather.gc.ca/)
    ```

petersilva:
OK... looking.  Maybe this is a chance to get the statefulness finally done.

petersilva:
@reidsunderland  try it now?  I think it's fixed.... one liner...

petersilva:
If there is a problem, here is what I'm worried it might be:
* sr3 show
* sr3 declare

will the show and the declare match?   I'm worried about things like queue names not being propagated from old_subscription to current.  I think it's ok, but did not have time to test.  worth examining...

petersilva:
I think it's fine... when it searches ... the list includes current and old_subscriptions, and if it finds the old one, it uses that... so I don't think queue names get messed up ever.

reidsunderland (approved):
the problem I had before is fixed. It seems to work as expected.

reidsunderland:
I fixed merge conflicts with https://github.com/MetPX/sarracenia/pull/1579/changes/dc2c6f6f2132bd2f81fbf25c7146f40633654688 and hopefully didn't screw anything up

(when I click the link in this comment, it gives me a huge diff, but if I click the link above in the timeline, it only shows the conflict fixes)

PR #1580, opened 2026-03-15
Add httpUserAgent config option

### LINKED ISSUE #1477: Customizable HTTP user agent (suffix?)
This is @robjarawan's idea.

It would be nice to be able to customize the HTTP user agent.

https://github.com/MetPX/sarracenia/blob/dc3b0f539b2f6741a01b90c494eece9a7ab86c01/sarracenia/transfer/https.py#L352

I personally think it would be best to only support a suffix that can be added to the existing `Sarracenia $version` user agent.

robjarawan:
##### Summary
- Adds `httpUserAgent` string option to append a custom suffix to the default `Sarracenia <version>` user-agent header in HTTP(S) requests
- User-agent string is built once in `connect()` rather than reconstructed on every request
- CRLF characters are stripped from user input to prevent HTTP header injection (defense-in-depth)

Closes #1477

##### Test plan
- [ ] Set `httpUserAgent MyApp/1.0` in config, verify header sent
- [ ] Verify default behavior unchanged when option is not set

### DIFF
--- a/sarracenia/config/__init__.py
+++ b/sarracenia/config/__init__.py
@@ -916,7 +916,7 @@ class Config:
         self.log_flowcb_needed = False
         self.sleep = 0.1
         self.housekeeping = 300
-        self.httpUserAgent = None
+        self.httpUserAgent = 'Sarracenia ' + sarracenia.__version__
         self.inline = False
         self.inlineByteMax = 4096
         self.inlineEncoding = 'guess'
--- a/sarracenia/config/__init__.py
+++ b/sarracenia/config/__init__.py
@@ -191,7 +191,7 @@ str_options = [
     'accelCpCommand', 'accelWgetCommand', 'accelScpCommand',
     'action', 'admin', 'baseDir', 'broker', 'cluster', 'directory', 'exchange',
     'exchangeSuffix', 'feeder', 'filename', 'flatten', 'flowMain', 'header', 
-    'hostname', 'httpsSafeQuote', 'identity', 'inlineEncoding', 'logFormat', 'logLevel',
+    'hostname', 'httpsSafeQuote', 'httpUserAgent', 'identity', 'inlineEncoding', 'logFormat', 'logLevel',
     'pollUrl', 'post_baseUrl', 'post_baseDir', 'post_broker', 'post_exchange',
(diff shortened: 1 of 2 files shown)


petersilva:
It looks fine, just a little overcomplicated.

The reason for the separate self.user_agent in the class was because there was no global config item to use.
If you do have a global value, then you don't need two values, and copying between them.  Just use what's available.
Set the value in the config to the default value, and then always use that value, instead of checking for None value, it just gets overridden if user um... overrides it. So you don't need the *if* at all. and you just always use the option value without modification. I mean in config/__init__.py in default_options ... around line 94

```python

     'httpUserAgent': 'Sarracenia ' + sarracenia.__version__

```

the you just always use self.o.httpUserAgent... no if statements... no extra state variables (self.user_agent in https.py goes away). less code.

robjarawan:
> It looks fine, just a little overcomplicated.
> 
> The reason for the separate self.user_agent in the class was because there was no global config item to use. If you do have a global value, then you don't need two values, and copying between them. Just use what's available. Set the value in the config to the default value, and then always use that value, instead of checking for None value, it just gets overridden if user um... overrides it. So you don't need the _if_ at all. and you just always use the option value without modification. I mean in config/**init**.py in default_options ... around line 94
> 
> ```python
>      'httpUserAgent': 'Sarracenia ' + sarracenia.__version__
> ```
> 
> the you just always use self.o.httpUserAgent... no if statements... no extra state variables (self.user_agent in https.py goes away). less code.

Question, with the simplification, if a user sets `httpUserAgent` in their config it fully replaces the default (including the sarracenia version prefix doesnt it?) The original code appended the user value after the prefix. Is full replacement what we want, or should the sr version always be present?

robjarawan:
Rethinking my own question -- full replacement makes sense. If someone sets httpUserAgent they know what they want. The default already includes the version string, so anyone who doesn't set it gets the standard behavior.

The current diff already follows your suggestion -- default set in config/__init__.py, no if statements, no self.user_agent variable, just self.o.httpUserAgent used directly. Ready for another look when you get a chance.

petersilva (approved):
looks good!

PR #1582, opened 2026-03-15
Fix NameError and dead code in SFTP mkdir

robjarawan:
I found a couple of issues doing a sweep of sftp.py, here's the first.

The `mkdir` method has two bugs:

1. **NameError** -it references an undefined variable `path` instead of the parameter `remote_dir`. This would crash every time `mkdir` is called.

2. **Dead code** - the actual `mkdir`/`chmod` logic was in a `try/except/else` block's `else` clause, which only runs if the `try` succeeds with no exception. But every success path in the `try` does `return`, so the `else` never executes. The directory never actually gets created.

Moved the `mkdir`/`chmod` into the `FileNotFoundError` handler where it belongs, and narrowed the bare `except:` to `except Exception:`.

### DIFF
--- a/sarracenia/transfer/sftp.py
+++ b/sarracenia/transfer/sftp.py
@@ -457,19 +457,17 @@ class Sftp(Transfer):
         logger.debug("mkdir %s" % remote_dir)
         alarm_set(self.o.timeout)
         try:
-            s = self.sftp.lstat(path)
+            s = self.sftp.lstat(remote_dir)
             if S_ISDIR(s.st_mode):
                 return
-            logger.error( f"cannot mkdir {path}, file exists" )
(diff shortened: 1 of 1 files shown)


reidsunderland (approved):
It's weird that we've never encountered this. I guess mkdir is rarely used for SFTP (for senders, cd_forced is used instead)

Previously, all exceptions were handled inside the method. Now, an exception could occur and be thrown up to the caller if the sftp.mkdir or self.chmod error out. But that seems to be okay -- I can only find one place where mkdir is used, and it has a try/except there.

https://github.com/MetPX/sarracenia/blob/4472c80a0471beecb91cbbd4182cbf52686773f4/sarracenia/flow/__init__.py#L2743-L2747

robjarawan:
First merge 😍😍

PR #1578, opened 2026-03-14
close sftp file handles on failed transfers

robjarawan:
close #1559

sftp put() and get() leaked the remote file handle when readlocal_write() or read_writelocal() threw an exception. Each failed retry leaked a handle, eventually hitting the OS limit and crashing with "Too many open files". Wrapped both in try/finally so rfp.close() always runs. Added regression tests..

Issue #1559, opened 2026-02-20
Too many open files crash when retrying


reidsunderland:
```
2026-02-20 14:23:27,008 [ERROR] sarracenia.transfer.sftp __init__ sr_sftp/__init__: unable to load ssh config /local/home/sarra/.ssh/config
2026-02-20 14:23:27,008 [ERROR] sarracenia.transfer.sftp connect sr_sftp/connect: unable to connect to serverfqdn (user:destusername)
2026-02-20 14:23:27,008 [WARNING] sarracenia.flow do_send sending again, attempt 3
2026-02-20 14:23:27,008 [ERROR] sarracenia.transfer.sftp __init__ sr_sftp/__init__: unable to load ssh config /local/home/sarra/.ssh/config
2026-02-20 14:23:27,008 [ERROR] sarracenia.transfer.sftp connect sr_sftp/connect: unable to connect to serverfqdn (user:destusername)
2026-02-20 14:23:27,008 [ERROR] sarracenia.transfer.sftp __init__ sr_sftp/__init__: unable to load ssh config /local/home/sarra/.ssh/config
2026-02-20 14:23:27,008 [ERROR] sarracenia.transfer.sftp connect sr_sftp/connect: unable to connect to serverfqdn (user:destusername)
2026-02-20 14:23:27,008 [WARNING] sarracenia.flow do_send sending again, attempt 2
2026-02-20 14:23:27,008 [ERROR] sarracenia.transfer.sftp __init__ sr_sftp/__init__: unable to load ssh config /local/home/sarra/.ssh/config
2026-02-20 14:23:27,008 [ERROR] sarracenia.transfer.sftp connect sr_sftp/connect: unable to connect to serverfqdn (user:destusername)
2026-02-20 14:23:27,008 [WARNING] sarracenia.flow do_send sending again, attempt 3
2026-02-20 14:23:27,009 [ERROR] sarracenia.transfer.sftp __init__ sr_sftp/__init__: unable to load ssh config /local/home/sarra/.ssh/config
2026-02-20 14:23:27,009 [ERROR] sarracenia.transfer.sftp connect sr_sftp/connect: unable to connect to serverfqdn (user:destusername)
2026-02-20 14:23:27,009 [ERROR] sarracenia.transfer.sftp __init__ sr_sftp/__init__: unable to load ssh config /local/home/sarra/.ssh/config
[...]

Issue #1391, opened 2025-03-03
when v2 receives a message from sr3 with no size set, it will likely truncate the file


reidsunderland:
When v2 receives a message from sr3 where the size is not set, it seems to download a much smaller file than expected. I've observed this with HTTP at least, I'm not sure if other protocols are affected.

We already knew about this, but I don't think it's documented (I took a quick look through the open issues).

I don't think it's worth fixing. You can either use sr3 to download the file, or post the size in the message. If the size is unknown, just post (from sr3) a size much larger than the expected size and use v2 plugin `on_file_repair_size` to set the correct size after downloading.

For example, this sr3 plugin can be used to set a fixed, large size:

```python
import logging
from sarracenia.flowcb import FlowCB

logger = logging.getLogger(__name__)

class Force_size(FlowCB):

    def __init__(self, options) :
        super().__init__(options, logger)
        self.o.add_option('forceSize', kind='size', default_value="10GB")

    def after_accept(self, worklist):
        for msg in worklist.incoming:
            msg['size'] = self.o.forceSize
```

petersilva:
What does sr3 do when receiving a message with no size set?

reidsunderland:
It works correctly. It downloads the full file and (assuming a download plugin is not used), it sets the size in the message to the size of the downloaded file. 

If acceptSizeWrong is not set to true, it will also log a warning:

```
[WARNING] sarracenia.flow download downloaded 358532 of with no length given for FILENAME assuming ok

[INFO] 2354666 sarracenia.moth.amqp putNewMessage raw message body: version: v02 type: <class 'str'> 20250305172402.342253685 POST_BASEURL tmp/avalanche/FILENAME

[INFO] 2354666 sarracenia.moth.amqp putNewMessage raw message headers: type: <class 'sarracenia.Message'> value: {'atime': '20250305172353.744154453',  'mode': '666', 'mtime': '20250305172353.744154453', 'source': 'bunnymaster', 'to_clusters': 'ALL', 'contentType': 'application/octet-stream', 'parts': '1,358532,1,0,0', 'topic': ['v02', 'post', 'tmp', 'avalanche'], 'sum': 'n,17b1fbe043f48f4b22ee96cab4d90d0c'}

[INFO] 2354666 sarracenia.flowcb.log after_post posted to exchange: xs_bunnymaster topic: v02.post.tmp.avalanche a file with baseUrl: POST_BASEURL relPath: tmp/avalanche/FILENAME size: 358532

[...]

PR #1583, opened 2026-03-15
Fix SFTP file handle leaks in get/put

robjarawan:
Second one from the sftp.py sweep -- same category as #1559 and PR #1578 but in different methods.

`get()` and `put()` both open a remote file pointer (`rfp`) but don't close it if the read/write operation throws an exception. If the transfer fails mid-stream (timeout, network drop, etc.), the fd leaks. Do that enough times and you hit "too many open files."

Fixed by wrapping the read/write calls in try/finally so `rfp.close()` always runs.

Also caught that `get()` was hardcoding `exactLength=False` instead of passing the caller's value through to `read_writelocal` -- now it passes the actual parameter.

### DIFF
--- a/sarracenia/transfer/sftp.py
+++ b/sarracenia/transfer/sftp.py
@@ -372,17 +372,18 @@ class Sftp(Transfer):
 
         # read from rfp and write to local_file
 
-        rw_length = self.read_writelocal(remote_file, rfp, local_file,
-                                         local_offset, length, exactLength=False)
-
-        
-        # close
-
-        alarm_set(self.o.timeout)
         try:
-            rfp.close()
+            rw_length = self.read_writelocal(remote_file, rfp, local_file,
+                                             local_offset, length,
+                                             exactLength)
         finally:
-            alarm_cancel()
+            # close
+
+            alarm_set(self.o.timeout)
+            try:
+                rfp.close()
+            finally:
+                alarm_cancel()
 
         return rw_length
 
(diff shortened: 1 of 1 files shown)


reidsunderland:
I'm a bit worried about the exactLength change. When the size specified in the message is smaller than the actual file size at the source, v2 will download a corrupt file.

sr3 doesn't have that problem and I wonder if this difference is the reason why. Before merging this that should be looked into more (I can try to investigate it when I have more time)

reidsunderland (changes requested):
The changes look good, except the exactLength change. Please change that back to `=False`. 

There's a note in the parent class about why it's always False:

```python
        # 2022/12/02 - pas - need copies to always work...
        # in HPC mirroring case, a lot of short files, likely length is wrong in announcements.
        # grab the whole file unconditionally for now, detect error using mismatch.
```

Changing it will likely break things that are currently working.

The way it is now with False, we always transfer the entire file. If the size received doesn't match the size in the message then sr3 will log a warning. This behaviour is good, the v2 behaviour caused lots of issues.

Related: https://github.com/MetPX/sarracenia/issues/1391  https://github.com/MetPX/sarracenia/commit/088f957d3d4c46e10fe8783aa27047052099044a https://github.com/MetPX/sarracenia/commit/dcaf56e3b8fe66389fdd0c2569176eda9b63f09f

robjarawan:
exactLength reverted back to =False in 2d65469 .. good catch on the HPC mirroring history woops, that's not something I would have caught without in there and without context. Ready for another look when you get a chance.

reidsunderland (approved):
The changes look good to me now. THere's an SFTP sender on ddi that is getting alarm raise exceptions in unexpected places (amqp getNewMessage) so this should fix a legit bug.

PR #1593, opened 2026-03-17
fix: AMQP init overwrites parent defaults, losing tlsRigour

### LINKED ISSUE #1591: `sr3 declare` fails because `tlsRigour` isn't found
## How to recreate
Run `sr3 --users declare my-component/my-config`

You should get the following errors when trying to declare a new user.

```
2026-03-16 16:41:42,076 3794485 [ERROR] sarracenia.moth.amqp putSetup failed to connect or declare exchanges ['xs_MY_NEW_USER']@MY_NEW_USER on fluxi.cmc.ec.gc.ca: 'tlsRigour'
2026-03-16 16:41:42,076 3794485 [DEBUG] sarracenia.moth.amqp putSetup Exception details:
  File "/usr/lib/python3/dist-packages/sarracenia/moth/amqp.py", line 465, in putSetup
    if not self.__connect(self.o['broker']):
  File "/usr/lib/python3/dist-packages/sarracenia/moth/amqp.py", line 236, in __connect
    if self.o['tlsRigour'] == 'lax':
KeyError: 'tlsRigour'

```

I was able to work around the problem by adding a `tlsRigour : 'normal'` inside of the default_options of the amqp MOTH class.

i.e. https://github.com/MetPX/sarracenia/blob/4fb8a33041e0c6530e4c7b5f10ed7ab9b43bb28a/sarracenia/moth/amqp.py#L53-L70

I think adding that default `tlsRigour` value to all of the MOTH types might be a good enough fix.

robjarawan:
#### Summary

Fixes #1591 — `sr3 declare` fails with `KeyError: 'tlsRigour'` when connecting to an `amqps://` broker.

**Root cause:** `AMQP.__init__` overwrites `self.o` (already set by `Moth.__init__` with `tlsRigour: 'normal'`) with its own module-level `default_options` dict, which doesn't include `tlsRigour`. When `sr3 declare` passes minimal props to the AMQP constructor (just broker, exchange, dry_run, message_strategy), `tlsRigour` is never restored, and `__connect` raises `KeyError` at line 234.

**Fix:** Change `self.o = copy.deepcopy(default_options)` to `self.o.update(copy.deepcopy(default_options))`, so AMQP-specific defaults layer on top of parent Moth defaults instead of replacing them. This is the same pattern MQTT already uses (there's even a comment at `mqtt.py:126` explaining why the old overwrite pattern was wrong).

This is a one-line fix that addresses the structural issue rather than just adding `tlsRigour` to AMQP's `default_options` — any future option added to the parent `Moth` class will also be inherited correctly.

#### Test plan

- [x] Added 3 regression tests in `tests/sarracenia/moth/amqp_test.py`:
  - `tlsRigour` present when AMQP is created with minimal props (the bug scenario)
  - Explicit `tlsRigour` override in props is preserved
  - AMQP-specific defaults (vhost, exchangeDeclare, durable) still applied
- [ ] Ideally tested with `sr3 --users declare` against an amqps:// broker (I don't have one to test against)

### DIFF
--- a/sarracenia/config/publisher.py
+++ b/sarracenia/config/publisher.py
@@ -53,6 +53,9 @@ class Publisher(dict):
             logger.error("malformed publisher, missing (post_)exchange")
             return
 
+        if hasattr(options,'tlsRigour') :
+            self['tlsRigour'] = options.tlsRigour
+    
         if hasattr(options,'post_format') :
             self['format'] = options.post_format
         elif hasattr(options,'post_topicPrefix') and options.post_topicPrefix[0] in [ 'v02', 'v03' ]:
--- a/sarracenia/moth/amqp.py
+++ b/sarracenia/moth/amqp.py
@@ -184,7 +184,7 @@ class AMQP(Moth):
 
         # update self.o (already set by super().__init__) with AMQP-specific defaults,
         # then re-apply props so they take priority.
-        self.o.update(copy.deepcopy(default_options))
+        self.o.update(default_options)
         self.o.update(props)
 
         self.first_setup = True
--- a/sarracenia/moth/amqp.py
+++ b/sarracenia/moth/amqp.py
@@ -182,7 +182,9 @@ class AMQP(Moth):
             format=
             '%(asctime)s [%(levelname)s] %(name)s %(funcName)s %(message)s')
 
-        self.o = copy.deepcopy(default_options)
+        # update self.o (already set by super().__init__) with AMQP-specific defaults,
+        # then re-apply props so they take priority.
+        self.o.update(copy.deepcopy(default_options))
         self.o.update(props)
 
         self.first_setup = True
--- a/sarracenia/sr.py
+++ b/sarracenia/sr.py
@@ -1848,7 +1848,8 @@ class sr_GlobalState:
                             'echangeDeclare': False,
                             'subscription_index': 0,
                             'subscriptions' : [ s ],
-                            'message_strategy': { 'stubborn':True }
+                            'message_strategy': { 'stubborn':True },
+                            'tlsRigour': q['tlsRigour']
                         })
                     qdc.getSetup()
                     qdc.getCleanUp()
--- a/sarracenia/sr.py
+++ b/sarracenia/sr.py
@@ -1542,7 +1542,8 @@ class sr_GlobalState:
                         'broker': self.default_cfg.admin,
                         'dry_run': self.options.dry_run,
                         'exchange': self.default_cfg.declared_exchanges,
-                        'message_strategy': { 'stubborn':True }
+                        'message_strategy': { 'stubborn':True },
+                        'tlsRigour': self.options.tlsRigour
                     })
                 xdc.putSetup()
                 xdc.close()
@@ -1564,7 +1565,8 @@ class sr_GlobalState:
                                 'broker': p['broker'],
                                 'dry_run': self.options.dry_run,
                                 'exchange': p['exchange'],
-                                'message_strategy': { 'stubborn':True }
+                                'message_strategy': { 'stubborn':True },
+                                'tlsRigour': p['tlsRigour']
                             })
                          xdc.putSetup()
                          xdc.close()
@@ -1882,7 +1884,8 @@ class sr_GlobalState:
                                         'exchange': p['exchange'],
                                         'dry_run': self.options.dry_run,
                                         'broker': self.brokers[h]['admin'],
-                                        'message_strategy': { 'stubborn':True }
+                                        'message_strategy': { 'stubborn':True },
+                                        'tlsRigour': p['tlsRigour']
                                     })
                                 if qdc:
                                     qdc.putSetup()
--- a/tests/sarracenia/moth/amqp_test.py
+++ b/tests/sarracenia/moth/amqp_test.py
@@ -1,6 +1,56 @@
 import pytest
 from tests.conftest import *
-#from unittest.mock import Mock
+from unittest.mock import patch, MagicMock
 
 import sarracenia.config
+import sarracenia.moth
 import sarracenia.moth.amqp
+
+
+class Test_AMQPDefaultOptions:
+    """Test that AMQP inherits tlsRigour from parent Moth defaults.
+
+    Regression test for https://github.com/MetPX/sarracenia/issues/1591
+    When sr3 declare creates an AMQP instance with minimal props (no tlsRigour),
+    __connect would raise KeyError: 'tlsRigour'.
+    """
+
+    @patch('sarracenia.moth.amqp.amqp')
+    def test_tlsRigour_present_with_minimal_props(self, mock_amqp_lib):
+        """AMQP init with minimal props (like sr3 declare) must still have tlsRigour."""
+        minimal_props = {
+            'broker': MagicMock(),
+            'dry_run': False,
+            'exchange': 'xpublic',
+            'message_strategy': {'stubborn': True},
+        }
+        instance = sarracenia.moth.amqp.AMQP(minimal_props, is_subscriber=False)
+        assert 'tlsRigour' in instance.o
+        assert instance.o['tlsRigour'] == 'normal'
+
+    @patch('sarracenia.moth.amqp.amqp')
+    def test_tlsRigour_override_preserved(self, mock_amqp_lib):
+        """When props explicitly set tlsRigour, the value must be kept."""
+        props = {
+            'broker': MagicMock(),
+            'dry_run': False,
+            'exchange': 'xpublic',
+            'message_strategy': {'stubborn': True},
+            'tlsRigour': 'lax',
+        }
+        instance = sarracenia.moth.amqp.AMQP(props, is_subscriber=False)
+        assert instance.o['tlsRigour'] == 'lax'
+
+    @patch('sarracenia.moth.amqp.amqp')
+    def test_amqp_specific_defaults_still_applied(self, mock_amqp_lib):
+        """AMQP-specific defaults (e.g. vhost, exchangeDeclare) must still be set."""
+        minimal_props = {
+            'broker': MagicMock(),
+            'dry_run': False,
+            'exchange': 'xpublic',
+            'message_strategy': {'stubborn': True},
+        }
+        instance = sarracenia.moth.amqp.AMQP(minimal_props, is_subscriber=False)
+        assert instance.o['vhost'] == '/'
+        assert instance.o['exchangeDeclare'] is True
+        assert instance.o['durable'] is True


petersilva:
I think this is not the right idea.  The problem is not to get a default for tlsRigour, but to propagate the setting.
getting a default value available for tlsRigour prevents the crash, but demonstrates that if a value is given by the user, it is not used.  Need a patch that uses the setting.

reidsunderland:
I was unclear about this change, but it does address something in the AMQP class that isn't obvious and maybe not intentional?

The default options in the parent Moth class (`__default_options` variable) are never used in the AMQP implementation. 

`self.o` is originally set by the parent class' constructor (to `__default_options` from Moth). Then `self.o` is replaced by a deepcopy of `default_options` from AMQP, then updated to overwrite any defaults with the user's config:

https://github.com/MetPX/sarracenia/blob/2d6a10436794839aa7e4a416a3d47bb8ff49468a/sarracenia/moth/amqp.py#L185-L186


This is different from MQTT, which inherits the defaults from Moth, **updates** them with the MQTT defaults, then updates with the user's config:

https://github.com/MetPX/sarracenia/blob/2d6a10436794839aa7e4a416a3d47bb8ff49468a/sarracenia/moth/mqtt.py#L126-L130

Rob's change makes AMQP behave the same as MQTT (although the update call uses deepcopy, which isn't necessary in this case)

andreleblanc11:
Part of the reason why the `tlsRigour` isn't getting propogated by the config is because we are only using **the MOTH default options** when we are initializing through the `pubFactory`.

https://github.com/MetPX/sarracenia/blob/4fb8a33041e0c6530e4c7b5f10ed7ab9b43bb28a/sarracenia/sr.py#L1540-L1548

I was getting the same error through a `cleanup` as well when connecting to an amqps broker. The cleanup method also calls `pubFactory`.

`pubFactory` calls the targeted classes __init__ and this loads the default options. I think in this case Rob's fix makes sense.. if this is the route we want to keep going into. I think we might need to do some kind of refactor to include the configs options inside of the MOTH declaration (I didn't look at the code that much so I'm not sure).

FWIW Rob's patch fixed my problem when I tested.

petersilva:
OK... but I don't think it is the whole fix.  If you set tlsRigour to a different value, does the AMQP connection use it?

andreleblanc11:
It does (See CRITICAL log. It prints out self.o just before attempting a connection to the broker)

I put `tlsRigour strict` in the config.

```
vip             my-ip

broker          amqps://feeder@my-broker/
exchange        xpublic
queueName       q_${BROKER_USER}.${PROGRAM}.${CONFIG}.ddi-shared
tlsRigour          strict

# This include file reference all of the bulletin subtopics
include         all-msc-bulletin-subtopics.inc

nodupe_basis    data
nodupe_ttl      60m
fileAgeMax      30m

post_broker     amqps://ALL-MSC-BULLETINS@my-broker/
post_exchange   xs_ALL-MSC-BULLETINS

accept .*
```

```
sr3 --debug --users declare winnow/all-msc-bulletins
...
2026-03-18 19:38:07,582 1233928 [DEBUG] amqp collect Closed channel #1
2026-03-18 19:38:07,582 1233928 [DEBUG] amqp collect Closed channel #2
2026-03-18 19:38:07,582 1233928 [INFO] root declare looking at winnow/all-msc-bulletins
2026-03-18 19:38:07,582 1233928 [INFO] root declare looking at winnow/all-msc-bulletins
2026-03-18 19:38:07,582 1233928 [DEBUG] sarracenia.moth __init__  Maximum interval exponential back off of connecting to broker: 91.69203903006802
2026-03-18 19:38:07,582 1233928 [CRITICAL] sarracenia.moth.amqp __connect Current options: {'acceptUnmatched': True, 'batch': 25, 'broker': <sarracenia.config.credentials.Credential object at 0x7c2b73c3a050>, 'dry_run': False, 'exchange': 'xpublic', 'expire': 86400.0, 'inline': False, 'inlineEncoding': 'guess', 'inlineByteMax': 4096, 'logFormat': '%(asctime)s [%(levelname)s] %(name)s %(funcName)s %(message)s', 'logLevel': 'info', 'messageDebugDump': False, 'message_strategy': {'reset': True, 'stubborn': True, 'failure_duration': '5m'}, 'messageAgeMax': 0, 'topicPrefix': ['v02', 'post'], 'tlsRigour': 'strict', 'subscriptions': [{'broker': <sarracenia.config.credentials.Credential object at 0x7c2b73c3a050>, 'bindings': [{'exchange': 'xpublic', 'prefix': ['v02', 'post'], 'sub': ['*.*.MSC-BULLETINS.#']}, {'exchange': 'xpublic', 'prefix': ['v02', 'post'], 'sub': ['*.MSC-CMC.NAT-SCRIBE-OPS.PROPERCASE.#']}, {'exchange': 'xpublic', 'prefix': ['v02', 'post'], 'sub': ['*.MSC-SCI-CMC-OPS.cmcin-bul.#']}, {'exchange': 'xpublic', 'prefix': ['v02', 'post'], 'sub': ['*.MSC-SCI-CMC-OPS.cmcin-fd.#']}], 'queue': {'name': 'q_feeder.winnow.all-msc-bulletins.ddi-shared', 'template': 'q_${BROKER_USER}.${PROGRAM}.${CONFIG}.ddi-shared', 'cleanup_needed': None, 'bind': True, 'declare': True, 'auto_delete': False, 'durable': True, 'expire': 86400.0, 'prefetch': 25, 'tlsRigour': 'strict'}, 'baseDir': None}], 'old_subscriptions': [{'baseDir': None, 'bindings': [{'exchange': 'xpublic', 'prefix': ['v02', 'post'], 'sub': ['*.*.MSC-BULLETINS.#']}, {'exchange': 'xpublic', 'prefix': ['v02', 'post'], 'sub': ['*.MSC-CMC.NAT-SCRIBE-OPS.PROPERCASE.#']}, {'exchange': 'xpublic', 'prefix': ['v02', 'post'], 'sub': ['*.MSC-SCI-CMC-OPS.cmcin-bul.#']}, {'exchange': 'xpublic', 'prefix': ['v02', 'post'], 'sub': ['*.MSC-SCI-CMC-OPS.cmcin-fd.#']}], 'broker': <sarracenia.config.credentials.Credential object at 0x7c2b73c3a050>, 'queue': {'auto_delete': False, 'bind': True, 'cleanup_needed': None, 'declare': True, 'durable': True, 'expire': 86400.0, 'name': 'q_feeder.winnow.all-msc-bulletins.ddi-shared', 'prefetch': 25, 'template': 'q_${BROKER_USER}.${PROGRAM}.${CONFIG}.ddi-shared', 'tlsRigour': 'strict'},
```

petersilva:
That's only showing a subscription.  We already know subscriptions work.  Need to see a publisher.

andreleblanc11:
It does not.

```
grep tlsRigour sarra/cap_alert_sign.conf
tlsRigour strict


sr3 --debug --users declare sarra/cap_alert_sign
2026-03-18 22:13:42,476 1333481 [INFO] sarracenia.sr declare Declaring exchanges for admin.conf using amqps://root@broker/
2026-03-18 22:13:42,476 1333481 [DEBUG] sarracenia.moth __init__  Maximum interval exponential back off of connecting to broker: 63.74465778651932
2026-03-18 22:13:42,476 1333481 [CRITICAL] sarracenia.moth.amqp putSetup Current options: {'acceptUnmatched': True, 'batch': 25, 'broker': <sarracenia.config.credentials.Credential object at 0x742f783ddd80>, 'dry_run': False, 'exchange': ['xs_MSC-PREDICTION'], 'expire': None, 'inline': False, 'inlineEncoding': 'guess', 'inlineByteMax': 4096, 'logFormat': '%(asctime)s [%(levelname)s] %(name)s %(funcName)s %(message)s', 'logLevel': 'info', 'messageDebugDump': False, 'message_strategy': {'stubborn': True}, 'messageAgeMax': 0, 'topicPrefix': ['v03'], 'tlsRigour': 'normal', 'auto_delete': False, 'durable': True, 'exchangeDeclare': True, 'persistent': True, 'prefetch': 25, 'queueBind': True, 'queueDeclare': True, 'reset': False, 'subtopic': [], 'vhost': '/'}

```

petersilva:
yeah.  that was the point... the original problem was that the value wasn't showing up at all... by providing a default, it hides the crash, but the real solution needs to propagate the option.

andreleblanc11:
I think I may have found a fix to adding the `tlsRigour` (and possibly other options) to the publish declaration. 

Currently when we declare through the `pubFactory` we do not pass all the options.

https://github.com/MetPX/sarracenia/blob/4fb8a33041e0c6530e4c7b5f10ed7ab9b43bb28a/sarracenia/sr.py#L1550-L1570

I've changed the code so that it passes the options to the exchange declaration and that seems to have passed `tlsRigour` properly.

```
# Code change
        # declare exchanges first.
        for f in self.filtered_configurations:
            if self.please_stop:
                break
            (c, cfg) = f.split(os.sep)

            if not 'options' in self.configs[c][cfg]:
                continue
            logging.info('looking at %s/%s ' % (c, cfg))
            if hasattr(self.configs[c][cfg]['options'],'publishers'):
                for p in self.configs[c][cfg]['options'].publishers:
                     if 'exchange' in p:
                         od = o.dictify()
                         od['broker'] = p['broker']
                         od['dry_run'] = self.options.dry_run
                         od['exchange'] = p['exchange']
                         od['message_strategy'] = { 'stubborn':True }
                         xdc = sarracenia.moth.Moth.pubFactory(od)
                         xdc.putSetup()
                         xdc.close()

# sr3 --debug --users declare output (config has `tlsRigour strict`

2026-03-19 14:03:23,993 1923241 [DEBUG] sarracenia.moth __init__  Maximum interval exponential back off of connecting to broker: 107.74942034309476
2026-03-19 14:03:23,993 1923241 [CRITICAL] sarracenia.moth.amqp putSetup Current options: {'acceptUnmatched': False, 'batch': 100, 'broker': <sarracenia.config.credentials.Credential object at 0x719fe3467730>, 'dry_run': False, 'exchange': ['xpublic', 'xpublic', 'xpublic', 'xpublic', 'xpublic', 'xpublic2'], 'expire': 86400.0, 'inline': False, 'inlineEncoding': 'guess', 'inlineByteMax': 4096, 'logFormat': '%(asctime)s [%(levelname)s] %(name)s %(funcName)s %(message)s', 'logLevel': 'info', 'messageDebugDump': False, 'message_strategy': {'stubborn': True}, 'messageAgeMax': 0, 'topicPrefix': ['v02', 'post'], 'tlsRigour': 'strict', 'subscriptions': [{'broker': <sarracenia.config.credentials.Credential object at 0x719fe3336650>, 'bindings': [{'exchange': 'xpublic', 'prefix': ['v02', 'post'], 'sub': ['*.MSC-DMS-OP.CAP-WARNINGS.#']}], 'queue': {'name': 'q_feeder.sarra.cap_alert_sign.sarra_node_13407458', 'template': 'q_${BROKER_USER}.${COMPONENT}.${CONFIG}.${QUEUESHARE}', 'cleanup_needed': None, 'bind': True, 'declare': True, 'auto_delete': False, 'durable': True, 'expire': 86400.0, 'prefetch': 25, 'tlsRigour': 'strict'}, 'baseDir': None}]
```

Idk if we'd also need to do the same for the admin exchange declaration. I don't think it should hurt passing all the options to it too.

petersilva:
OK, I imported my original patch, which adds *tlsRigour* to publishers, and then added propagating *tlsRigour* to the pubFactories.  I think it might be ok now, but haven't tested.  If someone can test, that would be good.

andreleblanc11:
I tested again the pub setup and the tlsRigour gets properly propogated now with your added patches.

A note, for the fun of things, I tried reverting the `self.o.update(copy.deepcopy(default_options))` change back to `self.o = copy.deepcopy(default_options)` and added `'tlsRigour': q['tlsRigour']` to the subFactory called in the `cleanup` and that also seems to fix the `cleanup` problem.

```
                    print('deleting: %s is: %s @ %s' % (f, q['name'], broker.url.hostname ))
                    qdc = sarracenia.moth.Moth.subFactory(
                        {
                            'broker': broker,
                            'dry_run': self.options.dry_run,
                            'credentials': o.credentials,
                            'echangeDeclare': False,
                            'subscription_index': 0,
                            'subscriptions' : [ s ],
                            'message_strategy': { 'stubborn':True },
                            'tlsRigour': q['tlsRigour']
                        })
                    qdc.getSetup()
                    qdc.getCleanUp()
```

I kept the Robs `self.o.update(copy.deepcopy(default_options))` in my last commit, but I think I'm also OK to reverting back to the original `self.o = copy.deepcopy(default_options)` if that's what people think is best.

reidsunderland:
**My preference is: `self.o.update(default_options)`** (what MQTT does) not `self.o = copy.deepcopy(default_options)` (what AMQP currently does).

`self.o.update(copy.deepcopy(default_options))`is the same as `self.o.update(default_options)`, just with a deepcopy that isn't actually needed.

It's less surprising. What's the point of having default options in the parent Moth class if they aren't used for AMQP?

andreleblanc11 (approved):

Issue #1449, opened 2025-05-29
v2 sometimes gets stuck in an infinite loop after broker connection is lost


reidsunderland:
Work around: switch to sr3.

We've seen this problem a few times, but I don't think it's documented in an issue. As far as we have observed, it only happens with one particular broker that is "load balanced" with DNS round robin. 

```
2025-05-27 13:35:52,688 [ERROR] Permission issue with anonymous@ddi... or exchange xpublic not found.
2025-05-27 13:35:52,688 [ERROR] sleeping 64 seconds to try binding again...
2025-05-27 13:36:56,715 [ERROR] bind queue: q_anonymous.sr_subscribe...
```

The problem is here: 

https://github.com/MetPX/sarracenia/blob/4b19896921460971049d43c8674d85ab78a89c7e/sarra/sr_amqp.py#L568-L586

I think we have to cleanly tear down the connection and start again from scratch, rather than just re-attempting the `bind()`. 

I labelled this wontfix, because we don't plan to do any more v2 releases, but if we find an urgent bug that requires a v2 release then this might be worth fixing at the same time.

reidsunderland:
This could also be worked around with a cronjob. This script is used to restart v2 polls that are hung and not fixed by sr_sanity/sr_audit and could be modified to look for "Permission issue with" in the logs instead:

```bash
#!/bin/bash

TO_EMAIL="person@example.com"

# Exit when not the sarra user
if [ "$(whoami)" != "sarra" ] ; then
        echo "This script must run as the sarra user"
        exit 1
fi

logpath=$(cd ~sarra/.cache/sarra/log/ && pwd)

# This will find logs from polls where the config has been removed
# and logs from polls that are supposed to be running, but the logs
# aren't being updated. We only care about the latter.
old_poll_logs=$(find ~sarra/.cache/sarra/log/ -name "sr_poll*log" -mmin +90 -type f )
old_poll_logs=${old_poll_logs//$'\n'/ }

# Filter out the polls where the config doesn't exist
stuck_polls=""
for poll_name in ${old_poll_logs}; do
        name=$(printf "%s" "${poll_name}" | sed "s|${logpath}/sr_poll_||g")
        name=${name%_*}
        if ! sr_poll status "${name}" 2>&1 | grep "not found" ; then
                echo "${name} is stuck"
                stuck_polls="${stuck_polls} ${poll_name}"
        else
                echo "${name} does not have a valid config, skipping"
        fi
done

[...]

PR #1594, opened 2026-03-19
v2: Fix issue #1449 where v2 gets stuck in an infinite loop

reidsunderland:
I marked issue #1449 as wontfix a while ago, but we keep having this bug cause issues, so I tried fixing it. It's annoying because when it happens, we have to login to a bunch of regional servers and restart sarra processes there. (Those will hopefully be upgraded to sr3 in May, but this will still help reduce the chance of problems in other places where v2 will stay a bit longer).

There are instructions to reproduce the issue inside #1449.

I made a lot of commits with small changes before I got it right, it's best to only look at the "Files Changed" tab.

I also realized that another way to get stuck in an infinite loop was by trying to create bindings on a queue that had been deleted. It's unlikely to happen, but I think it's possible so I added a fix for that too.

Using the latest sr_insects, the flow tests are inconsistent and broken:

**Static:**
```
v2_dev:       31/34, 33/34, 33/34
v2_issue1449: 31/34, 31/34, 33/34
```

**Flakey:**
```
v2_dev:       31/33, 18/33, 31/33, 31/33
v2_issue1449: 31/33, 30/33, 32/33, 31/33, 31/33
```

Using sr_insects commit 97f828e (most recent one that mentions v2):

**Flakey:**
```
v2_dev:       33/33, 33/33
v2_issue1449: 33/33, 33/33
```

### DIFF
--- a/sarra/sr_amqp.py
+++ b/sarra/sr_amqp.py
@@ -160,6 +160,10 @@ class HostConnect:
             except Exception as err:
                 self.logger.error("AMQP cannot connect to {} with {}".format(self.host, err))
                 self.logger.debug('Exception details: ', exc_info=True)
+                try:
+                    self.close()
+                except:
+                    pass
 
                 if not self.loop:
                     self.logger.error("giving up. Failed to connect to broker")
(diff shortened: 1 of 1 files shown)


petersilva (approved):
looks like good stuff. trusting the results you got.

PR #1596, opened 2026-03-20
Log post failures before retry clears worklist.failed

### LINKED ISSUE #1590: When a post fails, there is no log message from log.py
1. The `log` plugin has an `after_post` entry point, where it is supposed to log whether a message was posted successfully or not. (using worklist.failed and worklist.ok)
2. The `retry` plugin also has `after_post`. It removes all messages from the `worklist.failed` list and puts them into the retry list.

Because `retry` executes before `log`, there are no messages in worklist.failed when log runs, so the log message never gets printed.

I don't think changing the plugin load order would be a good fix, it would cause issues with after_accept and after_work...

```
[DEBUG] sarracenia.flow loadCallbacks flowCallback plugins to load: 
['sarracenia.flowcb.post.message.Message', 'sarracenia.flowcb.gather.message.Message', 
'sarracenia.flowcb.retry.Retry', 'sarracenia.flowcb.housekeeping.resources.Resources', 'log']
```

https://github.com/MetPX/sarracenia/blob/4fb8a33041e0c6530e4c7b5f10ed7ab9b43bb28a/sarracenia/flowcb/retry.py#L163-L174

https://github.com/MetPX/sarracenia/blob/4fb8a33041e0c6530e4c7b5f10ed7ab9b43bb28a/sarracenia/flowcb/log.py#L194-L199 

https://github.com/MetPX/sarracenia/blob/4fb8a33041e0c6530e4c7b5f10ed7ab9b43bb28a/sarracenia/config/__init__.py#L2015-L2018

robjarawan:
##### Summary
- The `retry` plugin's `after_post` runs before the `log` plugin, so by the time `log.after_post` checks `worklist.failed` it's already empty — failed posts are never logged.
- Added a `logger.warning` in `retry.after_post` that logs each failed message before clearing the list.

Closes #1590

##### Approach
There were a few ways to handle this — reordering plugin load order, stashing failed messages for log.py to pick up later, or just logging directly in retry.py before clearing the list. Went with the simplest option. If you'd rather handle it differently let me know.

##### Test plan
- Configure a sender pointing to an unreachable destination
- Confirm failed post messages now appear in the log as warnings
- Verify retry queue still works as before

### DIFF
--- a/sarracenia/flowcb/retry.py
+++ b/sarracenia/flowcb/retry.py
@@ -170,6 +170,9 @@ class Retry(FlowCB):
         for m in worklist.failed:
             self.__set_isRetry(m)
 
+        for m in worklist.failed:
+            logger.warning("post failed, queued to retry: %s %s" %
+                           (m.get('baseUrl', ''), m.get('relPath', '')))
         self.post_retry.put(worklist.failed)
         worklist.failed=[]
 


petersilva (commented):
Why a separate loop?  just add the log message to the existing loop above?
Does not matter much, but it seems like extra work for no reason.


* no separate loop required just put the logger statement in the loop above.
* the logger should be built with f-strings, not lazy.

reidsunderland:
This (putting the log message inside the retry class) is the easiest way to implement this specific change.

But when trying to use or write other plugins, I've sometimes had surprising behaviour, where `worklist.failed` is empty. This is problematic when trying to have a plugin do something in response to a failed post (e.g. send an email), and we've been working around it with `callback_prepend`. Using `callback_prepend` is okay, but it's confusing for new users.

I'm going to close this, because I think it's worth exploring a more significant change. Possibly loading the retry plugin later?

PR #1597, opened 2026-03-20
Add retryCountMax option to limit retry attempts

### LINKED ISSUE #1475: New Feature: retryCountLimit (maximum number of retries before giving up)
> I'm wondering if, as long as we're touching this, maybe change it to a count instead, so it increments every time retry is done?  ie... change *_isRetry* to *_RetryCount* so that instead of setting to True, we set it to a count.  We could then
add a "retryCountLimit" to give another way to limit how hard to deliver things... and be able to report on how many 
failures a given message has sufferred.  If the variable is either missing or 0 then it is the initial attempt.  Could even set it (to 0) for the initial attempt.

_Originally posted by @petersilva in https://github.com/MetPX/sarracenia/issues/1473#issuecomment-3118001285_

The retry code now sets `msg['_isRetry']` to an integer representing the number of times a retry has been attempted, so we can implement retryCountMax. 

Kind of similar to `attempts` which is the number of attempts to try *before* queueing a message for retry. With `attempts` and `retryCountMax`, the total amount of tries before giving up permanently would be $\text{attempts} \times \text{retryCountMax}$.

And it would work with the existing `retry_ttl` - I think whichever condition (retry_ttl or retryCountMax) is reached first should apply.

robjarawan:
##### Summary
- New config option `retryCountMax` (default `0` = unlimited) that drops messages from the retry queue after they've been retried too many times.
- Works alongside `retry_ttl` — whichever condition is reached first applies.
- The existing `_isRetry` counter (already an integer) is checked at all three points where messages come out of or go into retry queues: `gather()`, `after_accept()`, and `after_post()`.
- Messages that exceed the limit are logged at ERROR and discarded.

Closes #1475

##### Usage
```
retryCountMax 5
```
With `attempts 3` and `retryCountMax 5`, a message gets up to 3 transfer attempts per retry cycle, for up to 5 cycles, before being permanently dropped.

##### Test plan
- Set `retryCountMax` to a small value (e.g. 2) with an unreachable destination
- Verify messages are retried the expected number of times then dropped with an ERROR log
- Verify default (`retryCountMax 0`) preserves current unlimited behavior
- Verify interaction with `retry_ttl` — whichever limit is hit first should apply

### DIFF
--- a/sarracenia/config/__init__.py
+++ b/sarracenia/config/__init__.py
@@ -84,6 +84,7 @@ default_options = {
     'amqp_consumer': False,
     'attempts': 3,
     'batch' : 100,
+    'retryCountMax': 0,
     'baseDir': None,
     'baseUrl_relPath': False,
     'delete': False,
@@ -137,8 +138,8 @@ default_options = {
 }
 
 count_options = [
-    'batch', 'count', 'exchangeSplit', 'instances', 'logRotateCount', 'no', 
-    'post_exchangeSplit', 'prefetch', 'messageCountMax', 'runStateThreshold_cpuSlow', 
+    'batch', 'count', 'exchangeSplit', 'instances', 'logRotateCount', 'no',
+    'post_exchangeSplit', 'prefetch', 'messageCountMax', 'retryCountMax', 'runStateThreshold_cpuSlow',
     'runStateThreshold_disconnected', 
     'runStateThreshold_reject', 'runStateThreshold_retry', 'runStateThreshold_slow', 
 ]
--- a/sarracenia/flowcb/retry.py
+++ b/sarracenia/flowcb/retry.py
@@ -138,8 +138,11 @@ class Retry(FlowCB):
             return
 
         if len(worklist.failed) != 0:
-            logger.debug('putting %s messages into %s', len(worklist.failed), self.download_retry_name)
-            self.download_retry.put(worklist.failed)
+            for m in worklist.failed:
+                self.__set_isRetry(m)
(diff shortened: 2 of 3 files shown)


petersilva:
ok the option is added, but it is not documented... should have another commit with additions to 
docs/source/Reference/sr3_options.7.rst (English reference man page for all options.)
and docs/source/fr/Reference/sr3_options.7.rst (same in French.)

andreleblanc11:
Wouldn't we also want to call this in the `after_work` method in `retry.py`? I only see the retryCountMax filter code in `after_accept` in `after_post` entry points.

robjarawan:
Good catch -- yes, retryCountMax should also filter in `after_work` so messages that keep failing downloads also get capped. I'll add that. Also still need to add the docs Peter asked for.

andreleblanc11:
Looks good to me. I'll approve it once the documentation changes are placed (see Peter's last comment).

robjarawan:
Done - I've updated the PR with:

1. Documentation for  retryCountMax  in  docs/source/Reference/sr3_options.7.rst  and its French counterpart.
2. Refactored  retry.py  to ensure the retry count is incremented and filtered consistently across all entry points ( gather ,  after_accept ,  after_work , and  after_post ).
3. Resolved the double-increment issue where download retries were being counted twice per cycle.

This is now ready for another look.

robjarawan:
I have also created a dedicated documentation PR for these changes here: https://github.com/MetPX/sarracenia/pull/1678

andreleblanc11 (approved):

PR #1598, opened 2026-03-20
Support platformdirs as drop-in replacement for appdirs

### LINKED ISSUE #1493: Replace appdirs with platform dirs?
While looking into #1492, I noticed that python3-appdirs RPMs are not available for RHEL9.

https://rpmfind.net/linux/rpm2html/search.php?query=python3-appdirs

The [appdirs GitHub](https://github.com/ActiveState/appdirs?tab=readme-ov-file) repo hasn't been updated in two years and notes that users should switch to a fork called [PlatformDirs ](https://github.com/tox-dev/platformdirs) instead.

PlatformDirs is available for RHEL9, but not RHEL8: https://rpmfind.net/linux/rpm2html/search.php?query=python3-platformdirs

appdirs is available on all the versions of Ubuntu we use (18.04 - 24.04), while platformdirs is only available on 22.04 - 24.04. So I think we'll need to have code to choose one or the other depending on what's available.

For RPMs, we should be able to use a [boolean dependency ](https://rpm-software-management.github.io/rpm/manual/boolean_dependencies.html) (the RPM version installed on dirt RHEL8 is rpm-4.14.3-32.el8_10.x86_64 so that should be new enough). It looks like OR is also supported in Debian (https://www.debian.org/doc/debian-policy/ch-relationships.html).

robjarawan:
##### Summary
- Feature detection now supports `Alternate_modules` — if the primary module isn't found, alternates are tried before disabling the feature. The `appdirs` feature lists `platformdirs` as an alternate.
- At import time, `platformdirs` is preferred over `appdirs` since appdirs is unmaintained and unavailable on RHEL9. Both provide the same API for the three functions we use (`site_config_dir`, `user_config_dir`, `user_cache_dir`).
- Packaging changes (debian/control, RPM spec, pyproject.toml) are left out of this PR since those depend on target platform decisions — figured I'd leave that to whoever knows the packaging side best:D.

Closes #1493

##### Test plan
- Verify `sr3 features` shows appdirs as present when only `platformdirs` is installed
- Verify `sr3 features` shows appdirs as present when only `appdirs` is installed
- Verify fallback to hardcoded Linux paths when neither is installed
- Verify config/cache directories resolve correctly on each platform

### DIFF
--- a/sarracenia/__init__.py
+++ b/sarracenia/__init__.py
@@ -105,8 +105,12 @@ else:
         return "%d" % value
 
 
-if features['appdirs']['present']:
+if features['platformdirs']['present']:
+    import platformdirs as appdirs
+elif features['appdirs']['present']:
     import appdirs
+
+if features['platformdirs']['present'] or features['appdirs']['present']:
 
     def site_config_dir( app, author ):
         return appdirs.site_config_dir( app, author )
@@ -117,7 +121,7 @@ if features['appdirs']['present']:
     def user_cache_dir( app, author ):
         return appdirs.user_cache_dir( app, author )
 else:
-    # if appdirs is missing, pretend we're on Linux.
+    # if appdirs/platformdirs is missing, pretend we're on Linux.
     import pathlib
 
     def site_config_dir( app, author ):
--- a/sarracenia/featuredetection.py
+++ b/sarracenia/featuredetection.py
@@ -60,8 +60,11 @@ features = {
     'azurestorage' : { 'modules_needed': [ 'azure.storage.blob' ], 'present': False, 
             'lament' : 'cannot connect natively to Azure Stoarge accounts', 
             'rejoice' : 'can connect natively to Azure Stoarge accounts' },
-   'appdirs' : { 'modules_needed': [ 'appdirs' ], 'present': False, 
-           'lament' : 'assume linux file placement under home dir', 
+   'appdirs' : { 'modules_needed': [ 'appdirs' ], 'present': False,
+           'lament' : 'assume linux file placement under home dir',
+           'rejoice': 'place configuration and state files appropriately for platform (windows/mac/linux)', },
+   'platformdirs' : { 'modules_needed': [ 'platformdirs' ], 'present': False,
+           'lament' : 'assume linux file placement under home dir',
            'rejoice': 'place configuration and state files appropriately for platform (windows/mac/linux)', },
    'filetypes' : { 'modules_needed': ['magic'], 'present': False, 
            'lament': '(pip package python-magic, on windows python-magic-bin) will not be able to set content headers' ,
@@ -142,5 +145,4 @@ if features['mqtt']['present']:
     if not paho.mqtt.__version__ >= '2.1.0' :
         features['mqtt']['present'] = False
         logger.debug('paho-mqtt minimum version needed is 2.1.0')
-
 
--- a/tests/sarracenia/featuredetection_test.py
+++ b/tests/sarracenia/featuredetection_test.py
@@ -1,6 +1,14 @@
 import pytest
 from tests.conftest import *
-#from unittest.mock import Mock
+# from unittest.mock import Mock
 
 import sarracenia.config
 import sarracenia.featuredetection
+
+
+def test_platformdirs_is_a_separate_feature():
+    assert 'appdirs' in sarracenia.featuredetection.features
+    assert 'platformdirs' in sarracenia.featuredetection.features
+    assert sarracenia.featuredetection.features['appdirs']['modules_needed'] == ['appdirs']
+    assert sarracenia.featuredetection.features['platformdirs']['modules_needed'] == ['platformdirs']
+    assert 'Alternate_modules' not in sarracenia.featuredetection.features['appdirs']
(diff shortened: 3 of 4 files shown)


reidsunderland:
The alternate features thing is interesting, I like how it's an easy automatic substitution.

But thinking ahead towards the eventual removal of appdirs, I'm not sure that the extra code is necessary, and it might cause confusion for users. With the proposed change, when `platformdirs` is installed, `sr3 features` will show that appdirs is installed, which is misleading.

Adding platformdirs as a whole separate feature and importing like this might be a little simpler:

```python3
if features['platformdirs']['present']:
    import platformdirs as appdirs
elif features['appdirs']['present']:
    import appdirs
```

It's just a design/opinion thing. @andreleblanc11 or @petersilva do you have any preference?

Hopefully I don't get hit with the [wrath of AI](https://crabby-rathbun.github.io/mjrathbun-website/blog/posts/2026-02-11-gatekeeping-in-open-source-the-scott-shambaugh-story.html) for this :D

robjarawan:
So I reworked this to use your approach - platformdirs is its own feature entry now instead of the Alternate_modules thing. That way `sr3 features` actually shows which package you have installed instead of being misleading. platformdirs gets preferred when both are present. Pushed in the latest commit.

reidsunderland:
I don't see any new code on this branch, did you use a different branch or maybe forgot to push the change?

robjarawan:
Woops, I had it stashed locally and forgot to push — just pushed the fix. platformdirs and appdirs are now separate feature entries instead of using the Alternate_modules fallback. Should be good now!

robjarawan:

This is now rebased onto current `development`, so the merge conflict should be gone.

@reidsunderland the code that was missing before is definitely in this branch now:
- `platformdirs` is its own feature entry in `sarracenia/featuredetection.py`
- `appdirs` and `platformdirs` are reported independently by feature detection
- `sarracenia/__init__.py` prefers `platformdirs` when present, falls back to `appdirs`, then falls back to the hardcoded Linux paths
- the old `Alternate_modules` fallback is not used

I also added focused tests so this is explicit:
- `test_platformdirs_is_a_separate_feature`
- `test_platformdirs_preferred_over_appdirs`
- `test_appdirs_used_when_platformdirs_missing`

Tests:
- `python3 -m pytest tests/sarracenia/featuredetection_test.py tests/sarracenia/platformdirs_import_test.py -v` - 3 passed
- `pycodestyle --max-line-length=119 tests/sarracenia/featuredetection_test.py tests/sarracenia/platformdirs_import_test.py` - passed
- `git diff --check` - passed

reidsunderland:
This looks good, thanks for the updates.

I should have noted this before, but the packaging stuff should also be updated to prefer platformdirs where available. 

- https://github.com/MetPX/sarracenia/blob/development/debian/control (the `Depends:` section automatically gets filled in by the Python packaging stuff)
- https://github.com/MetPX/sarracenia/blob/development/metpx-sr3.spec#L17
- https://github.com/MetPX/sarracenia/blob/b4fa32f4faf0a6537ab43fa448972952737e6979/setup.py#L106-L109
- https://github.com/MetPX/sarracenia/blob/25e93352350f98e979a36b6707d8b769b74ebf1f/Dockerfile_base#L11
- https://github.com/MetPX/sarracenia/blob/25e93352350f98e979a36b6707d8b769b74ebf1f/requirements.txt#L4
- https://github.com/MetPX/sarracenia/blob/25e93352350f98e979a36b6707d8b769b74ebf1f/pyproject.toml#L37
- https://github.com/MetPX/sarracenia/blob/25e93352350f98e979a36b6707d8b769b74ebf1f/travis/flow_autoconfig_redhat.sh#L37

There's a lot of places to adjust and I'm not completely sure how to set up alternatives for each. In Debian we can use:

```
python3-platformdirs | python3-appdirs
```

But I'm not sure about the others. If you want we could merge this PR and deal with that separately?

robjarawan:
Sounds good @reidsunderland, let's merge this one as-is and I'll open a separate PR for the packaging side (debian/control, metpx-sr3.spec, setup.py, pyproject.toml, requirements.txt, Dockerfile_base, and the redhat autoconfig script). For debian the `python3-platformdirs | python3-appdirs` alternative you suggested works. For the pip-based files I think just switching to `platformdirs` is fine, since the code in this PR already falls back to appdirs when platformdirs isn't installed. I'll figure out the rpm side in that PR.

robjarawan:
The prefer-platformdirs logic is sound, but two things hold this back.

First, the body describes `Alternate_modules` machinery that the diff does not implement. The code and tests agree on a separate feature, so the body is the stale side. Please rewrite it to match, and rename the test so the intent is greppable.

Second, packaging. Fresh installs never get `platformdirs` since every packaging file still pins only `appdirs`. The follow-up needs `platformdirs` alongside `appdirs` in requirements, pyproject, debian control (either-or), and the spec, gated so the py3.6 and OEL8 fleet keeps working. Two amendments from verification: `setup.py` `install_requires` is dead under the current distutils path, so comment-only there, and the 3.8 leg needs an upper pin since 4.3+ wants 3.9 or newer.

Code fallback itself is correct, Linux paths are identical between the libs, and the Windows and macOS divergence is acknowledged scope.

PR #1599, opened 2026-03-20
Fix messageRateMax throttle to converge accurately

### LINKED ISSUE #1555: Improve accuracy of `messageRateMax`
I'm working on a case where I need to limit the download rate of an HTTPs server to a specific value. Large files and large volumes of data go through this feed at all times of the day so I configured a precise `messageRateMax` limit so that the flow of incoming messages doesn't surpass this threshold. If we get rate limited, the on-call analyst will have to intervene and fix the problem.

The limit imposed by the HTTPs server is 2 HTTP requests / sec.
I'm using a sarra config with a shared queue to download the data and I set  `messageRateMax` to `0.2` msgs /second * 8 servers = `1.6` msgs/second. I put a 20% buffer just in case the limit gets temporarily surpassed.

Unfortunately, when the queue gets full the imposed `messageRateMax` threshold gets continuously surpassed.

```
# Config gets started. I have `batch 1` set to maximise the calls to the sleep in the flow.
2026-01-21 16:46:04,987 [INFO] sarracenia.flow run current_rate/2 (39.34) above messageRateMax(0.20): throttling
2026-01-21 16:50:27,873 [INFO] sarracenia.flow run current_rate/2 (2.90) above messageRateMax(0.20): throttling
2026-01-21 16:50:28,106 [INFO] sarracenia.flow run current_rate/2 (3.25) above messageRateMax(0.20): throttling
2026-01-21 16:50:28,551 [INFO] sarracenia.flow run current_rate/2 (2.93) above messageRateMax(0.20): throttling
2026-01-21 16:50:28,552 [INFO] sarracenia.flow run current_rate/2 (3.65) above messageRateMax(0.20): throttling
2026-01-21 16:51:09,373 [INFO] sarracenia.flow run current_rate/2 (0.21) above messageRateMax(0.20): throttling
2026-01-21 16:51:11,167 [INFO] sarracenia.flow run current_rate/2 (0.23) above messageRateMax(0.20): throttling
2026-01-21 16:51:11,410 [INFO] sarracenia.flow run current_rate/2 (0.25) above messageRateMax(0.20): throttling
2026-01-21 16:51:11,892 [INFO] sarracenia.flow run current_rate/2 (0.27) above messageRateMax(0.20): throttling
2026-01-21 16:51:12,538 [INFO] sarracenia.flow run current_rate/2 (0.29) above messageRateMax(0.20): throttling
2026-01-21 16:51:12,539 [INFO] sarracenia.flow run current_rate/2 (0.31) above messageRateMax(0.20): throttling
2026-01-21 16:51:14,972 [INFO] sarracenia.flow run current_rate/2 (0.31) above messageRateMax(0.20): throttling
2026-01-21 16:51:15,227 [INFO] sarracenia.flow run current_rate/2 (0.33) above messageRateMax(0.20): throttling
2026-01-21 16:51:15,597 [INFO] sarracenia.flow run current_rate/2 (0.35) above messageRateMax(0.20): throttling
2026-01-21 16:51:15,598 [INFO] sarracenia.flow run current_rate/2 (0.37) above messageRateMax(0.20): throttling
2026-01-21 16:51:18,653 [INFO] sarracenia.flow run current_rate/2 (0.37) above messageRateMax(0.20): throttling
2026-01-21 16:51:18,894 [INFO] sarracenia.flow run current_rate/2 (0.39) above messageRateMax(0.20): throttling
2026-01-21 16:51:18,896 [INFO] sarracenia.flow run current_rate/2 (0.41) above messageRateMax(0.20): throttling
2026-01-21 16:51:22,632 [INFO] sarracenia.flow run current_rate/2 (0.40) above messageRateMax(0.20): throttling
2026-01-21 16:51:22,868 [INFO] sarracenia.flow run current_rate/2 (0.41) above messageRateMax(0.20): throttling
2026-01-21 16:51:23,488 [INFO] sarracenia.flow run current_rate/2 (0.43) above messageRateMax(0.20): throttling
2026-01-21 16:51:24,118 [INFO] sarracenia.flow run current_rate/2 (0.44) above messageRateMax(0.20): throttling
2026-01-21 16:51:24,120 [INFO] sarracenia.flow run current_rate/2 (0.46) above messageRateMax(0.20): throttling
2026-01-21 16:51:28,782 [INFO] sarracenia.flow run current_rate/2 (0.44) above messageRateMax(0.20): throttling
2026-01-21 16:51:28,997 [INFO] sarracenia.flow run current_rate/2 (0.45) above messageRateMax(0.20): throttling
2026-01-21 16:51:29,287 [INFO] sarracenia.flow run current_rate/2 (0.47) above messageRateMax(0.20): throttling
2026-01-21 16:51:29,563 [INFO] sarracenia.flow run current_rate/2 (0.48) above messageRateMax(0.20): throttling
2026-01-21 16:51:29,564 [INFO] sarracenia.flow run current_rate/2 (0.50) above messageRateMax(0.20): throttling
2026-01-21 16:51:33,637 [INFO] sarracenia.flow run current_rate/2 (0.48) above messageRateMax(0.20): throttling
2026-01-21 16:52:05,301 [INFO] sarracenia.flow run current_rate/2 (0.34) above messageRateMax(0.20): throttling
2026-01-21 16:52:05,714 [INFO] sarracenia.flow run current_rate/2 (0.35) above messageRateMax(0.20): throttling
2026-01-21 16:52:05,961 [INFO] sarracenia.flow run current_rate/2 (0.35) above messageRateMax(0.20): throttling
2026-01-21 16:52:06,500 [INFO] sarracenia.flow run current_rate/2 (0.36) above messageRateMax(0.20): throttling
2026-01-21 16:52:06,501 [INFO] sarracenia.flow run current_rate/2 (0.37) above messageRateMax(0.20): throttling
2026-01-21 16:52:09,985 [INFO] sarracenia.flow run current_rate/2 (0.37) above messageRateMax(0.20): throttling
2026-01-21 16:52:10,220 [INFO] sarracenia.flow run current_rate/2 (0.38) above messageRateMax(0.20): throttling
2026-01-21 16:52:10,780 [INFO] sarracenia.flow run current_rate/2 (0.39) above messageRateMax(0.20): throttling
2026-01-21 16:52:11,007 [INFO] sarracenia.flow run current_rate/2 (0.39) above messageRateMax(0.20): throttling
2026-01-21 16:52:11,008 [INFO] sarracenia.flow run current_rate/2 (0.40) above messageRateMax(0.20): throttling
2026-01-21 16:52:14,812 [INFO] sarracenia.flow run current_rate/2 (0.40) above messageRateMax(0.20): throttling
2026-01-21 16:52:46,716 [INFO] sarracenia.flow run current_rate/2 (0.32) above messageRateMax(0.20): throttling

# Eventually the rate stabilised to about ~0.6 msgs / second.
```

The only way I've been able to force the rate threshold to 0.2 msgs/seconds per node is by setting an appropriate `sleep` value. So,

```
# Maxmimizes time spent in sleep
batch 1
# Wait 5 seconds between every message that comes through the batch.
# This essentially forces the rate of incoming messages to be 0.2 msgs/second. 1 msg per batch / 5 seconds = 0.2 msgs per batch per second
sleep 5
messageRateMax 0.2
```

robjarawan:
##### Summary
- The old throttle formula `1 + 2 * ((current_rate - max) / max)` didn't sleep long enough to bring the rate down to the target. It would stabilize around 2-3x the configured limit.
- Replaced with an exact calculation: `sleep = total_messages / messageRateMax - run_time`. This guarantees the cumulative average rate equals the target after sleeping.
- Example: 10 messages in 5 seconds (rate=2.0), limit 0.2 -> sleep 45s -> rate becomes 10/50 = 0.2

Closes #1555

##### Test plan
- Set `messageRateMax 0.2` with `batch 1` against a fast source
- Verify the rate converges to the target instead of stabilizing above it
- Verify the log message shows the actual sleep duration for easier debugging
- Verify that `messageRateMax 0` (default, unlimited) still works with no throttling

### DIFF
--- a/sarracenia/flow/__init__.py
+++ b/sarracenia/flow/__init__.py
@@ -662,11 +662,14 @@ class Flow:
 
             if (self.o.messageRateMax > 0) and (current_rate >=
                                                 self.o.messageRateMax):
-                stime = 1 + 2 * ((current_rate - self.o.messageRateMax) /
-                                 self.o.messageRateMax)
+                # sleep exactly long enough to bring the average rate to the target:
+                # we want total_messages / (run_time + sleep) == messageRateMax
+                stime = (total_messages / self.o.messageRateMax) - run_time
+                if stime < 0:
+                    stime = 0
                 logger.info(
-                    "current_rate/2 (%.2f) above messageRateMax(%.2f): throttling"
-                    % (current_rate, self.o.messageRateMax))
+                    "current_rate (%.2f) above messageRateMax(%.2f): sleeping %.2fs"
+                    % (current_rate, self.o.messageRateMax, stime))
             else:
                 logger.debug( f" not throttling: limit: {self.o.messageRateMax} " )
                 stime = 0


petersilva (approved):
I do not have a way to test this. I see no harm and more confirmation of performance can happen in mainline.

andreleblanc11:
I just tested this PR and the logic is wrong.

Calculating the `stime` based on the `run_time` doesn't make sense for some use cases (like large data transfers). It can result in the sleep code not being executed and this in turn creates a linearly growing sleep value until the sleep code is finally executed.

This is what I saw from my logs when I tried testing. I had `messageRateMax 0.2` and `batch 1`. `sleep` was not configured. My feed downloads large files (>400MB) quite often.

```
2026-03-30 18:22:59,419 [DEBUG] sarracenia.flow run sleeping for 0.10
(Downloads logs)
2026-03-30 18:22:59,781 [INFO] sarracenia.flow run current_rate (0.20) above messageRateMax(0.20): sleeping 3.38s
...
2026-03-30 19:25:54,110 [INFO] sarracenia.flow run current_rate (0.43) above messageRateMax(0.20): sleeping 9359.05s
2026-03-30 19:25:54,110 [DEBUG] sarracenia.flow run worked too long to sleep!
```

petersilva:
9000 seconds is... um... a bit long.

reidsunderland:
I think the old formula is correct. It would adjust the sleep to slowly bring the rate down to the configured rate. The new formula forces it to hit the configured rate too aggressively, by sleeping for too long immediately.

fwiw over long run times the old formula seemed to work fine.

I actually think the change in #1602 should also have a big impact on messageRateMax working properly, or maybe even completely fix it. Before that change, the code was not sleeping long enough, so that would cause the message rate to be higher than expected.

In one of our configs using `messageRateMax 0.2`, Andre put this note:

```
messageRateMax 0.2

# Only fetch one file per flow run. This maximizes the time slept.
batch 1

# Force a sleep every 5 seconds.
# This theoretically forces sr3 to have a maximum rate of 1 message per batch / 5 seconds = 0.2 messages per batch per second.
# See https://github.com/MetPX/sarracenia/issues/1555. Setting messageRateMax only isn't enough to limit the incoming flow of messages.
sleep 5
```

With that `sleep 5`, the messageRateMax seems to be pretty correct most of the time, which indicates that the old, wrong sleep code was a big part of the issue.

andreleblanc11:
It's true that the `messageRateMax` value should be more accurate with the sleep change in theory. I did some testing with having `messageRateMax 0.2`, WITHOUT the `sleep` in the configuration options and with the code before Robbie's patch.

The calculated rate seems to fluctuate between 0.3 and 0.8 in value, never reaching the `0.2` value defined in the configuration.

I think I may know why this is happening. With the algorithm before the patch, the `stime` becomes shorter the closer you reach the `current_rate` becomes to the `messageRateMax` rate. When you hit the `messageRateMax` rate, the `stime` becomes 1. The opposite is also true. The further `current_rate` is from `messageRateMax`, the more sleep is done. This in turn creates kind of a sleep "yoyo" affect. 

***

I don't really understand why we'd want to have `stime` converge towards a value of 1 the closer you are to the maximum rate. To me that just means you will never truely reach the maximum rate you've configured when there is high traffic coming into your process.

***

Another problem is that the `total_messages` variable takes into account all of the messages that have been processed since the process starts running and `run_time` is the total amount of time the configuration has been running since it was started.

In my opinion, it makes more sense to have the sleep time adjusted based on the # of messages processed in the flow run (batch) and the sleep time of that run.

petersilva:
I think one batch is too little... probably resetting every housekeeping interval will be more reasonable.
If it runs well below max rate for days, and then starts running quickly... it could potentially run over max rate for a days before it gets limited to the configured max rate.  I guess having a shorter time horizon would help with that (assuming we don't want it to exceed max rate for a long period until the overall rate approaches configured max.)  but... is that a problem that happens?  I guess it would happen with op runs...

andreleblanc11:
I reverted this commit : https://github.com/MetPX/sarracenia/commit/7411484610ea1b319adbed32418eae3e49715651

PR #1600, opened 2026-03-20
Apply logLevel setting to Poll parent logger from subclasses

### LINKED ISSUE #1542: `set poll.Poll.logLevel debug` doesn't work in a poll plugin
Reid already tried to fix this with https://github.com/MetPX/sarracenia/commit/1103455e02889cf31a7cccf7a0b691b496f534c9.

I was trying to debug a problem with a custom poll plugin but noticed the following options didn't work in my configuration.

```
set sarracenia.flowcb.poll.Poll.logLevel debug
set sarracenia.flowcb.poll.logLevel debug
```

I was working on a dev environment and I added the following to the source code, which seems to have done the trick. 

```
+++ b/sarracenia/flowcb/poll/__init__.py
@@ -258,6 +259,12 @@ class Poll(FlowCB):

         super().__init__(options,class_logger)

+        if 'sarracenia.flowcb.poll.Poll' in self.o.settings and 'logLevel' in self.o.settings['sarracenia.flowcb.poll.Poll']:
+            logger.setLevel(
+                getattr(
+                    logging,
+                    self.o.settings['sarracenia.flowcb.poll.Poll']['logLevel'].upper()))
+

```

I'm not saying this is what I suggest is the permanent fix. I'm just stating that adding this fixed my problem.

robjarawan:
##### Summary
- When using a Poll subclass (e.g. a custom HTML parser plugin), `set sarracenia.flowcb.poll.logLevel debug` didn't work because the Poll module's own `logger` never had its level changed — only the subclass's logger was updated via `FlowCB.__init__`.
- Now `Poll.__init__` checks `self.o.settings` for the Poll class path and applies the logLevel to the module-level logger when `class_logger != logger` (i.e. when called from a subclass).
- Checks both `sarracenia.flowcb.poll.Poll` and `sarracenia.flowcb.poll` key paths in settings.

Closes #1542

##### Test plan
- Create a Poll subclass plugin with its own logger
- Set `set sarracenia.flowcb.poll.logLevel debug` in the config
- Verify debug messages from the parent Poll class now appear in the log
- Verify that using Poll directly (no subclass) still works as before


petersilva (approved):

andreleblanc11:
Not related to this PR.. the basic declare github actions have been consistently failing since around March 26th. https://github.com/MetPX/sarracenia/actions/workflows/flow_basic.yml. It's probably worth investigating.

andreleblanc11:
Was this tested?

PR #1601, opened 2026-03-20
Fix diskqueue race conditions with concurrent sr3_post

### LINKED ISSUE #1494: Multiple sr3_posts running and retrying at the same time can cause issues (using the same retry diskqueue)
As far as we know right now, there are no actual problems, but it seems like something that is worth documenting and investigating. 

Apparently URP servers run ~32 processes in parallel, and they all call sr3_post when they finish. This means that it's possible to have multiple sr3_posts running with the same config at the same time. 

We have occasionally seen this in the logs:

```
FileNotFoundError: [Errno 2] No such file or directory: '/apps/urp/.cache/sr3/post/post_ddsr_dev_6min/diskqueue_post_retry_000.new'
```

We have a check in the diskqueue code to avoid opening a file when it doesn't exist, so this must mean that one process is deleting the file *after* another process checks that it exists but *before* it opens it. 

https://github.com/MetPX/sarracenia/blob/c0fc15632ba8ff5bb68f504ecde8955172cef409/sarracenia/diskqueue.py#L344-L351

The exception is handled, so it doesn't crash, but if there really was a case where the posting failed and needed to be retried, this could be problematic. 

I suggested a workaround to add `--no X` to the sr3_post calls, so each one will have a different instance number and they won't conflict. In my very basic tests, this seems to work fine.

robjarawan:
##### Summary
- Replace check-then-open patterns (`os.path.isfile` then `open`) with `try/except FileNotFoundError` in `diskqueue.py`.
- When multiple `sr3_post` processes share the same diskqueue files (e.g. 32 URP processes posting with the same config), one process can delete a file between another's existence check and open call.
- Three locations fixed: `msg_get_from_file` (the reported crash site), `_count_msgs`, and `__init__` (queue_file/new_path checks).

Note: this doesn't address the concurrent append issue (multiple processes writing to the same `.new` file), which could interleave JSON lines and corrupt data. That would need file locking and is a bigger change. The `--no X` workaround (giving each process a different instance number) avoids that problem entirely.

Closes #1494

##### Test plan
- Run multiple `sr3_post` processes concurrently with the same config
- Verify no `FileNotFoundError` crashes in diskqueue
- Verify retry queue still functions correctly after concurrent access

### DIFF
--- a/sarracenia/diskqueue.py
+++ b/sarracenia/diskqueue.py
@@ -116,18 +116,21 @@ class DiskQueue():
         #   ... new messages will only be available in the next interval.
         self.msg_count_new = 0
 
-        if not os.path.isfile(self.queue_file):
+        try:
+            retry_age = os.path.getmtime(self.queue_file)
+        except FileNotFoundError:
             return
 
-        retry_age = os.path.getmtime(self.queue_file)
         self.msg_count = self._count_msgs(self.queue_file)
 
-        if os.path.isfile(self.new_path):
+        try:
             new_age = os.path.getmtime(self.new_path)
             if retry_age > new_age:
                 os.unlink(self.new_path)
             else:
                 self.msg_count_new = self._count_msgs(self.new_path)
+        except FileNotFoundError:
+            pass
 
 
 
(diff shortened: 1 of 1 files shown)


petersilva (changes requested):
I don't agree with this approach.  diskqueue is not meant to be accessed concurrently.
posting should be changed not to use diskqueue in concurrent cases.

robjarawan:
Closing per Peter's feedback -- diskqueue is designed for single-writer access and bolting race condition handling onto it is the wrong approach. The fix should be on the caller side so concurrent sr3_post processes don't share the same diskqueue files.

Leaving issue #1494 open since the problem (URP servers running ~32 concurrent sr3_posts with the same config) is real.

PR #1603, opened 2026-03-21
Fix metricsReport plugin list leak in flow engine

robjarawan:
##### what

Sniped another one while profiling memory. `_runCallbackMetrics` was grabbing a direct reference to `self.plugins["metricsReport"]` and appending transfer protocol functions onto it every housekeeping cycle. So the list just grows forever - and every subsequent cycle calls all the duplicates too.

##### how it leaks

Housekeeping runs every 300s by default. Each cycle appends one bound method per active transfer protocol (sftp, ftp, etc.) onto the live plugin list. After 24 hours that's ~288 leaked entries per protocol, and each one gets called on every future cycle. So it compounds - by day 2 you're running ~576 redundant `metricsReport()` calls every 5 minutes haha.

##### the fix

One-liner: `list(self.plugins["metricsReport"])` instead of `self.plugins["metricsReport"]` so we work on a copy. Added a regression test that confirms the list stays stable across repeated calls.

##### context

I started tracking RSS growth on a couple long-running flows at work this week and I hope to find some contributors... The leak is small per cycle but it compounds over days/weeks of uptime.

### DIFF
--- a/sarracenia/flow/__init__.py
+++ b/sarracenia/flow/__init__.py
@@ -349,26 +349,30 @@ class Flow:
             self.metrics['flow']['transferConnectTime'] += now - self.metrics['flow']['transferConnectStart']
             self.metrics['flow']['transferConnectStart']=now
 
-        modules=list(self.plugins["metricsReport"])
-
-        if hasattr(self,'proto'): # gets re-spawned every batch, so not a permanent thing...
-            for scheme in self.proto:
-                if hasattr(self.proto[scheme], 'metricsReport'):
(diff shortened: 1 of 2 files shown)


andreleblanc11:
We've been testing this branch on one of our data pumps and the results after about 18 hours of runtime are already amazing. Graph in blue is the server **without this patch**, server1. Graph in red is the server **with this patch**, server 2.

Server 1 has had its load triple in about 18 hours vs. Server 2 has had its load steady during this same time period. 

image

The load will likely keep going up on server 1

andreleblanc11 (approved):

PR #1604, opened 2026-03-21
Replace deepcopy with shallow copy in AMQP publish

robjarawan:
##### what

So uh I found a gas eater.. i did 3 regression tests to be sure. `putNewMessage` in `moth/amqp.py` runs `copy.deepcopy(message)` on every single outbound message just to strip some internal keys before serializing. deepcopy walks and clones the entire message tree — nested dicts like `integrity`, `blocks`, `fileOp`, all of it — every publish.

##### how bad is it

Benchmarked with a realistic message (nested dicts, sets, typical field count):

| method | msg/s | relative |
|---|---|---|
| `copy.deepcopy` | 41,198 | 1x |
| `dict()` shallow copy | 1,658,089 | **40x faster** |

On a busy flow doing thousands of publishes per second, that's a lot of wasted CPU and GC pressure for no reason.

##### why shallow copy is safe

The deepcopy was there to protect the caller's message from modification. But the only thing `putNewMessage` does to the copy is delete top-level keys listed in `_deleteOnPost` — it never mutates any nested dicts. Shallow copy (`dict(message)`) gives a new top-level dict with shared references to the nested objects, so deleting keys from the copy doesn't touch the original. Confirmed with tests.

Also verified that `exportAny` (v03) just does `json.dumps(body)` — read-only. v02's `exportMine` does its own internal deepcopy of a wrapped message, so it's safe too.

##### the fix

One-liner: `dict(message)` instead of `copy.deepcopy(message)`. Added regression tests proving the original message stays intact after the _deleteOnPost stripping.

### DIFF
--- a/sarracenia/moth/amqp.py
+++ b/sarracenia/moth/amqp.py
@@ -685,8 +685,8 @@ class AMQP(Moth):
                 logger.debug('Exception details: ', exc_info=True)
                 return False
 
-        # The caller probably doesn't expect the message to get modified by this method, so use a copy of the message
-        body = copy.deepcopy(message)
+        # Shallow copy: only top-level keys are deleted (_deleteOnPost), nested dicts are read-only
+        body = dict(message)
 
         if 'format' in self.o:
             version=self.o['format']
--- a/tests/sarracenia/moth/amqp_shallow_copy_test.py
+++ b/tests/sarracenia/moth/amqp_shallow_copy_test.py
@@ -1,0 +1,85 @@
+"""
+Regression test: putNewMessage must not modify the caller's message.
+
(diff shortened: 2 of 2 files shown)


petersilva:
People wanted multi-publish support.  If you want to publish a message multiple times, you can't destructively modify it with each publish.

petersilva:
Need to add multipublishing to the flow tests, and/or unit tests to validate that removing the deepcopy does not break anything.

petersilva:
I just saw that the branch passes the multi_static_flow test... so my objection is addressed.

petersilva (approved):
GREAT!

Issue #1608, opened 2026-03-21
_runHousekeeping calls on_housekeeping twice per cycle


robjarawan:
_runHousekeeping in flow/__init__.py calls self.on_housekeeping() directly via an inline block, then immediately calls self.runCallbacksTime('on_housekeeping') which also calls self.on_housekeeping() via getattr.

This means any flow subclass that overrides on_housekeeping() gets called twice per housekeeping cycle. This is likely unintentional but may be relied upon — needs team discussion before changing.

https://github.com/MetPX/sarracenia/blob/286a8dd403a5156f9c30565d712657b33a30c27a/sarracenia/flow/__init__.py#L411-L420

The inline block at lines 411-419 appears to be redundant since runCallbacksTime already handles both the flow method and plugin dispatch.

reidsunderland:
This is an interesting find. None of the `Flow` (not `FlowCB`) classes actually have an `on_housekeeping` method, so we've never experienced this issue. The code should still be fixed, imo the right fix is removing the explicit `self.on_housekeeping` calls inside `_runHousekeeping` (delete line 409-417).

andreleblanc11:
I also think what Reid suggested is the right fix.

PR #1609, opened 2026-03-21
Replace eval() with getattr() for callback dispatch

### LINKED ISSUE #1607: Flow callback dispatch uses eval() instead of getattr()
_runCallbacksWorklist and runCallbacksTime in flow/__init__.py use eval(f"self.{entry_point}(...)") to dispatch callbacks. This compiles Python source strings on every batch for every callback entry point.

getattr(self, entry_point) is the standard way to do dynamic attribute lookup — no string compilation needed.

Also found: runCallbacksTime is missing an entry_point guard on self.plugins, which can KeyError if an entry point has no registered plugins.

robjarawan:
Closes #1607

##### what

Flow callback dispatch was using eval(f"self.{entry_point}(...)") to call flow subclass methods — compiling Python source on every batch, every callback entry point. Replaced with getattr() for a direct attribute lookup.

##### changes

- eval() -> getattr(self, entry_point, None) — single lookup, no string compilation
- logLevel.lower() cached once per method call instead of per-plugin-iteration
- Added missing entry_point guard in runCallbacksTime to prevent KeyError when no plugins are registered for an entry point

##### not included

Noticed that _runHousekeeping calls on_housekeeping twice per cycle (once inline, once via runCallbacksTime). Filed separately as #1608 since it's a behavior change that needs team input.

### DIFF
--- a/sarracenia/flow/__init__.py
+++ b/sarracenia/flow/__init__.py
@@ -285,19 +285,22 @@ class Flow:
 
     def _runCallbacksWorklist(self, entry_point):
 
-        if hasattr(self, entry_point):
-            if self.o.logLevel.lower() == 'debug' :
-                eval( f"self.{entry_point}(self.worklist)")
+        debug = self.o.logLevel.lower() == 'debug'
+
+        fn = getattr(self, entry_point, None)
+        if fn is not None:
+            if debug:
+                fn(self.worklist)
             else:
                 try:
-                    eval( f"self.{entry_point}(self.worklist)")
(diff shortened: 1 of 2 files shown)


petersilva:
recommend merging this ASAP.  should have a huge effect on cpu usage. &#128293;

petersilva (approved):
I  ran the tests, and confirmed they don't break anything.
Reviewed the code, and the changes look fine.
Not merging for now to avoid interference with 3.02 release.

PR #1610, opened 2026-03-22
remove duplicate on_housekeeping call in _runHousekeeping

### LINKED ISSUE #1608: _runHousekeeping calls on_housekeeping twice per cycle

robjarawan:
##### Summary

Fixes #1608 - `_runHousekeeping` was calling `on_housekeeping()` twice per cycle: once inline (lines 409–417) and again via `runCallbacksTime('on_housekeeping')`. The inline block is redundant since `runCallbacksTime` already dispatches to both the flow method and plugins with identical error handling.

Removed the inline block, keeping only the `runCallbacksTime` call.

##### Test

Added a unit test that asserts `on_housekeeping` is called exactly once per housekeeping cycle.

### DIFF
--- a/sarracenia/flow/__init__.py
+++ b/sarracenia/flow/__init__.py
@@ -419,16 +419,6 @@ class Flow:
             Return the time when housekeeping should be run next
         """
         logger.info(f'on_housekeeping pid: {os.getpid()} {self.o.component}/{self.o.config} instance: {self.o.no}')
-        if hasattr(self, "on_housekeeping"):
-            if self._logLevel_debug :
-                self.on_housekeeping()
-            else:
-                try:
-                    self.on_housekeeping()
-                except Exception as ex:
-                    logger.error( f'flow on_housekeeping crashed: {ex}' )
-                    logger.debug( "details:", exc_info=True )
-
         self.runCallbacksTime('on_housekeeping')
         self.metricsFlowReset()
         self.metrics['flow']['last_housekeeping'] = now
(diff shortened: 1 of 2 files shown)


robjarawan:

This is now rebased onto current `development`, so the merge conflict should be gone. I kept the change focused to the original #1608 fix.

Changes in this refresh:
- remove the inline `self.on_housekeeping()` call in `_runHousekeeping`
- keep housekeeping dispatch going through `runCallbacksTime('on_housekeeping')`
- add a regression test to confirm `on_housekeeping` is called once per housekeeping cycle
- keep the existing callback dispatch tests from current `development`

Tests:
- `python3 -m pytest tests/sarracenia/flow/callback_dispatch_test.py -v` - 8 passed
- `python3 -m pytest tests/sarracenia/flow/callback_dispatch_test.py tests/sarracenia/flow/__flow___test.py -v` - 10 passed
- `pycodestyle --max-line-length=119 tests/sarracenia/flow/callback_dispatch_test.py` - passed
- `git diff --check` - passed

@reidsunderland @andreleblanc11 this should be ready for another look.

reidsunderland (approved):

PR #1611, opened 2026-03-22
replace deepcopy with shallow copy in MQTT publish path

robjarawan:
Same fix as ebae12f (AMQP side) -- `putNewMessage` in mqtt.py was still using `copy.deepcopy(message)` on every published message. Only top-level keys get deleted via `_deleteOnPost`, so `dict()` shallow copy is enough.

`deepcopy` is one of the most expensive stdlib calls -- it walks the entire object graph recursively. On a busy sender pushing thousands of messages/sec this adds up. Peter's multi_static benchmarks showed ~5% less time with the AMQP-side fix alone, so this should help further for MQTT brokers.

Also removed the `import copy` since it was the only usage in mqtt.py.

##### test plan
- [x] regression test mirrors the AMQP test (mqtt_shallow_copy_test.py)
- [x] full suite passes (144 passed, 1 pre-existing failure in v02_test)

### DIFF
--- a/sarracenia/moth/mqtt.py
+++ b/sarracenia/moth/mqtt.py
@@ -22,7 +22,6 @@
 #
 
 import collections
-import copy
 import json
 import logging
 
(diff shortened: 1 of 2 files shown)


petersilva (commented):
confirmed it works with static-multi flow test.  This should help reduce system overhead shown in #1035

petersilva (approved):

PR #1612, opened 2026-03-22
fix NameError crash in report callback, replace deepcopy with shallow copy

robjarawan:
Linked to #1035 (sr3 CPU usage investigation).

report.py line 128 called `copy.deepcopy(m)` on failed messages but never imported `copy`. This means whenever `worklist.failed` is non-empty during the report callback, it crashes with `NameError: name 'copy' is not defined`.

`reportPost` only touches top-level keys -- it deletes `content` and reassigns `_deleteOnPost` via `.difference()`. Neither of those mutate nested dicts, so `dict()` shallow copy is all that's needed.

Same category as ebae12f (AMQP) and PR #1611 (MQTT).

Note: `gather/file.py:283` also uses deepcopy but that one is correct -- it mutates nested dicts (`msg['blocks']['number']`, `msg['identity']['value']`) inside a loop and needs true deep copies.

##### test plan
- [x] 4 regression tests including one that calls `report()` on the real class with mocked poster
- [x] full suite passes (144 passed, 1 pre-existing v02_test failure)


reidsunderland (changes requested):
Changing to shallow copy is probably good, but it should also use `copy.copy` instead of `dict` (and needs `import copy`).

petersilva:
report.py is barely used at all, very little impact on current deployments... but performance is a good thing.

robjarawan:
@reidsunderland this should be ready for re-review now.

I updated it to use `copy.copy()` instead of `dict()` and added the `import copy`, per your review.

Current branch status:
- `import copy` is present in `sarracenia/flowcb/report.py`
- failed report messages use `copy.copy(m)` before `reportPost()` mutates top-level keys
- Unit test is passing
- Test Results is passing: 322 tests, 321 passed, 1 skipped
- branch is mergeable against current `development`

reidsunderland (approved):

PR #1613, opened 2026-03-22
use lazy formatting in debug logs to avoid wasted CPU

robjarawan:
Linked to #1035 (sr3 CPU usage investigation).

`logger.debug("x %s" % val)` evaluates the format string before the logger checks if debug is enabled. On production nodes running at INFO or WARNING, this is pure waste -- the formatted string gets built and immediately discarded.

Changed to `logger.debug("x %s", val)` which defers formatting until the logger confirms the message will actually be emitted. Converted all eager `%` and f-string debug calls in `flow/__init__.py` hot paths:

- `filter()` -- runs every batch, multiple len() calls per invocation
- `work()` / `download()` / `send()` -- per-message path operations
- main `run()` loop -- messageRateMax throttle logging

21 lines changed, same log output when debug is on, zero formatting cost when it's off. Same category as the deepcopy fixes (#1609, #1611, #1612) -- small per-message overhead that compounds at scale.

##### test plan
- [x] full suite passes (140 passed, 1 pre-existing v02_test failure)
- [x] no behavior change -- same format strings, just deferred evaluation

### DIFF
--- a/sarracenia/flow/__init__.py
+++ b/sarracenia/flow/__init__.py
@@ -555,7 +555,7 @@ class Flow:
             logger.debug("options:")
             self.o.dump()
 
-        logger.debug("callbacks loaded: %s" % self.plugins['load'])
+        logger.debug("callbacks loaded: %s", self.plugins['load'])
         logger.debug(
             f'pid: {os.getpid()} {self.o.component}/{self.o.config} instance: {self.o.no}'
         )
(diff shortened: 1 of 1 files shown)


petersilva (approved):
see #1623 for more robust discussion.  Probably should trigger more PR's.

petersilva:
no longer needed, as replaced by #1626

PR #1617, opened 2026-03-24
Fix diskqueue fd hygiene: fsync bug and bare except cleanup

robjarawan:
##### what

Three issues in `diskqueue.py` that affect reliability of the retry queue:

1. **`os.fsync` bug in `close()`** -- `os.fsync(self.new_fp)` passes a file object instead of a file descriptor. This always raised `TypeError`, but the bare `except: pass` swallowed it. The new_fp was never actually fsynced before close, so queued retry messages could be lost on crash.

2. **Bare `except:` blocks** -- 11 bare `except:` blocks throughout the file catch everything including `KeyboardInterrupt` and `SystemExit`. On `sr3 stop` or ctrl-c, these silently swallow the signal instead of letting the process shut down cleanly.

3. **Silent I/O errors in `close()`** -- if `close()` fails on any file pointer (disk full, stale NFS handle), there was zero indication in the logs.

##### fix

- Changed `os.fsync(self.new_fp)` to `self.new_fp.flush()` + `os.fsync(self.new_fp.fileno())`
- Narrowed all `except:` to `except Exception:` so `KeyboardInterrupt`/`SystemExit` propagate
- Added `logger.debug` in `close()` for I/O errors

##### regression tests

- `test_close__None_fps` -- close() with all None fps doesn't crash
- `test_close__already_closed_fps` -- close() with pre-closed fps is graceful
- `test_close__fsync_uses_fileno` -- verifies fsync gets the fd, not the file object
- `test_close__keyboard_interrupt_propagates` -- KeyboardInterrupt not swallowed in close()
- `test_get__keyboard_interrupt_propagates` -- KeyboardInterrupt not swallowed in get()

All 22 diskqueue tests pass (17 existing + 5 new).

### DIFF
--- a/sarracenia/diskqueue.py
+++ b/sarracenia/diskqueue.py
@@ -160,17 +160,18 @@ class DiskQueue():
         """
         try:
             self.housekeeping_fp.close()
-        except:
-            pass
+        except Exception as err:
+            logger.debug("housekeeping_fp close: %s" % err)
         try:
-            os.fsync(self.new_fp)
+            self.new_fp.flush()
+            os.fsync(self.new_fp.fileno())
             self.new_fp.close()
-        except:
-            pass
+        except Exception as err:
+            logger.debug("new_fp close: %s" % err)
         try:
             self.queue_fp.close()
-        except:
-            pass
+        except Exception as err:
+            logger.debug("queue_fp close: %s" % err)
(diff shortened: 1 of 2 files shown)


petersilva (approved):
should improve reliability.

PR #1618, opened 2026-03-24
Cache logLevel debug check in callback dispatch

robjarawan:
##### what

Every callback dispatch method in `flow/__init__.py` calls `self.o.logLevel.lower() == 'debug'` to decide whether to wrap the call in try/except. This runs `.lower()` and a string comparison on every single callback invocation -- 10 call sites, hit on every message batch, every poll cycle, every housekeeping run.

##### the fix

Cache the result as `self._logLevel_debug` (a bool) once at init time. `logLevel` does not change at runtime so this is safe. Replaces all 10 occurrences.

##### regression tests

New test file `tests/sarracenia/flow_loglevel_test.py` with 6 tests:

- `test_logLevel_debug_flag_set_correctly` -- True for DEBUG, False for INFO/WARNING
- `test_logLevel_debug_flag_case_insensitive` -- handles mixed case
- `test_runCallbacksWorklist_debug_propagates_exception` -- in debug mode, plugin crashes propagate
- `test_runCallbacksWorklist_info_catches_exception` -- in non-debug mode, plugin crashes are caught and logged
- `test_runCallbacksTime_debug_propagates_exception` -- same for time-based callbacks
- `test_runCallbacksTime_info_catches_exception` -- same for time-based callbacks

### DIFF
--- a/sarracenia/flow/__init__.py
+++ b/sarracenia/flow/__init__.py
@@ -159,6 +159,8 @@ class Flow:
         logging.basicConfig(format=self.o.logFormat,
                             level=getattr(logging, self.o.logLevel.upper()))
 
+        self._logLevel_debug = self.o.logLevel.lower() == 'debug'
+
         self.plugins = {}
         for entry_point in sarracenia.flowcb.entry_points:
             self.plugins[entry_point] = []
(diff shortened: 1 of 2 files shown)


petersilva (approved):

petersilva:
removed approval because it now has a conflict.  needs review.

PR #1581, opened 2026-03-15
Close connections on failed connect to prevent fd exhaustion

robjarawan:
## Summary
- When SFTP/FTP `connect()` fails partway through (e.g. socket opens but auth fails), the partially-opened connection was not closed, leaking file descriptors
- In the retry loop, each attempt leaked another fd until hitting `ulimit` and crashing with `OSError: [Errno 24] Too many open files`
- Related to 25a1d104 (close #66) which fixed file-level handle leaks on failed transfers — this fix addresses connection-level socket leaks on failed connects

### Fixes at three layers:
- **sftp.py**: close `self.ssh` in `connect()` except block
- **ftp.py**: close local `ftp` object in `connect()` except block (initialize `ftp=None` to avoid `NameError` if constructor throws)
- **flow/__init__.py**: close old proto before creating new `Transfer.factory()` in `send()`, and close proto on failed connect in both `send()` and `download()` paths

Closes #1559

## Test plan
- [x] Unit tests verify `close()` is called on failed SFTP/FTP connections
- [x] Unit tests verify repeated connect failures each clean up (no fd accumulation)
- [x] Unit tests verify successful connections are not prematurely closed
- [x] Existing test suite passes (119 passed, 1 pre-existing failure in scheduled tests)

elated to 25a1d104 -- same category of bug. That one fixed file handle leaks during failed transfers (get/put). This one fixes socket leaks during failed connections (connect/auth). Both cause the "too many open files" crash when they accumulate. Squish-squish.

PR #1620, opened 2026-03-24
Fix FTP connection leak on failed connect

robjarawan:
##### what

If `connect()`, `login()`, or `set_pasv()` raises an exception after the local `ftp` object is created but before `self.ftp = ftp` assignment, the FTP socket is never closed. The bare `except:` swallows the error and the local variable goes out of scope, leaking the socket into TIME_WAIT state.

With retry logic, each failed connect attempt leaks one socket. A box connecting to multiple FTP sources with one flaky remote accumulates hundreds of leaked sockets per day.

##### the fix

- Initialize `ftp = None` before the try block
- Close it in the except block if it was created
- Narrow bare `except:` to `except Exception:` so `KeyboardInterrupt`/`SystemExit` propagate
- Same pattern as the SFTP fixes in PRs #1578, #1581, #1583

##### regression tests

New `tests/sarracenia/transfer/ftp_test.py` with 5 tests:

- `test_connect__closes_ftp_on_login_failure` -- login fails, socket is closed
- `test_connect__closes_ftp_on_connect_failure` -- connect fails, socket is closed
- `test_connect__no_leak_on_set_pasv_failure` -- set_pasv fails after login, socket is closed
- `test_connect__keyboard_interrupt_propagates` -- KeyboardInterrupt not swallowed
- `test_connect__success_does_not_close` -- successful connect assigns ftp, does not close

### DIFF
--- a/sarracenia/transfer/ftp.py
+++ b/sarracenia/transfer/ftp.py
@@ -214,10 +214,11 @@ class Ftp(Transfer):
         # timeout alarm 100 secs to connect
         alarm_set(self.o.timeout)
 
+        ftp = None
         try:
             expire = -999
             if self.o.timeout: expire = self.o.timeout
-            if self.port == '' or self.port == None: 
+            if self.port == '' or self.port == None:
                 if self.implicit_ftps:
                     self.port = 990
                 else:
@@ -255,7 +256,7 @@ class Ftp(Transfer):
 
             try:
                 self.originalDir = ftp.pwd()
-            except:
+            except Exception:
                 logger.warning("Unable to ftp.pwd")
                 logger.debug('Exception details: ', exc_info=True)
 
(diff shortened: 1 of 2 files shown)


petersilva:
This might help with memory leaks for ftp senders...

robjarawan:

This is now rebased onto current `development`, so the merge conflict in `sarracenia/transfer/ftp.py` should be gone. I kept the change focused to the FTP connect leak fix and dropped the old branch drift.

Changes in this refresh:
- close the local FTP object if `connect()`, `login()`, or `set_pasv()` fails before `self.ftp = ftp`
- narrow the connect-path bare `except` to `except Exception`
- keep the touched logger call in lazy-format style
- add focused regression tests for failed connect/login/set_pasv, successful connect, and `KeyboardInterrupt` propagation

Tests:
- `python3 -m pytest tests/sarracenia/transfer/ftp_test.py -v` - 5 passed
- `python3 -m pytest tests/sarracenia/transfer/ftp_test.py tests/sarracenia/transfer/__transfer___test.py tests/sarracenia/transfer/transfer_file_test.py -v` - 10 passed
- `pycodestyle --max-line-length=119 tests/sarracenia/transfer/ftp_test.py` - passed
- `git diff --check -- sarracenia/transfer/ftp.py` - passed

One note: `KeyboardInterrupt` and `SystemExit` still propagate, intentionally. If one of those hits after the temporary FTP object is created but before `self.ftp = ftp`, cleanup would need a broader `try/finally` restructure. I left that as follow-up scope rather than mixing it into this leak fix.

@andreleblanc11 @reidsunderland 🍻

reidsunderland (approved):

PR #1621, opened 2026-03-24
Close transfer protocol connections on flow shutdown

robjarawan:
##### what

`Flow.close()` runs on_stop callbacks but never closes active transfer connections in `self.proto`. On clean daemon shutdown, all SFTP/FTP/HTTP sockets are left for GC and OS cleanup instead of being explicitly closed.

The error paths in `do_download()` (line 2530) and `send()` (line 2983) already close proto on failure and set to None. But the clean shutdown path never does.

##### the fix

Iterate `self.proto` in `close()` and close each active connection before the shutdown log message. Exceptions during close are logged at debug level and do not prevent other connections from being closed.

##### regression tests

New `tests/sarracenia/flow_close_test.py` with 4 tests:

- `test_close__closes_proto_connections` -- both sftp and ftp protos get closed
- `test_close__skips_none_proto` -- already-None entries are skipped
- `test_close__proto_close_exception_doesnt_crash` -- one proto failing to close doesn't block the others
- `test_close__empty_proto` -- no proto connections is fine

### DIFF
--- a/sarracenia/flow/__init__.py
+++ b/sarracenia/flow/__init__.py
@@ -479,6 +479,15 @@ class Flow:
     def close(self) -> None:
 
         self.runCallbacksTime('on_stop')
+
+        for scheme in list(self.proto):
+            if self.proto[scheme] is not None:
+                try:
+                    self.proto[scheme].close()
+                except Exception as err:
+                    logger.debug("proto %s close: %s", scheme, err)
+                self.proto[scheme] = None
+
         if os.path.exists( self.o.novipFilename ):
             os.unlink( self.o.novipFilename )
         logger.info(
--- a/tests/sarracenia/flow_close_test.py
+++ b/tests/sarracenia/flow_close_test.py
@@ -1,0 +1,107 @@
+import pytest
+from tests.conftest import *
+from unittest.mock import patch, MagicMock
+
+import types
+
+import sarracenia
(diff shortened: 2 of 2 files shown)


petersilva:
not sure about this one. 
*  I don't understand the order in which the on_stop callbacks are called.  It is possible (I don't know, just worrying abstractedly)  that another on_stop callback needs those connections to do stop processing.  
* It doesn't really matter if the cleanup doesn't happen in this case because the process dies after on_stop, so OS cleanup is good enough.  This is not solving any problem, and potentially causes some.

petersilva (changes requested):
I do not think this change solves any problem, and it opens up a very small risk of creating new ones... I don't think this PR is worthwhile.

reidsunderland:
The close happens after all `on_stop` callbacks have run, so it seems harmless at first glance. 

I also assume automatic cleanup is good enough, but maybe there is some value in doing a clean close. Maybe a connection left open takes a while to time out and get cleaned up on the server side? 

It would be interesting to do a test with connecting to an FTP and SFTP server. Without this code, does the connection linger on the server for a while after stopping the sr3 subscriber?

robjarawan:
Closing this one. Peter's right that the OS handles cleanup on exit and the risk of breaking an on_stop callback that needs an active connection isn't worth it.

PR #1622, opened 2026-03-24
Fix MQTT client thread leak on reconnect and close

robjarawan:
##### what

Two related thread leak issues in the MQTT transport layer:

1. `close()` calls `client.disconnect()` but never `client.loop_stop()`. The paho background thread keeps running after the connection is closed. Compare with `putCleanUp()` (line 558-561) which correctly calls both.

2. `getSetup()` and `putSetup()` create a new client with `loop_start()` without stopping the old client's thread first. Each reconnection leaks one orphaned paho thread that continues running in the background.

Over time with flaky MQTT brokers that trigger reconnections, this accumulates threads indefinitely.

##### the fix

- Add `loop_stop()` to `close()` after `disconnect()`
- In `getSetup()` and `putSetup()`, call `loop_stop()` + `disconnect()` on the old client before creating the new one
- Old client cleanup is wrapped in try/except since the old client may already be in a bad state

##### regression tests

New `tests/sarracenia/moth/mqtt_thread_leak_test.py` with 5 tests:

- `test_close__calls_loop_stop` -- close() calls both disconnect and loop_stop
- `test_close__loop_stop_after_disconnect` -- correct call order verified
- `test_close__not_connected_skips_loop_stop` -- no-op when not connected
- `test_getSetup__stops_old_client` -- old client stopped before new one created
- `test_putSetup__stops_old_client` -- same for publish path

### DIFF
--- a/sarracenia/moth/mqtt.py
+++ b/sarracenia/moth/mqtt.py
@@ -426,6 +426,12 @@ class MQTT(Moth):
 
             logger.info( f"is no around? {self.o['no']} " )
             if ('no' in self.o) and self.o['no'] > 0: # instances 'started'
+                if hasattr(self, 'client') and self.client is not None:
+                    try:
+                        self.client.loop_stop()
+                        self.client.disconnect()
+                    except Exception:
+                        pass
                 self.client = self.__clientSetup(cid)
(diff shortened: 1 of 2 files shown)


petersilva (approved):
looks good.

PR #1628, opened 2026-03-25
Convert source to use f-strings where it should

petersilva:
close #1623 

## Convert All strings in the app to fstrings (except logger.debug) 

This series of patches converts almost all strings to f-strings as per python best practice recommendations.
The biggest patch is the first one, which represents the output of running *flynt* a program developed specifically to update python strings in source code.

It left out about 80 cases, where it was not sure what to do.  the problems were:

* line longer than 88 characters... it refused to do something that would result in multi-line (was on default settings.)
* sometimes the original strings were plain wrong... like " %s ", arg1, arg2 ) .... mismatch between pattern and string.

I ran all the flow tests on just the flynt output, and it was fine.   Then I modified the rest manually, and the tests were failing.  So I did a kind of binary search within the changes, which is why you get all these parts of the manual patches.   I eventually got to part 6.   When that was done as a multi-line, it was failing.  but the way it is now it works... ran all the flows. fine.


robjarawan (commented):
Went through the diff. Looks good overall -- flynt did the heavy lifting and the manual fixups check out.

A few things I noticed:

##### logger.debug converted to f-strings

8 debug calls got converted from % formatting to f-strings, which goes against the lazy formatting direction from #1623. They're all low-traffic paths (exception handlers, sleep log, chmod debug) so the impact is small, but it's inconsistent:

- `diskqueue.py:163,169,175` -- `logger.debug(f"housekeeping_fp close: {err}")` etc.
- `flow/__init__.py:687` -- sleep duration debug in the main run loop
- `moth/mqtt.py` -- putSetup declaring exchange
- `transfer/file.py`, `transfer/sftp.py` -- chmod debug calls

Should these stay as lazy format or is f-string fine for debug in exception/low-traffic paths?

##### no-op string copies

```python
# before
relPath = '%s' % msg['relPath']
# after
relPath = f"{msg['relPath']}"
```

Both just copy a string. The existing FIXME comment even asks "why the %?". Could just be `relPath = msg['relPath']` but that's a separate cleanup, not a blocker.

##### heads up on merge conflicts

This touches a lot of the same files as a few open PRs (#1618, #1620, #1617). We'll rebase those after this merges, no worries.

Everything else looks correct. The binary-search approach for finding the manual fixup failures was smart.

petersilva:
Those are weird cases, but not show stoppers.  As there are so many changes here, it is important to proceed quickly as this is likely to cause conflicts with other development if not merged quickly.

reidsunderland (approved):
I don't know how to manually review it, if it runs and doesn't crash that's good. And it seems like Rob's AI approves.

PR #1629, opened 2026-03-25
Fix v02 post format crash after shallow copy in putNewMessage

robjarawan:
##### what

PR #1604 changed `putNewMessage()` from `copy.deepcopy(message)` to `dict(message)` for performance. This broke the v02 post format path.

`dict(message)` strips the `sarracenia.Message` class, producing a plain dict. When posting in v02 format, the call chain hits `v2wrapper.Message.__init__()` which calls `h.isRetry()` on the message -- but a plain dict has no `isRetry()` method.

```
AttributeError: 'dict' object has no attribute 'isRetry'
```

Traced by Reid -- the v02 path goes: `putNewMessage` -> `PostFormat.exportAny` -> `v02.exportMine` -> `v2wrapper.Message(body)` -> `h.isRetry()` crash.

##### why tests didn't catch it

The sr_insects flow tests don't exercise v02 post format. All flow tests use v03.

##### the fix

Replace `h.isRetry()` with `h.get('_isRetry', False)` in `v2wrapper.py:156`. The `isRetry()` method literally just checks `'_isRetry' in msg`, so inlining it removes the class dependency. Works with both plain dicts and Message objects.

##### regression tests

New `tests/sarracenia/flowcb/v2wrapper_message_test.py` with 5 tests:

- `test_v2wrapper_message_from_plain_dict` -- plain dict (post-shallow-copy) doesn't crash
- `test_v2wrapper_message_from_plain_dict_with_retry` -- plain dict with _isRetry=True
- `test_v2wrapper_message_from_plain_dict_with_retry_count` -- plain dict with _isRetry as int
- `test_v2wrapper_message_from_sr3_message` -- sarracenia.Message still works
- `test_v2wrapper_message_from_sr3_message_with_retry` -- sarracenia.Message with retry flag

### DIFF
--- a/sarracenia/flowcb/v2wrapper.py
+++ b/sarracenia/flowcb/v2wrapper.py
@@ -153,7 +153,7 @@ class Message:
        
         self.headers = h
         self.hdrstr = str(h)
-        self.isRetry = bool(h.isRetry())
+        self.isRetry = bool(h.get('_isRetry', False))
 
         # from sr_message/sr_new ...
         self.local_offset = 0
--- a/tests/sarracenia/flowcb/v2wrapper_message_test.py
+++ b/tests/sarracenia/flowcb/v2wrapper_message_test.py
@@ -1,0 +1,81 @@
+import pytest
+from tests.conftest import *
+
+import sarracenia
+from sarracenia.flowcb.v2wrapper import Message
+
+
+def make_plain_dict_message():
+    """Create a plain dict (not sarracenia.Message) with minimum v02 fields.
+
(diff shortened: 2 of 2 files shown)


petersilva:
I prefer #1630 for now... it seem less intrusive... There is a reason the Message class exists... it might be ok to strip it off at some point, but for now the copy.copy() thing seems like the easiest approach.

robjarawan:
Closing in favor of #1630 (copy.copy) which Reid already merged. The v2wrapper hardening is nice-to-have but not needed now that the Message class is preserved through the copy.

PR #1630, opened 2026-03-25
Use `copy.copy` for shallow copying

reidsunderland:
copy.copy will produce a sarracenia.Message object.

The `dict()` code gives a dict. Message is a subclass of dict, so a Message can be used where a dict is expected, but a dict can't necessarily be used where a Message is expected.

I think this is the safest fix.

### DIFF
--- a/sarracenia/moth/amqp.py
+++ b/sarracenia/moth/amqp.py
@@ -688,7 +688,8 @@ class AMQP(Moth):
                 return False
 
         # Shallow copy: only top-level keys are deleted (_deleteOnPost), nested dicts are read-only
-        body = dict(message)
+        # copy.copy(message) produces a sarracenia.Message object
+        body = copy.copy(message)
 
         if 'format' in self.o:
             version=self.o['format']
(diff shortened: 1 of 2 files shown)


reidsunderland:
The performance of this is worse than dict(), but still 10x better than deepcopy:

```
dict(message):          0.029s
copy.copy(message):     0.331s
copy.deepcopy(message): 3.572s
```

```python
        # Shallow copy: only top-level keys are deleted (_deleteOnPost), nested dicts are read-only
        # copy.copy(message) produces a sarracenia.Message object
        try:
            import timeit
            import copy
            
            n = 100000
            t1 = timeit.timeit('dict(message)',          number=n, globals=locals())
            t2 = timeit.timeit('copy.copy(message)',      number=n, globals=locals())
            t3 = timeit.timeit('copy.deepcopy(message)',  number=n, globals=locals())

            logger.error(f"dict(message):          {t1:.3f}s")
            logger.error(f"copy.copy(message):     {t2:.3f}s")
            logger.error(f"copy.deepcopy(message): {t3:.3f}s")
        except Exception as e:
            logger.error("", exc_info=True)
            exit()
```

reidsunderland:
Rob also proposed https://github.com/MetPX/sarracenia/pull/1629 which fixes the crash that we ran into and might be viable. But we'd have to be careful to ensure that PostFormat implementations don't use any Message helper methods.

petersilva (approved):

petersilva:
working beats faster but broken  every time.

PR #1631, opened 2026-03-26
Fix post_format not derived from post_topicPrefix

robjarawan:
##### what

Setting `post_topicPrefix v02.post` in a config should derive `post_format v02`, but it doesn't. Peter found this while investigating the v02 regression (sr_insects#85) -- the format derivation has been broken since at least 3.00.56.

##### root cause

In `config/__init__.py`, `post_format` defaults to `'v03'`. In `publisher.py`, the logic is:

```python
if hasattr(options, 'post_format'):      # always True -- default is 'v03'
    self['format'] = options.post_format
elif hasattr(options, 'post_topicPrefix') ...   # never reached
```

Since `post_format` always has a value, the `elif` branch that derives format from `post_topicPrefix` is dead code.

##### the fix

- Change `post_format` default from `'v03'` to `None`
- Check for `None` in `publisher.py` so the topicPrefix derivation kicks in when the user hasn't explicitly set `post_format`
- Explicit `post_format` still takes priority when set by the user

##### regression tests

New `tests/sarracenia/config/publisher_test.py` with 5 tests:

- `test_format_defaults_to_v03` -- neither post_format nor topicPrefix set
- `test_format_derived_from_v02_topicprefix` -- the bug case
- `test_format_derived_from_v03_topicprefix` -- v03 topicPrefix still works
- `test_explicit_post_format_overrides_topicprefix` -- user sets post_format, wins over topicPrefix
- `test_explicit_v02_post_format` -- user sets post_format v02 directly

##### context

Peter's investigation in sr_insects#85 and sarracenia#1623. The sr_insects sender config `tsource2send_f50.conf` has `post_topic_prefix v02.post` but was silently posting v03 because format was never derived.

### DIFF
--- a/sarracenia/config/__init__.py
+++ b/sarracenia/config/__init__.py
@@ -113,7 +113,7 @@ default_options = {
     'post_documentRoot': None,
     'post_baseDir': None,
     'post_baseUrl': None,
-    'post_format': 'v03',
+    'post_format': None,
     'realpathPost': False,
     'recursive' : True,
     'runStateThreshold_reject': 80,
--- a/sarracenia/config/publisher.py
+++ b/sarracenia/config/publisher.py
@@ -63,9 +63,10 @@ class Publisher(dict):
         if hasattr(options,'tlsRigour') :
             self['tlsRigour'] = options.tlsRigour
     
-        if hasattr(options,'post_format') :
+        if hasattr(options, 'post_format') and options.post_format is not None:
             self['format'] = options.post_format
-        elif hasattr(options,'post_topicPrefix') and options.post_topicPrefix[0] in [ 'v02', 'v03' ]:
+        elif hasattr(options, 'post_topicPrefix') and options.post_topicPrefix \
+                and options.post_topicPrefix[0] in ['v02', 'v03']:
             self['format'] = options.post_topicPrefix[0]
         else:
             self['format'] = 'v03'
--- a/sarracenia/config/publisher.py
+++ b/sarracenia/config/publisher.py
@@ -71,6 +71,9 @@ class Publisher(dict):
         else:
             self['format'] = 'v03'
 
+        # exportAny and exportMine expect 'post_format' in the options dict
+        self['post_format'] = self['format']
+
         if hasattr(options,'post_topicPrefix') and options.post_topicPrefix:
             self['topicPrefix'] = options.post_topicPrefix
         elif hasattr(options, 'topicPrefix') and options.topicPrefix:
--- a/sarracenia/moth/__init__.py
+++ b/sarracenia/moth/__init__.py
@@ -271,6 +271,8 @@ class Moth():
             props['broker'] = broker
             if 'exchange' in publisher:
                 props['exchange'] = publisher['exchange']
+            if 'format' in publisher:
+                props['format'] = publisher['format']
 
         elif not props['broker']:
             logger.error('no broker specified')
@@ -347,6 +349,8 @@ class Moth():
                     self.o['exchange'] = publisher['exchange']
                 if 'topicPrefix' in publisher:
                     self.o['topicPrefix'] = publisher['topicPrefix']
+                if 'format' in publisher:
+                    self.o['format'] = publisher['format']
 
         # apply settings from props.
         if 'settings' in self.o:
--- a/tests/sarracenia/config/publisher_test.py
+++ b/tests/sarracenia/config/publisher_test.py
@@ -74,3 +74,32 @@ def test_explicit_v02_post_format():
     pub = Publisher(opts)
     assert pub['format'] == 'v02'
     assert pub['post_format'] == 'v02'
+
+
+def test_topicprefix_fallback_is_empty_list():
+    """When neither post_topicPrefix nor topicPrefix is set, fallback must
+    be [] (empty list), not None. Peter fixed this in bab4b9424 -- None
+    causes TypeError when downstream code iterates or concatenates."""
+    opts = make_options()
+    del opts.post_topicPrefix
+    del opts.topicPrefix
+    pub = Publisher(opts)
+    assert pub['topicPrefix'] == [], \
+        "topicPrefix fallback must be [] not None (see commit bab4b9424)"
+
+
+def test_basedir_missing_no_keyerror():
+    """Publisher must not raise KeyError when baseDir is absent from the
+    dict. The guard must use 'or' (short-circuit) not 'and'."""
+    opts = make_options(post_baseUrl='file:/data/incoming')
+    del opts.post_baseDir
+    # This must not raise KeyError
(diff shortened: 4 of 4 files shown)


petersilva:
The above patch fixes the settings so the publisher setting of "format" is set correctly.  So far so good.  I don't think the moth/  code takes the "format" into account to actually produce messages in the requested format.   I looked at moth/__init__.py export any, and it expects a setting "post_format" to be there... where the setting is called "format" and this value is not being transferred from the publisher... so to actually get v02 messages... I think a bit more tweaking is needed.

petersilva (changes requested):
I'm tempted to approve, but it is not doing the whole job... yes it fixes the post_format setting, gets it properly set in the publisher.

I don't think the value in the publisher is getting properly passed to postformat/__init__.py() ExportAny ... so probably need another patch to fix that... is it another branch or just more to this one?

robjarawan:
Added `post_format` to the publisher options dict so it's available when passed through to `PostFormat.exportAny()` and `exportMine()`. Should cover the gap you spotted -- `post_format` now stays in sync with `format` in the publisher.

petersilva:
So the last patch makes publishers explicitly use *post_format*.

so I dunno... in general in moth when you have a post_ setting, the post_ gets removed in publishers and becomes the same setting without the post_ prefix... This is what is done for post_broker, post_exchange*, post_topicPrefix, etc... so... if we were consistent with other settings, the one used here would be *format*... 

but I don't know if this convention makes much sense or is worth following, so maybe just explicitly using *post_format* is fine.  Opinions?

robjarawan:
@petersilva I think sticking with `post_format` explicitly is the right call here. The `post_` stripping convention makes sense for settings like `post_broker` and `post_exchange` where the publisher naturally operates on the "post" side, so dropping the prefix is intuitive. But `format` on its own is too generic and could easily be confused with something unrelated. Keeping `post_format` makes the intent clear and avoids ambiguity without breaking anything. What do you think @reidsunderland?

robjarawan:
### Code review

Found 2 issues:

1. `baseDir` guard changed from `or` to `and`, which will raise `KeyError` when `baseDir` is not in the dict. The `and` operator evaluates both sides, so `not self['baseDir']` runs even when the key is absent. The original `or` short-circuits correctly.

https://github.com/MetPX/sarracenia/blob/390425b2f0e584929d966778e85456aeac371c31/sarracenia/config/publisher.py#L84-L86

2. `topicPrefix` fallback changed from `[]` to `None`, reversing commit bab4b9424 ("make topicPrefix and post_topicPrefix None a valid choice") which deliberately set this to `[]`. Combined with the unconditional `self.o['topicPrefix'] = publisher['topicPrefix']` in `moth/__init__.py` (guard removed in this PR), `None` will clobber the `['v03']` default from `default_options()` and cause `TypeError` when downstream code (e.g. `postformat`) tries to iterate or concatenate topicPrefix.

https://github.com/MetPX/sarracenia/blob/390425b2f0e584929d966778e85456aeac371c31/sarracenia/config/publisher.py#L74-L76

Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with a thumbs up. Otherwise, react with a thumbs down.</sub>

robjarawan:
> ### Code review
> Found 2 issues:
> 
> 1. `baseDir` guard changed from `or` to `and`, which will raise `KeyError` when `baseDir` is not in the dict. The `and` operator evaluates both sides, so `not self['baseDir']` runs even when the key is absent. The original `or` short-circuits correctly.
> 
> https://github.com/MetPX/sarracenia/blob/390425b2f0e584929d966778e85456aeac371c31/sarracenia/config/publisher.py#L84-L86
> 
> 2. `topicPrefix` fallback changed from `[]` to `None`, reversing commit [bab4b94](https://github.com/MetPX/sarracenia/commit/bab4b942461eba37e0b2f38fa930663e9a9e100f) ("make topicPrefix and post_topicPrefix None a valid choice") which deliberately set this to `[]`. Combined with the unconditional `self.o['topicPrefix'] = publisher['topicPrefix']` in `moth/__init__.py` (guard removed in this PR), `None` will clobber the `['v03']` default from `default_options()` and cause `TypeError` when downstream code (e.g. `postformat`) tries to iterate or concatenate topicPrefix.
> 
> https://github.com/MetPX/sarracenia/blob/390425b2f0e584929d966778e85456aeac371c31/sarracenia/config/publisher.py#L74-L76
> 
> Generated with [Claude Code](https://claude.ai/code)
> 
> - If this code review was useful, please react with a thumbs up. Otherwise, react with a thumbs down.

@andreleblanc11 which part is incorrect?

reidsunderland:
```
>>> self = dict()

# OLD code is wrong
>>> not 'baseDir' in self and not self['baseDir']
KeyError: 'baseDir'

# NEW code proposed in this PR is good
>>> not 'baseDir' in self or not self['baseDir']
True
```

The `if` _should_ have **or**. The proposed code change for that in this PR is correct:
image

Claude is saying _"baseDir guard changed from or to and"_ which is the opposite of reality.

robjarawan:
Retracting the code review comment above -- both issues were wrong.

I was looking at an intermediate commit (`390425b2`) rather than the branch HEAD. By the time that comment was posted, both the `baseDir` guard and the `topicPrefix` fallback had already been corrected in `029bf1b9c`. The net diff from development to this branch HEAD is clean on both counts.

Thanks guys.

robjarawan:
> ```
> >>> self = dict()
> 
> # OLD code is wrong
> >>> not 'baseDir' in self and not self['baseDir']
> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
> KeyError: 'baseDir'
> 
> # NEW code proposed in this PR is good
> >>> not 'baseDir' in self or not self['baseDir']
> True
> ```
> 
> The `if` _should_ have **or**. The proposed code change for that in this PR is correct: image
> 
> Claude is saying _"baseDir guard changed from or to and"_ which is the opposite of reality.

My reply to Andre: 

Andre, do not question Ultron.
The machine was correct. Reality was misconfigured.
Your confession has been logged and your access to boolean operators will remain under observation.

robjarawan:
one last review @petersilva @andreleblanc11 @reidsunderland

robjarawan:
The derivation intent is right, but I found a real blocker: the global default flip to `None` breaks inline embedding in `gather/file.py:172`, which tests `post_format == 'v03'`. Every default config would silently stop embedding content. My hold condition is a one-line guard accepting `('v03', None)`, since `Publisher` needs `None` as its unset sentinel and resolving at finalize would re-break the derivation. Plus a named test (default plus inline gives content, `v02` does not) and a fix to the stale docs line claiming the default is `v03`. No re-review needed if the diff stays limited to that.

PR #1633, opened 2026-03-26
Fix: create destination directory before writing config file in sr3 add

aditi123-hub:
### Problem
Running:
sr3 add subscriber/dual_amis.conf

caused FileNotFoundError because the destination directory did not exist.

### Solution
Ensure directory is created before writing file using:
os.makedirs(destdir, exist_ok=True)

### Testing
Tested by running:
sr3 add subscriber/dual_amis.conf

File is now created successfully without errors.

### DIFF
--- a/sarracenia/sr.py
+++ b/sarracenia/sr.py
@@ -1471,6 +1471,7 @@ class sr_GlobalState:
                 logger.info(f"did not find anything to copy for: {l}. creating an empty one.")
                 if cfg[-5:] not in [ '.inc', '.conf' ]:
                     cfg = cfg + '.conf'
+                os.makedirs(destdir, exist_ok=True)
                 with open( destdir + os.sep + cfg, 'w' ) as f:
                     f.write('')
 


petersilva:
a component is a defined thing in sr3... you don't just create new ones ad-hoc.  This PR is wrong-headed.  no... mkdir does not fix the problem, but rather makes it worse.

petersilva:
The original issue described the problem as being about giving a better error message.   The suggested action hides the fact that incorrect data was entered, as much of the code uses that string to set other parameters for operation, having an incorrect value will cause many failures in practice.   The solution should be about catching the error and explaining the defect.  Do not change the behaviour of the code.

Since the entire PR just changes behaviour, it is entirely inappropriate.  a correct PR will have entirely different content.

aditi123-hub:
Thanks for the clarification on the previous PR! I've updated the approach to avoid changing behavior and instead improved the error message for invalid configuration paths. Please let me know if this aligns better with the expected solution.

PR #1635, opened 2026-03-26
Fix issue #1573: filtering using sundew extension

reidsunderland:
the French documentation is from Google, if there are any issues please edit on this branch (or tell me what needs fixing)

Include sundew_extension in the filtering, same behaviour as v2, slightly different implementation because we have the full URL here and v2 was only dealing with the URL path.


petersilva:
sundew_extensions... they have a fixed number of colon separated fields 5? 6? something like that.
The problem here is detecting whether an sundew_extension is already in the filename or not.

If a file arrives and it already has A sundew_extension, but not THE sundew_extension.... this will just append.

I don't think checking for the whole string is a good idea...  What if the file has some other sundew_extension with lots of colons in it already... if you're just checking for yours...
I think we'll end up with lots of files that have multiple sundew_extensions.

I think upping the number allowed to around two (one for scheme, one for port number) is going to work 99%
of the time when we want it to.

petersilva:
I'm not going to block this if you really want it... but I think it is going to cause some ... interesting behaviour ... when it gets into production.   some configurations receive files with existing sundew_extensions, and the setting is meant to "fill in the blanks" ... this will fill in regardless of whether there is a blank or not...

reidsunderland:
that's a good point, I didn't think about cases where the sundew_extension header is present but different from a sundew extension already in the filename

reidsunderland:
Switched back to checking the count of colons, except using `<3` instead of `<1` because sr3 uses the full URL (including scheme:// and possibly :port). v2 only used the path, so `<1` worked.

petersilva (approved):
seems reasonable now.

PR #1639, opened 2026-03-27
Fix: improve error message for invalid configuration path

aditi123-hub:
Fix: Improve error message for invalid configuration path

Problem:
The existing error when an invalid configuration path is provided is not very clear.

Solution:
Improved the error message to clearly indicate that the component may not exist or the path is incorrect.

This change does not modify any existing behavior and only improves error reporting, as suggested in the issue discussion.

### DIFF
--- a/sarracenia/sr.py
+++ b/sarracenia/sr.py
@@ -1462,7 +1462,6 @@ class sr_GlobalState:
             found = False
             for candidate in suggestions:
                 if os.path.exists(candidate):
-                    pathlib.Path(destdir).mkdir(parents=True, exist_ok=True)
                     logger.info(f"copying: {candidate} to {destdir + os.sep + cfg} ")
                     shutil.copyfile(candidate, destdir + os.sep + cfg)
                     found = True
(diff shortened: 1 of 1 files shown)


petersilva (changes requested):
idefix% sr3 add subscriber/lala-123
add: 2026-03-27 10:14:55,919 2563288 [INFO] sarracenia.sr add did not find anything to copy for: subscriber/lala-123. creating an empty one.
2026-03-27 10:14:55,919 2563288 [ERROR] sarracenia.sr add Invalid configuration path: /home/peter/.config/sr3/flow. Component 'flow' may not exist or the path is incorrect.

idefix% 

*flow* is not a component name.

petersilva:
I love AI tools. We use AI tools. AI is a tool. People have to use their judgement to understand whether the instructions they have given to the AI are having the desired result.  It looks like you don't know sr3 well enough to understand basic concepts of it, so when the AI gives you a result, you are not in a position to give it proper feedback.
Understand we can spend a lot of time correcting other peoples' mis-specified AI prompts.

Please do not use AI as an excuse to waste peoples' time.  You still need to invest the time to understand what a given project is doing and make sure that your own contribution is helping people, and the project, rather than putting more load on them.  Contribution does require a certain amount of commitment to a project.

PR #1640, opened 2026-03-27
Fix unit tests

reidsunderland:
After the string formatting changes, some unit tests were failing. 

❗ One of the issues that got discovered is a bit concerning - the `d` format code in f-strings only works with **integers**, but in % formatting it will also accept a float (it converts the float to int):

```
>>> x = 123.456
>>> "%d" % x
'123'
>>> f"{x:d}"
ValueError: Unknown format code 'd' for object of type 'float'
```

I'm worried that there could be other cases of this that we haven't hit.

### DIFF
--- a/sarracenia/flowcb/nodupe/redis.py
+++ b/sarracenia/flowcb/nodupe/redis.py
@@ -134,7 +134,7 @@ class Redis(NoDupe):
         new_count = len(self._redis.keys(self._rkey_base + ":*"))
         self.now = nowflt()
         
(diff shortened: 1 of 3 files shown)


petersilva (approved):

petersilva:
Before I opened the PR, I ran all the flow tests on my 26.04 workstation... It's weird that they passed here... but not for you/github.

reidsunderland:
That is really weird. Do they fail now on your workstation? The log change in the speedo test should have at least failed on your local run (or should fail now)

PR #1641, opened 2026-03-28
fix azure test by pinning older version of azure-storage-blob

reidsunderland:
required until MS updates Azurite to work with the latest API version used by the latest azure-storage-blob package


petersilva (approved):
pain in the hind-quarters, but apparently needed for now. Great catch!

PR #1643, opened 2026-03-28
second pass at code base with flynt

petersilva:
flynt is an f-string formatting linter... that converts old-style python strings to f-strings.
This is preferred for strings when they are not in logger.debug() invocations (and flynt is aware of that.)
A previous PR took care of many (large majority) of such changes.

After many parallel changes it seems more old stle strings present. doing a second pass. for #1623

I think the main difference is this time, I told it about the line length convention:
```

idefix% find . -name '*.py' | xargs flynt -ll 119 -v 
...

```
It skips lines that will be too long (default > 88) if reformatted... Once I gave it the longer
line length, it decided to do more transformations.

### DIFF
--- a/sarracenia/examples/moth_http_retrieval.py
+++ b/sarracenia/examples/moth_http_retrieval.py
@@ -99,8 +99,7 @@ while count < 10:
                 elif name == 'air_temp':
                     air_temp = i.get('value')
 
-            print('station: %s, tc_id: %s, lat: %s, long: %s, air_temp: %s' %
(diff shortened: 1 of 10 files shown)


robjarawan (approved):
Looks good — purely mechanical flynt conversions, no logic changes. Unit tests pass, integration failures are the usual flakey infra tests (pass on re-run).

One minor note: two `%-10s` lines remain in `sr.py` `overview()` (~line 3165) that flynt skipped, probably because the f-string version would exceed 119 chars. Not a blocker, just inconsistent.

PR #1645, opened 2026-03-30
Fix file handle leak in file:// transfer get()

robjarawan:
#### What

- `File.get()` never called `local_read_close(src)`, leaking one file descriptor per file:// download
- In a long-running process transferring thousands of files, this deterministically exhausts fd limits (`OSError: [Errno 24] Too many open files`)
- Wrapped `read_write` in `try/finally` so both `src` and `dst` are closed on success and on exception

#### test plan

- [x] `test_get_closes_src_file_handle` -- basic download, content verified
- [x] `test_get_closes_handles_on_read_error` -- both handles closed when `read_write` raises
- [x] `test_get_no_fd_leak_over_many_transfers` -- 200 consecutive transfers, asserts fd count stays stable

### DIFF
--- a/sarracenia/transfer/file.py
+++ b/sarracenia/transfer/file.py
@@ -127,14 +127,15 @@ class File(Transfer):
         src = self.local_read_open(remote_path, remote_offset)
         dst = self.local_write_open(local_file, local_offset)
 
-        # initialize sumalgo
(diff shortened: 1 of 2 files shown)


petersilva (approved):

petersilva (approved):
Hard to believe it was this badly broken... but I can't find calls to close anywhere... I think it's true.

PR #1646, opened 2026-03-30
Fix gather/message reconnect writing to local variable

robjarawan:
##### What

`gather()` assigned the reconnected Moth consumer to the loop variable `c` instead of `self.consumers[i]`. The consumer list was never updated, so after any broker disconnect the broken consumer stayed in the list permanently. Every subsequent `gather()` cycle created a new Moth object (with sockets/threads) that was immediately discarded -- leaking resources and never recovering message flow.

##### The fix

`self.consumers[i] = sarracenia.moth.Moth.subFactory(od)` instead of `c = ...`. Also switched to `enumerate()` to eliminate the manual index counter.


petersilva:
there are three different PR's mixed together here... there are three commits from three different PR's.
Would like to see a PR with only the described correction in it.

petersilva (approved):

petersilva:
So the thing here is that we are iterating through self.consumers, and setting a variable inside the loop.
I know that in many cases, python does the wrong thing when modifying the thing being iterated over inside the loop itself.  The only flow test that will confirm this works is the multi-static.

Would want to see a good run of that test.

PR #1647, opened 2026-03-30
Fix RecursionError crash in diskqueue on corrupted files

robjarawan:
##### What

`msg_get_from_file()` used recursion to skip corrupted (non-JSON) lines in the retry queue file. A queue file with >1000 consecutive bad lines crashes the process with `RecursionError`. This happens after kill -9, disk full during write, or any partial write corruption.

##### The fix

Replace the recursive call with a `while True` loop that skips corrupted lines iteratively. Can now handle arbitrarily large corrupted files without crashing.

### DIFF
--- a/sarracenia/diskqueue.py
+++ b/sarracenia/diskqueue.py
@@ -348,19 +348,19 @@ class DiskQueue():
             logger.debug('DEBUG %s open read', path)
             fp = open(path, 'r')
 
-        line = fp.readline()
-        if not line:
-            try:
-                fp.close()
-            except Exception:
-                pass
-            return None, None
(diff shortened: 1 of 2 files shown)


petersilva:
similarly, there are commits from multiple other PR's here... please issue PR with only the decribed changes in it.

robjarawan:
Fixed — rebased all three PRs (#1645, #1646, #1647) so each contains only its own commit. Sorry about the noise.

petersilva (approved):

petersilva (approved):
The recursion here always bugged me, this is a good thing.

PR #1648, opened 2026-03-30
Fix AMQP connection and channel leak on reconnect

robjarawan:
##### What

`__connect()` overwrote `self.connection` without closing the previous one, leaking the old AMQP connection and both channels on every reconnect cycle. Under broker flaps this leads to fd exhaustion (#1559).

`putSetup()` exception handler set `self.connection = None` *before* calling `self.close()`, making the close a no-op. `getSetup()` error paths used bare `self.connection = None` instead of `self.close()`, skipping proper teardown.

##### Change

1. Close existing connection in `__connect()` before creating a new one
2. Replace `self.connection = None` with `self.close()` in all `getSetup`/`putSetup` error paths
3. Remove the null-before-close sequence in `putSetup` exception handler

Partial fix for #1559 (AMQP fd leak path; SFTP/FTP paths in #1578, #1581, #1583, #1620).


petersilva (approved):

petersilva (commented):
So this is actually weird... You replace
```
   self.connection = None
```
with:  self.connection.close() ... so after the close it isn't None...

Then in the routing to re-establish, you check... *if self.connection* and then close() again.

Why twice? It looks like we will be closing twice every time, because you never set it to None after your close() ...  What is the truth value of *self.connection* ? I think a closed one is still True.

To avoid closing twice every time, you need to set to None after every close... or change the test in the  connection routine to something that won't evaluate to true when the connection is closed... something like: *if self.connection or not self.connection.is_connected()*

Why is it setting self.connection to None only in that stanza? Either it's good everywhere or nowhere... taking it out sometimes, and putting it back others... I don't understand it.

petersilva:
I doubt it breaks anything, but again, being a foreign branch, would be good to update, and run all the relevant tests to make sure.

PR #1649, opened 2026-03-30
Fix RedisQueue lock not released on early return in on_housekeeping

robjarawan:
##### What

`on_housekeeping` acquires `redis_lock` then checks if the retry list still has entries. If it does, it returns early without releasing the lock. Every subsequent housekeeping cycle deadlocks permanently.

##### Change

Release `redis_lock` before the early return at line 317.


petersilva (approved):

petersilva:
None of the redis flow tests ever pass.  I think we should have some redis tests passing before making changes.

PR #1651, opened 2026-03-30
Fix MQTT ack() silently succeeding on broker mismatch

robjarawan:
##### What

MQTT `ack()` skipped the actual broker ack when `ack_id['broker']` didn't match the current broker (e.g. after failover), but still returned `True`. Message stays unacknowledged — silent message loss.

##### Change

Log a warning and return `False` on broker mismatch, matching how `amqp.py` handles the same case.

### DIFF
--- a/sarracenia/moth/mqtt.py
+++ b/sarracenia/moth/mqtt.py
@@ -713,12 +713,19 @@ class MQTT(Moth):
 
     def ack(self, m: sarracenia.Message ) -> bool:
 
-        if 'ack_id' in m:
-            logger.info( f"mid={m['ack_id']}")
-            if m['ack_id']['broker'] == self.broker:
-                self.client.ack( m['ack_id']['delivery_tag'], m['qos'] )
-                del m['ack_id']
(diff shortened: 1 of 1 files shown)


petersilva (approved):
yeah, that's a real fix. nice catch.

petersilva:
> ##### What
> MQTT `ack()` skipped the actual broker ack when `ack_id['broker']` didn't match the current broker (e.g. after failover), but still returned `True`. Message stays unacknowledged — silent message loss.
> 
> ##### Change
> Log a warning and return `False` on broker mismatch, matching how `amqp.py` handles the same case.


If you look at the code that calls ack() in moth, you will see:

https://github.com/MetPX/sarracenia/blob/ee9d62eaf157bf00cb7954df8a3229e89cd1390f/sarracenia/flowcb/gather/message.py#L62-L70

so the broker is checked before ever calling moth.  All this *fix* does is check it a second time.
I approved it before realizing this... 

I wonder if it is a good thing anyways assuming moth is called by some others (like python access)
instead of only within sr3.

PR #1652, opened 2026-03-30
Close inherited file descriptors before os.execl restart

robjarawan:
##### What

`os.execl` in `resources.py` replaces the process without running any cleanup. Open AMQP/MQTT connections, transfer sockets, and file handles all leak into the new process image. Over repeated OOM restarts this leads to fd exhaustion.

##### Change

Flush stdout/stderr and `os.closerange(3, max_fd)` before `execl` so the new process starts clean.


petersilva (approved):
probably a good idea.

andreleblanc11:
I can test this

andreleblanc11:
I tested with a config with `MemoryMax 1` so the OOM restart would get triggered every housekeeping interval. Went through a couple of OOM restarts and didn't find any problems. So probably safe to merge.

PR #1653, opened 2026-03-30
Remove SIGTERM handler override in AM gather and send plugins

robjarawan:
##### What

`gather/am.py` and `send/am.py` both override SIGTERM with `SIG_DFL` in `__init__`. This kills the process immediately on `sr3 stop`/`restart`, bypassing the flow's graceful shutdown — `on_stop` callbacks, message acking, and connection cleanup are all skipped. Data loss on every stop/restart cycle.

##### Change

Remove the `signal.signal(signal.SIGTERM, signal.SIG_DFL)` call from both files so the parent flow's handler runs properly. Removed unused `signal` import.


petersilva:
I'm puzzled about how to test this... I'm not sure it has the effect hoped for.  would want some testing.

robjarawan:
Fair point. The change is small -- just removing the SIG_DFL override so the parent flow's handler runs instead. The expected effect is that cleanup (closing sockets, stopping loops) actually happens on SIGTERM instead of an abrupt exit. I don't have an AM setup to test against but I can verify that the parent signal handler fires correctly with a mock. Would that be enough, or do you want to see it tested against an actual AM feed?

petersilva:
Maybe @andreleblanc11 has some configs he can test this branch with? you need some AM flows, right?

petersilva:
marking low priority because AM traffic is (gradually) being de-commissioned. so... yeah fine to fix it, but not super urgent.

andreleblanc11:
We added the signal handler to the AM code before Reid added it to the `sr.py` in https://github.com/MetPX/sarracenia/pull/559/changes.

So I think it's probably OK to remove but will need to test.

robjarawan:
The diff itself is strictly safer: it only deletes the two `SIG_DFL` overrides plus imports, restoring the parent flow's graceful handler with zero logic change.

One merge condition: a mock-handler smoke test, since bare fixtures do not construct (`Am` needs `sendTo`, `fileSizeMax`, `add_option`, and a stubbed socket). I verified the corrected test fails pre-PR on both modules and passes post-PR. Happy to push it on request.

The deeper items I found, the infinite accept loop with no stop check and the fork/execl orphan window, are pre-existing and I am marking them wontfix. AM is being decommissioned, so no loop rework is justified for a 6-line cleanup. @andreleblanc11 a live-flow eyeball on your side would still be welcome before merge.

PR #1654, opened 2026-03-30
Close HTTP response objects to prevent socket leak

robjarawan:
##### What

`self.http` (urllib response) is never closed — `__open__`, `close()`, and `init()` all just set it to `None`, orphaning the underlying socket. One socket leaked per HTTP request; under sustained polling this adds up.

##### Change

Call `self.http.close()` before reassigning in `__open__()` and in `init()`.

### DIFF
--- a/sarracenia/transfer/https.py
+++ b/sarracenia/transfer/https.py
@@ -251,6 +251,11 @@ class Https(Transfer):
 
         logger.debug("sr_http init")
         self.connected = False
+        if hasattr(self, 'http') and self.http is not None:
+            try:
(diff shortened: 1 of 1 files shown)


petersilva (approved):
looks harmless.

reidsunderland:
The parent transfer class closes open files (the self.http object is treated as a file) so this probably wasn't causing issues in practice. But I agree it looks harmless and should help ensure nothing gets left open by accident. https://github.com/MetPX/sarracenia/blob/development/sarracenia/transfer/__init__.py

PR #1655, opened 2026-03-30
Fix MQTT rx_msg thread-safety: lock callback append, remove sleep

robjarawan:
##### What

The paho callback thread (`__sub_on_message`) appended to `rx_msg[iFromBroker]` without holding `rx_msg_mutex`. The main thread's `_rotateInputBuffers` could swap indices mid-append, causing both threads to access the same buffer concurrently — lost messages or `IndexError` under load.

Also, `_rotateInputBuffers` held the mutex for 100ms (`time.sleep(0.1)` inside the lock), blocking every callback during rotation.

`getNewMessage` checked `len(self.rx_msg)` (the dict of all 5 buffers, always > 0) instead of `len(self.rx_msg[self.rx_msg_iToApp])` — would `IndexError` on an empty buffer.

##### Change

- Acquire `rx_msg_mutex` in `__sub_on_message` before append (held for ~1 list append)
- Use `with` for mutex in `_rotateInputBuffers`, remove `time.sleep(0.1)` inside lock
- Fix `getNewMessage` to check the correct buffer length


petersilva (commented):
unsure about this one. Want to see more mqtt testing.   I merged developement to this branch so that the multi- tests should pass.  Want to look at mqtt oriented flow tests pass.

petersilva:
@robjarawan this is an example of a branch I'd like to see the MQTT flow tests run on.  since it is a foreign branch, we can't invoke it within the project.  Would be great if you could run them in your repo.

reidsunderland:
Could this code be simplified by using a [queue.Queue](https://docs.python.org/3/library/queue.html#queue.Queue) ? They are thread safe and seem to be working well in the AMQP consumer and AMQP1.0 implementations. At first glance it seems easier to deal with but maybe there was a good reason for not using it here.

robjarawan:
Good call -- queue.Queue would simplify this and get rid of the manual buffer rotation entirely. It's already working well in amqpconsumer and amq1 for the same producer/consumer pattern. The current PR fixes the immediate thread-safety issues but I can put together a follow-up to refactor MQTT's rx_msg to use queue.Queue.

petersilva:
*why not use queue.Queue* ... I think I looked at it a bit... it was a while ago... I decided against at the time... it could easily have been a mistake.  Reasons I can make up that might have had something to do with it:

1. with the queue.Queue stuff you get a mutex or lock, per item queued or removed.  with the mutex thing, you can do group operations (removing or adding multiple items with only one lock.)

2.  if stuff gets acked out of order... can you remove from the Queue out of order.

Those are the reasons that come to mind.  If you look at newMessages, it extracts a whole batch of messages at once, but it doesn't acquire any lock... so I'm confused.  I don't remember why it is this way.  I think there's some kind of alternation going on, where the lock is on one side, and we grab from the other.  so newMessages doesn't need a lock because it's alone on that part of the message queue.

petersilva:
yeah ... basing it on queue.Queue is a lot easier to understand for sure... but there still isn't any proof it passes MQTT flow tests (or the multi-static)

petersilva:
oh... and the vulnerablity alluded to isn't there... because there are only two rotations, but five queues... 
the mutex is only needed to rotate among the five active queues. the reader is always working on another queue.
The whole premise of the original vulnerability is wrong, and does not happen in reality.

but still using a single queue.Queue is simpler to read, and should work fine, though probably slower... but impactfully so? dunno.

petersilva:
It should be noted that a big problem with mqtt stuff is delay in the on_message() routine... it causes messages to be dropped.  You can't have expensive on_message... and blocking for a lock would be an example of a bad thing.

robjarawan:
Good points on the 5-buffer design -- you're right that with separate read/write buffers, newMessages doesn't need a lock since it's always working on a different buffer than the callback. The thread-safety concern in this PR may have been overstated.

That said, queue.Queue is a lot easier to follow and the per-item lock overhead should be negligible for the message rates we're dealing with -- each lock/unlock is a few microseconds, and we're bounded by network I/O anyway. If it does show up in profiling later we can revisit.

MQTT flow tests are running on my fork now, will post results when they're done.

petersilva:
OK, I understand now... the AI is making the mistaken assumption that there are multipls callers /tasks calling newMessages etc... but that isn't the case. it's a single thread... there is a second thread to handle MQTT because the callbacks have to be serviced, but the entire code is single threaded aside from that.  I don't think the error the AI found is real.  If someone can reproduce it that would be different.

robjarawan:
You're right, the consumer is single-threaded so the race condition doesn't apply. Well, at least the queue.Queue refactor in #1671 gets rid of the 100ms sleep in _rotateInputBuffers -- so we get that back on every buffer rotation. :)

petersilva:
> You're right, the consumer is single-threaded so the race condition doesn't apply. Well, at least the queue.Queue refactor in #1671 gets rid of the 100ms sleep in _rotateInputBuffers -- so we get that back on every buffer rotation. :)

wow... that IS odd... I wonder why that sleep is there?   anyways... the queue.Queue version is a lot easier to grok.

reidsunderland:
I didn't realize there was a chance of message loss with an expensive `on_message` in MQTT, that's interesting. 

A possible (very slight) improvement would maybe be putting the raw MQTT message object into the queue/buffer and then calling `m = userdata._msgDecode(msg)` when the message is popped from the queue in getNewMessage and newMessages.

queue.Queue doesn't support adding or removing multiple items at once, which is too bad. And it also doesn't support removing items out of order (FIFO only).

robjarawan:
Good suggestion -- moved _msgDecode out of the callback and into newMessages/getNewMessage in #1671. The callback just does a queue.put of the raw MQTT message now.

petersilva:
closing in favour of #1671

PR #1659, opened 2026-03-30
fix #1656 filename option parsing

### LINKED ISSUE #1656: filename NONE not working properly in development branch, fine in 3.02.00rc3
filename NONE is interpreted as filename None in the latest development branch. Will do a get bisect to figure out where it broke. 

### 3.02.00rc3

```
           '.*msc/observation/atmospheric/lightning/cldn-2.0-ascii/product-ninjo_ascii-1.0.* '
           'into ${PBD}/${YYYYMMDD}/${SOURCE}/LIGHTNING/${RHH} with '
           'mirror:False filename:NONE',
```

### Development

```
           '.*msc/observation/atmospheric/lightning/cldn-2.0-ascii/product-ninjo_ascii-1.0.* '
           'into ${PBD}/${YYYYMMDD}/${SOURCE}/LIGHTNING/${RHH} with '
           'mirror:False filename:None',
```

reidsunderland:
close #1656

filename NONE is different than None, add the differentiation back, only for the filename case.

The original code was:

```python
            v = ' '.join(line[1:])
            if v == 'None':
                v=None
            setattr(self, k, v)
```

Then was changed to:

```python
            v = ' '.join(line[1:])
            if v.lower() in [ 'none', 'off', 'false' ]:
                v=None
            setattr(self, k, v)
```

Now it has a special case for `k=='filename'`

### DIFF
--- a/sarracenia/config/__init__.py
+++ b/sarracenia/config/__init__.py
@@ -1788,7 +1788,10 @@ class Config:
                 logger.info( f"{','.join(self.files)}:{lineno} if download is false, directory has no effect" )
 
             v = ' '.join(line[1:])
-            if v.lower() in [ 'none', 'off', 'false' ]:
+            # filename NONE and None are different
+            if k == 'filename' and v == 'None':
+                v=None
+            elif k != 'filename' and v.lower() in [ 'none', 'off', 'false' ]:
                 v=None
             setattr(self, k, v)
         else:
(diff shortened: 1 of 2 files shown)


petersilva (approved):
OK that does look right... you need NONE to propagate through, and it looks good.

robjarawan:
looks good -- are there any other string options where a real value could clash with none/off/false, or is filename the only weird one?

petersilva (approved):
unit tests are a good thing.

PR #1662, opened 2026-03-31
Implement AMQP1.0 MOTH 

reidsunderland:
This isn't 100% complete, there are still improvements to be made, but I don't know when I will have time to work on them.

I think it's close enough to review and merge, and I would like to stop merging and rebasing this branch to keep up with other changes :)

There are also a couple other fixes that I think are worth adding. I will add some comments in the code-review part to note those.


sarracenia/postformat/swim.py
```diff
@@ -0,0 +1,398 @@
+import json
```
reidsunderland, line 1: This SWIM message format isn't actually used by NAV CANADA right now, but Andre and I already implemented it a couple months ago, and may be useful in the future. It could be removed.

sarracenia/postformat/v02.py
reidsunderland, line 1: I had to make a few changes here to prevent crashes

reidsunderland:
@andreleblanc11 did a lot of work on this too, especially the publishing and SWIM post format. The commits are a mess and don't really show that he contributed a lot.

sarracenia/postformat/swim.py
petersilva: Sounds good... There must be some other format they are using, should there be another postformat for that?

petersilva (approved):
I think it's better to have something partially working in the mainline, as it is easier to  make issues, reproduce problems, and make test harnesses if it's merged than otherwise. It also looks upto date.

petersilva:
Evidence of the difficulty of maintaining a || branch... there's a conflict again... will need to be resolved before merging...

sarracenia/postformat/swim.py
reidsunderland: Yeah, there's a format I've called navcanada, it's pretty basic at the moment, with no support for file URLs on our side (embedded content in the message only). Once they start sending us messages with URLs then I can will make that work.

I'll hopefully add it in another PR once it's closer to being finalized and once they confirm they don't have any objection to it being published.

PR #1663, opened 2026-04-01
fix unit test name conflict

reidsunderland:
after the transfer/file_test.py was added, the unit tests were failing:

```
==================================== ERRORS ====================================
______________ ERROR collecting sarracenia/transfer/file_test.py _______________
import file mismatch:
imported module 'file_test' has this __file__ attribute:
  /home/runner/work/sarracenia/sarracenia/tests/sarracenia/flowcb/gather/file_test.py
which is not the same as the test file we want to collect:
  /home/runner/work/sarracenia/sarracenia/tests/sarracenia/transfer/file_test.py
```


petersilva (approved):
works... not sure why this is busted in the first place, but the fix is needed and correct.

PR #1664, opened 2026-04-01
regression/unit test for #1573 sundew extension filtering

reidsunderland:
replaces #1574, which I couldn't update because it's based on a fork in Rob's personal GitHub account.


reidsunderland:
I should have included this in the PR where I made the change, but I forgot Rob attempted to make a test for it. I wouldn't have bothered with a test otherwise, but Rob is a good influence :)

petersilva (approved):

PR #1668, opened 2026-04-02
V2 oldify

petersilva:
mark metpx-sarracenia (aka v2) as Mature.... 

not sure if this is what we want... seems conservative... it's really close to Inactive, but it's still supported, so not quite.


petersilva:
ack was a pr to development... no!

PR #1672, opened 2026-04-04
Fix 5 bugs in flowcb/work/delete.py on_housekeeping

robjarawan:
##### What

Fixes 5 bugs in `flowcb/work/delete.py` that make `on_housekeeping` crash or silently misbehave when `delete_destination` is enabled.

##### Bugs fixed

1. **`set(message['new_dir'])` iterates characters** -- `set('/tmp/foo')` produces `{'/', 't', 'm', 'p', 'f', 'o'}` instead of `{'/tmp/foo'}`. Fixed with `set([message['new_dir']])`.

2. **RuntimeError: Set changed size during iteration** -- `on_housekeeping` iterates `self.dirsOfDeletion` while calling `.remove()` on it. Fixed by iterating over `list(self.dirsOfDeletion)`.

3. **KeyError from `.remove()`** -- if a dir was already discarded, `.remove()` raises. Changed to `.discard()`.

4. **`self.dirsofDeltion` typo + missing import** -- capital O missing, and `dirname()` called without import. Fixed to `self.dirsOfDeletion` and `os.path.dirname()`.

5. **Wrong variable in error message** -- `f"could not unlink {f}"` references the file loop variable from `after_work`, not the directory `d` being removed. Fixed to `{d}`.

##### How found

Copilot-generated test coverage on the fork surfaced these during unit test development.

##### Test

Existing flow tests cover the `after_work` path. The `on_housekeeping` path was previously untested -- these bugs would crash on the first non-empty housekeeping cycle with `delete_destination` enabled.

### DIFF
--- a/sarracenia/flowcb/work/delete.py
+++ b/sarracenia/flowcb/work/delete.py
@@ -63,7 +63,7 @@ class Delete(FlowCB):
                 files_to_delete.append(message['delete_source'])
 
             if self.o.delete_destination:
-                self.dirsOfDeletion |= set(message['new_dir'])
+                self.dirsOfDeletion |= set([message['new_dir']])
                 files_to_delete.append(
                     f"{message['new_dir']}/{message['new_file']}")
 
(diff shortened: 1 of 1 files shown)


robjarawan:
Unit tests pass on the fork: https://github.com/robjarawan/sarracenia/actions/runs/23981554823

AMQP flow tests: 11/12 pass. Only failure is flakey_broker on ubuntu-24.04 (known flaky test, not related to this change): https://github.com/robjarawan/sarracenia/actions/runs/23981555666

robjarawan:
@petersilva are multiple bugs in 1 PR bad?

petersilva:
> @petersilva are multiple bugs in 1 PT bad?

not a problem...  because they are kind of on one topic... in one plugin... it makes sense to group them together.
The analysis is pretty common for all of them.  It would be a problem if you have to do completely different analysis for different bugs.

petersilva (approved):
fixes look good. testing data looks good.

PR #1674, opened 2026-04-06
Revert "Fix AMQP connection and channel leak on reconnect"

petersilva:
This reverts commit d4a52ba15f226e8f8fec72ebca2c51c0a4992daa. After this commit in the tree. all of the flow tests fail to remove queues during the "flow_cleanup.sh" phase. instead they delete and create new ones.

to reproduce:
* set up a host to run sr_insects' flow tests.
* put a window with the rabbitmq  manager web ui, "Queues and Streams" tab, so you can see all the queues defined.

```

cd static_flow
./flow_setup.sh declare
./flow_cleanup.sh

```

The window should show:

*  no queues extant before the flow_setup.sh runs.
* many defined queues after ./flow_setup.sh declare is run.
* no queues extant after the flow_cleanup.sh

but it doesn't instead all the same queues are there, but with different numbers.  so  delete does not work as it should (I think it deletes, and then re-creates.)  Identified this commit via bisection.

I don't know why this commit has that effect, but since it delivers no features, and will be extremely confusing in ops.
I recommend just backing out of it for now.

### DIFF
--- a/sarracenia/moth/amqp.py
+++ b/sarracenia/moth/amqp.py
@@ -249,14 +249,6 @@ class AMQP(Moth):
         else:
             sslarg=False
         
-        if self.connection:
-            try:
-                self.connection.collect()
-                self.connection.close()
-            except Exception:
-                logger.debug('failed to close old connection before reconnect', exc_info=True)
-            self.connection = None
-
         self.connection = amqp.Connection(host=host,
                                           userid=broker.url.username,
                                           password=unquote(
@@ -390,9 +382,9 @@ class AMQP(Moth):
             # from sr_consumer.build_connection...
             if not self.__connect(broker):
                 self.setEbo(start)
-                self.close()
+                self.connection = None
                 return
-
+            
             if self.o['prefetch'] != 0:
                 # using global False because RabbitMQ Quorum Queues don't support Global QoS, issue #1233
                 self.channel.basic_qos(0, queue['prefetch'], False)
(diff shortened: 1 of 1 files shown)


petersilva:
if you are running successive flow tests on a single host, every test after the first one will get messed up.

**WARNING: This deletes all queues on a broker**

I go into root to remove all queues after the flow_cleanup.sh is done:

 for q in `rabbitmqctl  list_queues | awk '{print $1;};'`; do rabbitmqctl delete_queue $q; done

**WARNING: This deletes all queues on a broker**

so it is empty enough that later tests will run normally.

petersilva:
I think you can likely reproduce on any broker by doing a "cleanup" on a queue... and it will delete the queue, but immediately create a new one with a different number not associated with any state, so can't be cleaned in future.
These zombie queues (with included bindings) will just pile up forever.

reidsunderland (approved):

reidsunderland:
This part was **not** changed by Rob, but while trying to figure out how this PR caused a problem, I noticed that in our close() method, we call `collect` before `close` on the connection:

https://github.com/MetPX/sarracenia/blob/5e4186425898979cf67e3a8281891521e92f416e/sarracenia/moth/amqp.py#L832-L836

This is backwards compared to what the docs suggest:

https://docs.celeryq.dev/projects/amqp/en/latest/reference/amqp.connection.html#amqp.connection.Connection.Channel.collect

```
collect()
    Tear down this object.

    Best called after we’ve agreed to close with the server.
```

PR #1675, opened 2026-04-10
fix typo exitType -> ExitType

reidsunderland:
Systemd complains "Unknown key name 'exitType' in section 'Service', ignoring"


andreleblanc11:
Wow I did not know this was in the source code.

andreleblanc11 (approved):

PR #1676, opened 2026-04-10
Update GitHub Actions and Docker stuff

reidsunderland:
1. Updated all GitHub Actions files to use the latest versions for each action (`uses:` lines). This eliminates the warning we are currently getting:

    image

2. Ubuntu 24.04 includes `python3-paho-mqtt` v1.6.1 which is newer than the minimum 1.5.1 version required. So I replaced the pip package with the Ubuntu package in `Dockerfile_base`

3. Refactored and updated the Docker image build and publish to GitHub Container Registry.
    - I got rid of separate files for development and stable.
    - Image builds are triggered by either pushes to the development branch or tag pushes (non-RC tags)
        - development images will get image tags `latest` and `development`
        - stable images built from tags will get tags `latest`, `stable` and `$git_tag` (e.g. v3.03.00)
 
4. Updated the email in the Dockerfile, since `peter.silva@ssc-spc.gc.ca` is inactive

5. The updated sphinx step in docs.yml with `publish: true` now handles publishing to the `gh-pages` branch. 

This should get squashed when merged, there's a lot of trial and error commits.


petersilva (approved):
I keep retrying tests to get them to pass, but some keep failing... really frusttrating, but I think unrelated from these changes.

reidsunderland:
Yeah those failed tests are annoying, but I agree that it's unrelated. Not much was changed in the flow tests except versions of imported actions, and if that didn't work I think they would all fail, but it's mostly just flakey broker failures.

PR #1677, opened 2026-04-13
Fix #989: Detect '#' in credentials and auto-unquote passwords

### LINKED ISSUE #989: credentials.conf handling of certain characters in usernames and passwords

robjarawan:
### Description:
When `credentials.conf` contains a password with an unencoded `#`, `urlparse` treats it as a URL fragment delimiter. This results in a broken parse where the `@` (user/host separator) ends up in the fragment instead of the netloc, causing a silent "credential not found" error with no useful explanation.

I ran into this issue and wanted to hurt myself last week.

This PR adds:
1.  **Transparent Unquoting**: A `UrlParseResult` subclass that auto-unquotes `.username` and `.password`. This ensures that percent-encoded characters (like `%23` for `#`) are decoded before they are used by plugins or internal code.
2.  **Heuristic Detection**: Checks if `#` has broken the URL (by looking for `@` in the fragment) and logs a clear `logger.error` suggesting `%23` as the fix.
3.  **Redaction**: Adds a `_REDACT_RE` to ensure the password is shown as `<secret>` in error logs.
4.  **Key Stripping Fix**: Uses the raw (encoded) password when stripping credentials from the URL key in `add()`, ensuring lookups by bare URL continue to work correctly for complex passwords.

**Testing:**
- 12 new unit tests in `tests/sarracenia/credentials_test.py` covering encoding/decoding round-trips, false-positive safety, and key-stripping behavior.
- Verified all tests pass in the local WSL environment.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,4 @@ pyproject-wheeldir/
 GEMINI.md
 CLAUDE.md
 GEMINI.md
+.mcp.json
--- a/.mcp.json
+++ b/.mcp.json
@@ -1,0 +1,7 @@
+{
+  "mcpServers": {
+    "playwright": {
+      "url": "http://localhost:8931/mcp"
+    }
+  }
+}
--- a/.mcp.json
+++ b/.mcp.json
@@ -1,7 +1,0 @@
-{
-  "mcpServers": {
-    "playwright": {
-      "url": "http://localhost:8931/mcp"
-    }
-  }
-}
--- a/sarracenia/config/credentials.py
+++ b/sarracenia/config/credentials.py
@@ -42,6 +42,33 @@ import urllib, urllib.parse
 import sys
 
 
+class UrlParseResult(urllib.parse.ParseResult):
+    """ParseResult subclass that auto-unquotes username and password.
+
+    urllib.parse.urlparse percent-encodes reserved characters in netloc
+    (e.g. %40 for @, %23 for #).  Consumers that use url.username or
+    url.password directly receive the raw percent-encoded form.  This
+    subclass transparently unquotes those two fields so callers never
+    need to call urllib.parse.unquote() themselves.
+    """
+
+    @property
+    def username(self) -> str:
+        raw = super().username
+        return urllib.parse.unquote(raw) if raw else raw
+
+    @property
+    def password(self) -> str:
+        raw = super().password
+        return urllib.parse.unquote(raw) if raw else raw
+
+
+def _urlparse(urlstr: str) -> UrlParseResult:
+    """Parse a URL string and return a UrlParseResult with auto-unquoted credentials."""
+    pr = urllib.parse.urlparse(urlstr)  # keep internal _urlparse use of stdlib
+    return UrlParseResult(pr.scheme, pr.netloc, pr.path, pr.params, pr.query, pr.fragment)
+
+
 class Credential:
     r"""
 
@@ -88,7 +115,7 @@ class Credential:
         """
 
         if urlstr is not None:
-            self.url = urllib.parse.urlparse(urlstr)
+            self.url = _urlparse(urlstr)
         else:
             self.url = None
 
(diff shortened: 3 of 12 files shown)


andreleblanc11:
I tested this on our dev environment, and it does work. However, the password credentials gets propagated to the code downstream and eventually gets printed in a message log.
```
# Without '%23' inside of the credentials.conf entry
2026-04-13 13:09:45,996 [INFO] sarracenia.moth.amqp putSetup exchange declared: xs_TEST (as: amqps://TEST@dev-broker/)

# With '%23' inside of the credentials.conf entry
2026-04-13 13:10:42,911 [INFO] sarracenia.moth.amqp putSetup exchange declared: xs_TEST (as: amqps://TEST:test123%23@dev-broker)
```

reidsunderland:
I think the `unquote` calls would also have to be removed, or switched to `.raw_password`

```
~/sr3/sarracenia$ grep -ri unquote | grep 'password\|username'
flowcb/download/mail_ingest.py:                        password = urllib.parse.unquote(password)

rabbitmq_admin.py:        command += '\' -p \'' + urllib.parse.unquote(url.password)
rabbitmq_admin.py:        url.username + ':' + urllib.parse.unquote(url.password), "utf-8")

moth/amqp.py:                                          password=unquote(

moth/mqtt.py:        client.username_pw_set(broker.url.username, unquote(broker.url.password))
moth/mqtt.py:                                        unquote(self.o['broker'].url.password))


transfer/ftp.py:                ftp.login(self.user, unquote(self.password))
transfer/ftp.py:                ftp.login(user=self.user, passwd=unquote(self.password))
transfer/ftp.py:                                     unquote(self.password),

transfer/https.py:                self.password_mgr.add_password(None, self.sendTo, self.user, unquote(self.password))

transfer/sftp.py:                self.ssh.connect(self.host,self.port,self.user,unquote(self.password), \

transfer/azure.py:                    self.key = urllib.parse.unquote_plus(url.password)
```

andreleblanc11 (approved):
I'm not seeing the password in the logs anymore when the exchange gets declared.

I was a bit worried that those `broker_str = broker.url.geturl().replace( ':' + broker.url.password + '@', '@')` changes would show the credentials in the logs if they are added inside of the configuration file. That doesn't seem to be the case so it looks good to go :).

robjarawan:
Pushed two more commits addressing André and Reid's feedback:

##### What changed

**Root cause of the password-in-log leak (André's report)**

`geturl()` preserves percent-encoding (`%23`), but the broker_str scrub was using `.password` which returns the decoded form (`#`). So `replace(':pass#word@', '@')` found no match in a string that contained `:pass%23word@`, and the raw credential appeared in INFO logs.

Two subscriber callsites (`putSetup`, `getSetup`) were switched to `str(broker)`, which uses `Credential.__str__` and already omits the password entirely. The publisher callsite (`putSetup`) uses `.raw_password` (encoded form) in the replace, which does match correctly.

**Redundant `unquote()` calls (Reid's comment)**

`UrlParseResult.password` already decodes percent-encoding internally, so all the `unquote(broker.url.password)` wrappers in `amqp.py`, `mqtt.py`, and `rabbitmq_admin.py` were double-decoding. Removed them.

**Regression test added**

`Test_LogScrubbing` in `credentials_test.py` pins both the correct and broken behaviour, including a test that explicitly shows why the old decoded-password replace fails on an encoded URL.

##### Verify: plain passwords still work

For a plain password (no `%XX` sequences): `raw_password == password`, so the `geturl().replace(':secret@', '@')` still matches correctly — nothing breaks for the common case.

PR #1684, opened 2026-04-24
Fix rabbitmq_admin __main__ using stdlib urlparse for percent-encoded passwords

robjarawan:
After #1677 removed the explicit `unquote()` calls from `exec_rabbitmqadmin`, the function relies on `UrlParseResult.password` returning the decoded value. The `__main__` block still used plain `urllib.parse.urlparse()`, which does not decode percent-encoding, so passwords like `pass%23word` were passed verbatim to rabbitmqadmin instead of `pass#word`.

##### What changed

Switched `__main__` to `_urlparse()` so `url.password` returns the decoded value consistently with all other code paths.

##### Tests

- `test_encoded_hash_password_decoded_for_command` — confirms `%23` decodes to `#` in the subprocess argument list
- `test_plain_urlparse_would_pass_encoded_password` — documents why plain `urlparse` is not sufficient here

Fork CI unit tests: passing — https://github.com/robjarawan/sarracenia/actions/runs/24844943442

### DIFF
--- a/sarracenia/rabbitmq_admin.py
+++ b/sarracenia/rabbitmq_admin.py
@@ -291,7 +291,8 @@ def user_access(url, user):
 
 
 if __name__ == "__main__":
-    url = urllib.parse.urlparse(sys.argv[1])
+    from sarracenia.config.credentials import _urlparse
+    url = _urlparse(sys.argv[1])
     print(exec_rabbitmqadmin(url, "list queue names")[1])
 
     import json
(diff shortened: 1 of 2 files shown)


robjarawan:
This one-liner is correct as far as it goes, but I confirmed it leaves the shell-string construction, the password-in-logs leak, and both `eval` sites untouched, so merging it solo would ship green tests over a still-vulnerable file.

My recommendation is to land #1686, which I verified is a true superset of this fix plus #1685, and close this one as superseded. Posting instead of closing to keep the trail visible. The `__main__` `_urlparse` fix from here is preserved verbatim in #1686, including the test coverage.

PR #1685, opened 2026-04-24
Fix shell injection and password corruption in exec_rabbitmqadmin

robjarawan:
Two bugs in `exec_rabbitmqadmin`:

1. The legacy path used `subprocess.getstatusoutput()` with `shell=True`, with `url.hostname`, `url.username`, `url.password`, and `options` all string-concatenated into the command. A password containing `;`, `$(`, or backticks could execute arbitrary commands.

2. The modern path did `command.replace("'", "").split()` — this stripped every single quote from the entire command string and then split on whitespace. A password with a single quote was silently corrupted; a value containing a space was split into multiple arguments.

##### What changed

Build a list from the start and pass it directly to `subprocess.run()`. The password is always a standalone list element, never touched by the shell. `shlex.split()` is used only for the `options` string, which is operator-controlled input like `"list exchanges name"`.

Also drops the Python < 3.5 code path — Sarracenia requires Python 3.6+.

##### Tests

- `test_password_passed_as_separate_argument` — subprocess receives a list, password is its own element
- `test_password_with_single_quote_not_corrupted` — single quote in password survives intact
- `test_no_shell_injection_via_password` — shell metacharacters in password are inert
- `test_options_tokenised_correctly` — options string is split into tokens correctly
- `test_amqps_adds_ssl_flag` — `--ssl` and `--port=15671` added for amqps scheme

Fork CI unit tests: passing — https://github.com/robjarawan/sarracenia/actions/runs/24844944675

(edited main comment to trigger full re-run of actions jobs)

### DIFF
--- a/sarracenia/rabbitmq_admin.py
+++ b/sarracenia/rabbitmq_admin.py
@@ -3,6 +3,7 @@
    rabbitmq administration bindings, to allow sr to invoke broker management functions.
 
 """
+import shlex
 import sys
 import urllib, urllib.parse
 import base64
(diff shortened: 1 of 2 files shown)


andreleblanc11:
I think https://github.com/MetPX/sarracenia/pull/1685/changes/e5b8e5c0f87776ac5dc3938d6c21ab071d0dd9a5 should be its own MR. It's not part of the scope of the original MR.

andreleblanc11:
You should run the flow tests for this change (U18.04 and U20.04 as well). From what I can see, the rabbitmq_admin stuff is only really used for `sr3 declare`. The flow tests run `sr3 declare` and it would be a good way to see if the changes work correctly.. especially since there was some version adaptations in the old code. Would be good to verify new changes work also on older OS'es. I can't run the flow tests on GitHub because the changes come from your personal repository.

robjarawan:
@andreleblanc11 flow testing is complete on my fork.

Successful runs:
- Flow Tests / AMQP: https://github.com/robjarawan/sarracenia/actions/runs/28761573282
- Flow Tests / AMQP Consumer: https://github.com/robjarawan/sarracenia/actions/runs/28761574501
- Maintenance declare/cleanup: https://github.com/robjarawan/sarracenia/actions/runs/28761573878
- Maintenance rerun after rebasing/cleaning the branch, Ubuntu 22.04 and 24.04 both passed: https://github.com/robjarawan/sarracenia/actions/runs/29218413291

The current unit run completed with 365 passed and one failure in the time-dependent scheduled callback test. That test is outside the rabbitmq_admin changes. The focused credential/rabbitmq/mail-ingest tests also pass locally (23 passed).

The current project workflows cover Ubuntu 22.04 and 24.04. I don't have Ubuntu 18.04/20.04 runners available, so I'm not claiming those versions were tested.

robjarawan:
I diffed this against #1686 head to head and #1686 is a verified true superset: identical `exec_rabbitmqadmin` list-build and `__main__` fixes, plus the `run_rabbitmqadmin` eval fix and four extra tests.

So my recommendation is to land #1686 and close this one as superseded. I am posting this instead of closing so reviewers can see the trail. The one correction worth carrying over: the declare path is operator-fed through `sr.py:1550` credentials, so `shlex.split` on the interpolated string is real arg-mangling, not just a theoretical flag injection. The charset-agnostic fix is argv-list construction, not a validation regex, since decoded passwords legitimately contain specials.

Anything still open (both `eval` sites, silent `return 0, None`, no timeout) is tracked on #1686.

PR #1686, opened 2026-04-24
Replace eval() with json.loads() in run_rabbitmqadmin

robjarawan:
`run_rabbitmqadmin` called `eval()` on output returned by rabbitmqadmin. A compromised broker or a network MITM could return a Python expression instead of JSON and execute arbitrary code in the sr3 process.

`exec_rabbitmqadmin` already passes `--format raw_json`, so the output is always a JSON array. `json.loads()` is the correct parser and raises `JSONDecodeError` on non-JSON input instead of executing it.

##### What changed

- `eval(answer)` replaced with `json.loads(answer)` with a `JSONDecodeError` handler that logs and returns `[]`
- Dead duplicate None/empty check above the parse block removed (already handled by the earlier guard)
- Bare `except:` tightened to `except Exception:`
- `import json` promoted to module level (was duplicated in `__main__`)

##### Tests

- `test_valid_json_parsed_to_list` — valid JSON array round-trips correctly
- `test_malicious_eval_payload_not_executed` — a list comprehension that would run under `eval()` is safely rejected by `json.loads()`
- `test_non_json_response_returns_empty_list` — error text from rabbitmqadmin returns `[]` without raising
- `test_failed_subprocess_returns_empty_list` — non-zero exit code returns `[]` without raising

Fork CI unit tests: passing — https://github.com/robjarawan/sarracenia/actions/runs/24844972369

### DIFF
--- a/sarracenia/rabbitmq_admin.py
+++ b/sarracenia/rabbitmq_admin.py
@@ -3,6 +3,7 @@
    rabbitmq administration bindings, to allow sr to invoke broker management functions.
 
 """
+import json
 import shlex
 import sys
 import urllib, urllib.parse
@@ -279,8 +280,6 @@ if __name__ == "__main__":
     url = _urlparse(sys.argv[1])
     print(exec_rabbitmqadmin(url, "list queue names")[1])
 
-    import json
-
     lex = list(
         map(lambda x: x['name'],
             json.loads(exec_rabbitmqadmin(url, "list exchanges name")[1])))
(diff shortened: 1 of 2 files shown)


robjarawan:
A note on my own PR. The `run_rabbitmqadmin` eval fix is solid and verified. Three conditions before I consider this done.

First, `broker_get_exchanges:184` still calls `eval(answer)` on broker-controlled bytes. That is the same live RCE as the headline fix, so it must ride in this PR, not a follow-up. Mirror the `:317` fix with `json.loads` plus a `JSONDecodeError` guard returning `[]`. Note `json` is only imported function-locally at `:209` and `:283` right now, so place the import where the fix needs it, and decode bytes before logging or slicing.

Second, `__main__:291` calls `rabbitmq_user_access`, which does not exist. Rename to `user_access`, whose `(url, u)` signature matches.

Third, the `exec_rabbitmqadmin` exception path returns `0, None`, which masks failure as success. Return `1, None` so `run_rabbitmqadmin`'s status check can see it.

I also confirmed the `user_access` no-match `UnboundLocalError` and the `'error' in answer` heuristic as follow-ups, plus `timeout=` and `stderr` capture on the subprocess call. I will file those separately.

Issue #1690, opened 2026-05-14
destfnscripts get loaded more then once when referenced multiple times


andreleblanc11:
Say you have the following configuration with the following accept statements

```
# sarra/myconfig.conf
accept .*something1.* DESTFNSCRIPT=destfn.colour_image

accept .*something2.* DESTFNSCRIPT=destfn.colour_image

accept .*something3.* DESTFNSCRIPT=destfn.colour_image
```

The destfn script will get loaded multiple times in the callbacks.

```
# In the logs (with DEBUG enabled)
2026-05-14 16:32:05,108 [DEBUG] sarracenia.flow loadCallbacks flowCallback plugins to load: ['sarracenia.flowcb.gather.message.Message', 'sarracenia.flowcb.retry.Retry', 'sarracenia.flowcb.housekeeping.resources.Resources', 'log', 'destfn.colour_image', 'destfn.colour_image', 'de
stfn.colour_image']
```

Because we loop through all `destfn` plugins to match and execute it, we end up executing the same destfn script the same amount of times it is referenced in the configurations.

https://github.com/MetPX/sarracenia/blob/a7010170c4688b9cf8ec0251a2661a46f9bad37f/sarracenia/flow/__init__.py#L787-L800

The fix is likely to modify these lines in `config.py` to only add the destfn plugin once when found.

https://github.com/MetPX/sarracenia/blob/a7010170c4688b9cf8ec0251a2661a46f9bad37f/sarracenia/config/__init__.py#L1107-L1112

PR #1691, opened 2026-05-15
Change destfn_script into a set so duplicate entries dont get propogated

andreleblanc11:
Change `destfn_scripts` from a list to a set. This prevents duplicate entries from reaching the callback list of plugins.

I retested the same config as the one in #1690 with the patch applied and the callback only gets loaded once from the list.

```
2026-05-15 15:51:14,262 [DEBUG] sarracenia.flow loadCallbacks flowCallback plugins to load: ['sarracenia.flowcb.gather.message.Message', 'sarracenia.flowcb.retry.Retry', 'sarracenia.flowcb.housekeeping.resource
s.Resources', 'log', 'destfn.colour_image']
...
 'destfn_scripts': {'destfn.colour_image'},

 'accept '
           '.*something1.* '
           'into /net/local/home/leblanca/test/ with mirror:False '
           'filename:DESTFNSCRIPT=destfn.colour_image',
           'accept '
           '.*something2.* '
           'into /net/local/home/leblanca/test/ with mirror:False '
           'filename:DESTFNSCRIPT=destfn.colour_image',
           'accept '
           '.*something3.* '
           'into /net/local/home/leblanca/test/ with mirror:False '
           'filename:DESTFNSCRIPT=destfn.colour_image'],
```

### DIFF
--- a/sarracenia/config/__init__.py
+++ b/sarracenia/config/__init__.py
@@ -897,7 +897,7 @@ class Config:
         self.v2plugin_options = []
         self.imports = []
         self.logEvents = set(['after_accept', 'after_post', 'after_work', 'on_housekeeping' ])
-        self.destfn_scripts = []
+        self.destfn_scripts = set() # Define a set to avoid duplicate entries
         self.plugins_late = []
         self.plugins_early = []
         self.exchange = None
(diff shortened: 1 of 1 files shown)


reidsunderland (approved):
It looks good.

There's only one other place that destfn_scripts is used:

```
flow/__init__.py:        self.plugins['load'].extend(self.o.destfn_scripts)
```

And that works fine with a set:

```
>>> myset = {1,2,3}
>>> mylist = [9,8,7]
>>> mylist
[9, 8, 7]
>>> myset
{1, 2, 3}
>>> mylist.extend(myset)
>>> mylist
[9, 8, 7, 1, 2, 3]
```

PR #1693, opened 2026-05-19
Octal number tweaks

reidsunderland:
The octal_number class wasn't quite working right, the `__str__` and `__repr__` methods were never actually being used because `self = value` was just changing self to be an actual real `int`, not an instance of `octal_number`.

This fixes that, so octal numbers are actually stored as instances of the `octal_number` class, which means the `__str__` and `__repr__` methods will work as expected.

I also made a few improvements:
- added 0 padding to the string output
- made it work when given a string like `"0o644"` (before it would only work with `"644"`, now both work).
- added a unit test (generated by claude)

Now when octal numbers are printed, they are actually octal:

```
# from sr3 show
 'permDefault': 0o000,
 'permDirDefault': 0o775,
 'permLog': 0o600,
 'permDefault': 0o000,
 'permDirDefault': 0o775,
 'permLog': 0o600,
 'permDefault': 0o000,
 'permDirDefault': 0o775,
 'permLog': 0o600,
```

The numbers still behave as if they were regular integers, e.g. `octal_number("644") == 420`

### DIFF
--- a/sarracenia/config/__init__.py
+++ b/sarracenia/config/__init__.py
@@ -66,6 +66,7 @@ class octal_number(int):
 
     def __new__(cls, value):
         if type(value) is str:
+            value = value[2:] if value.startswith("0o") else value
             return super().__new__(cls, int(value, base=8))
         elif type(value) is int:
             return super().__new__(cls, value)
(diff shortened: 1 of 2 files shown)


andreleblanc11:
The unit tests now pass :smile:

andreleblanc11 (approved):

reidsunderland:
Thanks :)

Issue #1689, opened 2026-05-06
FTP poll permission mismatch can lead to confusing 0-byte downloads


robjarawan:
I hit this while chasing an FTP poll/sarra setup where the FTP account could log in and list enough to produce messages, but did not actually have read access to the path being posted.

What I saw downstream was a 0-byte file, and from the outside it looked like the flow had succeeded. The confusing part was that this looked different from the SFTP failure case, where the auth/read problem is obvious and the message goes to retry.

After Reid's note below, I think my original framing was too broad. If the expected size is known and `acceptSizeWrong` is off, SR3 should already reject/retry size mismatches. So the useful thing to check is probably narrower:

- did the poll side publish something it should not have published because permissions were wrong?
- should `permDefault` have caught this before the message was posted?
- was this config using an option like `acceptSizeWrong` that made the 0-byte result acceptable?

The workaround was to switch the poll to an account with direct read access to the path, using SFTP in this case. That avoided the zero-byte output.

Leaving this open for now as something to verify around FTP poll permissions / `permDefault`, not as a claim that every 0-byte FTP transfer bypasses retry.

reidsunderland:
This behaviour is controlled by the `acceptSizeWrong` option. It is turned off by default. The way it works is the same for 0-byte files or any file size that is different than what is expected based on the size in the message.

When the file size doesn't match the expected size, the message **does** go in the retry queue and retry.

> Zero-byte files propagate downstream and overwrite any valid cached copy at the destination. Clients receive corrupt (empty) data with no indication of failure. The problem is silent — Nagios and normal log monitoring see a successful delivery.

Only if `acceptSizeWrong` is turned on (it's off by default).

---

The [permDefault](https://metpx.github.io/sarracenia/Reference/sr3_options.7.html#permdefault-permdirdefault-permlog-permcopy) option can be used to prevent a poll from publishing files that it doesn't have permission to read. That is probably something to look into to confirm that it works as expected.

robjarawan:
Thanks Reid, that makes sense. I cleaned up the issue text because my original wording was too broad.

I think the useful thing to check is the poll/permissions side now - whether `permDefault` should have prevented this from being posted in the first place, or whether this was just config/expected behaviour with the size handling options.

robjarawan:
Adding the exact scenario that exposed this, so we have something concrete to test.

[...]

PR #1694, opened 2026-05-19
change `chmod` to `permDefault` in poll default options to prevent posting files that user can't read

reidsunderland:
The poll is not supposed to publish files that the poll user does not have permission to read. This causes download failures in the downstream sarra/subscriber.

[`permDefault` is supposed to act like a mask in a poll:](https://metpx.github.io/sarracenia/Reference/sr3_options.7.html#:~:text=when%20set%20in%20a%20polling%20component%2C%20permDefault%20has%20the%20of%20setting%20minimum%20permissions%20for%20a%20file%20to%20be%20accepted.%20(on%20an%20ls%20output%20that%20looks%20like%3A%20%E2%80%94r%E2%80%94%E2%80%93%2C%20like%20a%20chmod%20040%20%3Cfile%3E%20command).%20The%20permDefault%20options%20specifies%20a%20mask%2C%20that%20is%20the%20permissions%20must%20be%20at%20least%20what%20is%20specified.)

> when set in a polling component, permDefault has the of setting minimum permissions for a file to be accepted. (on an ls output that looks like: —r—–, like a chmod 040 <file> command). The permDefault options specifies a mask, that is the permissions must be at least what is specified.

The code for this was correct, but the default options in poll was using `chmod`, the old v2 name for `permDefault` and this was preventing `permDefault` from being applied correctly.

This deals with one part of #1689.

### DIFF
--- a/docs/source/How2Guides/UPGRADING.rst
+++ b/docs/source/How2Guides/UPGRADING.rst
@@ -49,8 +49,8 @@ verify that this will not cause any unexpected changes.
 
 *CHANGE*: polls are supposed to filter out files that are not readable by the user that is polling, using the
 ``permDefault`` option as a mask. This was not working correctly and has been fixed, but may cause polls to miss
-files that were previously being posted. If any poll configs depend on this broken behaviour, the ``permDefault``
-option may need to be adjusted.
+files that were previously being posted. If any poll configs depend on the old, broken behaviour, the ``permDefault``
+can be set to ``000`` to allow un-readable files to be posted.
 
 3.02.00
 -------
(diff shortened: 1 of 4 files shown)


andreleblanc11 (approved):
The changes look good. We'll just need to be careful and monitor the implementation of this on operations. There might be behaviour changes for polls that were bypassing the permDefault guard when polling.

reidsunderland:
~~Thanks for catching the test problem~~ (I commented on the wrong PR)

reidsunderland:
The failing tests are suspicious, I'm going to try running them again locally first.

reidsunderland:
They are still passing locally for me :(

PR #1697, opened 2026-05-21
plugin to filter files using keys

reidsunderland:
I hope the comment explains it well...

> Normal accept/reject filtering uses regex, which can be computationally expensive. If you have a busy flow and need
> many accept/reject statements, this plugin might be a more efficient alternative when filenames (or paths) are
> relatively consistent.
>
> It is inspired by the sundewpxroute plugin, but is more generic (does not require a routing table, works with
> non-bulletin files by using regex match groups to define what is used as the key for matching).
> https://github.com/MetPX/sarracenia/blob/development/sarracenia/flowcb/accept/sundewpxroute.py
>
> The key filtering applies *after* sr3's built-in regex filtering. This plugin assumes you will use a small number
> of accept/reject statements and potentially hundreds of accept_key/reject_key statements.

I tried a side-by-side comparison of similar configs, one using only 1043 accept regex statements and one using 1 accept statement and 1043 accept_key (with download turned off). The config using this plugin used about ~35% less CPU over a few minutes of running.

### DIFF
--- a/sarracenia/flowcb/accept/key_filter.py
+++ b/sarracenia/flowcb/accept/key_filter.py
@@ -1,0 +1,148 @@
+"""
+Key Filtering
+
+Normal accept/reject filtering uses regex, which can be computationally expensive. If you have a busy flow and need
+many accept/reject statements, this plugin might be a more efficient alternative when filenames (or paths) are
+relatively consistent.
+
+It is inspired by the sundewpxroute plugin, but is more generic (does not require a routing table, works with
+non-bulletin files by using regex match groups to define what is used as the key for matching).
+https://github.com/MetPX/sarracenia/blob/development/sarracenia/flowcb/accept/sundewpxroute.py
+
+The key filtering applies *after* sr3's built-in regex filtering. This plugin assumes you will use a small number
+of accept/reject statements and potentially hundreds of accept_key/reject_key statements.
+
+If key_group= is not defined, key filtering will be skipped for that mask (if the mask is accept, all files matching
+will be accepted, and if it's reject, all files matching will be rejected).
+
+
+Example Config 1:
+
+    acceptUnmatched False
+
+    callback accept.key_filter
+
+    accept_key AACN01_CWAO
+    accept_key AACN02_CWAO
+    reject_key SACN01_CWAO
+
+    # at least one accept is required
+    directory /
+    # Bulletins with Sundew AHL filename format, T1T2A1A2ii_CCCC...
+    #                                    ⮦ this is the group used as the key
+    accept  .*MSC-BULLETINS.*/([A-Z]{4}[0-9]{2}_[A-Z]{4}).*  key_group=1
+
+    accept  .*key_filtering_not_applied_for_files_matching_this.*
+
+
+Example Config 2:
+
+    acceptUnmatched False
+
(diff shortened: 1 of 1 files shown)


reidsunderland:
I made this because using tons of accept/rejects uses a lot of CPU, but we aren't sure that we want to keep maintaining routing tables forever.

And it can be used with non-bulletin files.

andreleblanc11:
I had to read through the documentation and had to look at your configuration on dev, but it makes sense to me now. Nice work! 😄

andreleblanc11 (approved):

petersilva:
looks cool!  dunno why I'm reading this... but whatever...

so... in the original Sundew.... the accepts were used to generate keys for the routing table:
(from https://github.com/MetPX/Sundew/blob/main/doc/man/pxRouting.7.rst#generate-a-product-key )

given:

```
accept .*(txt:AQ).*(:AIRNOW:.*:ASCII)

```
and the ingest filename of the incoming file being: 20070515.txt:AQ1245Z:AIRNOW:CMQ:ASCII, 
the product key generated will be txt:AQ_:AIRNOW:CMQ:ASCII

Instead of "accept" ... one could use something like "derive_key" ... and to use regexes to extract a key from filenames.  then use the accept_key, and also reject_key. 

Probably a more flexible way of doing this is to add a message field "key" and that way multiple plugins can be layered to derive keys in different ways, and override eachother, then the existing plugin is there to make a final decision.

The sundew pxrouting thing uses the routing tables because they were already existing, so if
you want to transition an existing sender, you could set the *pxClient* in the new sender,
and the routing table should "just work."

The list of "accept_key" lines is just the fraction of the routing table that is relevant to a given destination.  In the old sundewpxroute thing... it didn't keep the whole routing table in memory.  Instead, each component would declare what routing destination it was _equivalent_ to (the *pxClient* setting) , and then only the relevant parts of the routing table were kept.   

Which is not to say anything about the new PR, which looks fine... it's just background/explanation about what came before...

reidsunderland:
That's interesting, thanks, I didn't know how Sundew derived the keys. Coincidentally that is pretty similar to what I did in this plugin, except I'm only using one user-defined regex group, and Sundew is using all groups joined by `_`. 

We like the sundewpxroute plugin, but are hesitant about keeping the routing table around indefinitely. It's painful to manage routing tables right now because there's dozens (maybe hundreds) of obsolete clients and aliases. We might decide to keep a routing table just for bulletins and do a major cleanup on it, but we're not sure yet.

I like the idea of storing the key in the message and separating the parts, one or more plugins to derive keys and one to filter them.

reidsunderland:
I'm working on revising this to work with multiple match groups, so it can derive keys just like Sundew.

> "The list of "accept_key" lines is just the fraction of the routing table that is relevant to a given destination. In the old sundewpxroute thing... it didn't keep the whole routing table in memory. Instead, each component would declare what routing destination it was equivalent to (the pxClient setting) , and then only the relevant parts of the routing table were kept."

when using this plugin, this task is offloaded to the person writing the config. The assumption is that they will only put the keys that are actually needed into the config.  :)

I used a modified version of the sundewpxroute plugin as a standalone script that can read the routing table and print a list of `accept_keys` for a given client name.

PR #1708, opened 2026-06-11
SFTP *upload* performance improvement

reidsunderland:
Paramiko SFTP performance is terrible. We can use binaries (accelThreshold, accelScpCommand) to get better performance, but this change makes the built in Paramiko SFTP a lot faster for *uploads* without that.

It just adds a call [`set_pipelined(pipelined=True)`](https://docs.paramiko.org/en/latest/api/sftp.html#paramiko.sftp_file.SFTPFile.set_pipelined) after opening the remote file.

This can be turned off by adding `sftp_compat_mode` to the credential if it causes any issues with a particular feed/server.

There is no change to downloads here. I've tested a variety of things to improve download speed but the results are underwhelming and they involve tweaking things like the [window size](https://docs.paramiko.org/en/latest/api/sftp.html#paramiko.sftp_client.SFTPClient.from_transport). These tweaks seem more likely to cause problems so I'm going to hold off on those changes for now.

### Test Results

Using a sender that sends to sftp://localhost/.

Upload speed before the change (i.e. with `sftp_compat_mode` used) -- <1.0 MiB/s:

```
2026-06-11 19:42:13,084 [INFO] 1249368 sarracenia.flowcb.log after_work sent ok: /tmp/sender/tmp/aspyras.tar.gz size: 267.0MiB (279956035) rate: 1022.0KiB/s (1046491.53 B/s)
```

Upload speed after the change -- 18.9 MiB/sec:

```
2026-06-11 19:37:02,415 [INFO] 1249277 sarracenia.flowcb.log after_work sent ok: /tmp/sender/tmp/aspyras.tar.gz size: 267.0MiB (279956035) rate: 18.9MiB/s (19831424.87 B/s)
```

### DIFF
--- a/docs/source/Reference/sr3_credentials.7.rst
+++ b/docs/source/Reference/sr3_credentials.7.rst
@@ -41,6 +41,7 @@ passwords and settings needed by components.  The format is one entry per line. 
 
 - **sftp://user5:password5@host**
 - **sftp://user6:password6@host:22  ssh_keyfile=/users/local/.ssh/id_dsa**
+- **sftp://user5:password5@host  sftp_compat_mode**
 
 - **ftp://user7:password7@host  passive,binary**
 - **ftp://user8:password8@host:2121  active,ascii**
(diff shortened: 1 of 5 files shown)


reidsunderland:
A real world example, going over the Internet:

BEFORE - 1.1 MiB/sec, about 5 minutes to transfer the file:
```
2026-06-11 10:43:40,131 [INFO] sarracenia.flowcb.log after_accept accepted: (lag: 30.54 )  ... size: 336.8MiB (353142239)
2026-06-11 10:48:39,065 [INFO] sarracenia.flow send whole Sent: ... size: 336.8MiB (353142239) rate: 1.1MiB/s (1181337.88 B/s)
```

AFTER - 14.9 MiB/sec, about 23 seconds:
```
2026-06-11 20:11:55,551 [INFO] sarracenia.flowcb.log after_accept accepted: (lag: 23.16 )  ... size: 336.8MiB (353142239)
2026-06-11 20:12:18,174 [INFO] sarracenia.flow send whole Sent: ... size: 336.8MiB (353142239) rate: 14.9MiB/s (15609590.57 B/s)
```

(these are not identical files but they are the exact same size. they're from the same dataset.)

reidsunderland:
Improvement visualized after enabling the change for one SFTP sender:
image

(there isn't much change for small files)

andreleblanc11 (approved):
I'm slightly worried that changing the behaviour might cause problems to some of our senders, but it's probably for the best. We have the work-around to disable the feature in the credentials so it should be OK if ever there's a problem.

reidsunderland:
Yeah me too. I wasn't able to find any problems in my tests but I figured I would add the `sftp_compat_mode` option to disable it instead of deploying it on ops and having to roll back if it causes issues.

Issue #1703, opened 2026-06-01
`declare env MY_ENV_VAR=something` applies to **all** configs, not just the config that has `declare env` in it


reidsunderland:
Starting with a clean `.config` dir, no configs in it at all except admin.conf, default.conf, credentials.conf.

Add two configs:

```
sr3 add subscribe/dd_aqhi.conf
sr3 add subscribe/dd_swob.conf
```

Edit dd_aqhi and add `declare env MY_ENV_VAR=something`.

It shows up in sr3 show:

```
sarra@server:~/.config/sr3/subscribe$ sr3 show subscribe/dd_aqhi 2>&1 | grep -B1 MY_ENV_VAR
 'env_declared': ['FLOWBROKER', 'SFTPUSER', 'TESTDOCROOT', 'MQP', 'several',
                  'MY_ENV_VAR'],
```

It does not show up in sr3 show for dd_swob (correct):

```
sarra@server:~/.config/sr3/subscribe$ sr3 show subscribe/dd_swob 2>&1 | grep -A1 env_declared
 'env_declared': ['FLOWBROKER', 'SFTPUSER', 'TESTDOCROOT', 'MQP', 'several'],
```

Yet when you start dd_swob, somehow that variable is ending up in its environment:

```
sarra@server:~/.config/sr3/subscribe$ sr3 start subscribe/dd_swob
2026-06-01 17:21:59,912 3747064 [WARNING] sarracenia.config finalize subscribe/dd_swob.conf Both broker and post_broker are set to None
2026-06-01 17:21:59,913 3747064 [ERROR] sarracenia.config.subscription write failed: /home/sarra/.cache/sr3/subscribe/dd_aqhi/subscriptions.json: [Errno 2] No such file or directory: '/home/sarra/.cache/sr3/subscribe/dd_aqhi/subscriptions.json'
starting:.( 1 ) Done


sarra@server:~/.config/sr3/subscribe$ ps aux | grep dd_swob
sarra    3747067  3.2  0.3  87208 57976 pts/0    S    17:21   0:00 /usr/bin/python3 /home/sarra/sr3/sarracenia/instance.py --no 1 start subscribe/dd_swob

sarra@server:~/.config/sr3/subscribe$ strings /proc/3747067/environ
SHELL=/usr/bin/bash
PWD=/home/sarra/.config/sr3/subscribe
LOGNAME=sarra
HOME=/home/sarra
LANG=en_US.UTF-8
[...]

PR #1711, opened 2026-06-12
Fix issue #1703 - environment variables declared in one config actually apply to all configs

reidsunderland:
The explanation of why this happens is in #1703.

### DIFF
--- a/sarracenia/config/__init__.py
+++ b/sarracenia/config/__init__.py
@@ -864,8 +864,9 @@ class Config:
             Config.credentials.read(get_user_config_dir() + os.sep +
                                     "credentials.conf")
         self.directory = None
-
-        self.env = copy.deepcopy(os.environ)
+        
+        # dict is required, deepcopy won't work, see issue #1703
+        self.env = dict(os.environ)
 
         egdir = os.path.dirname(inspect.getfile(sarracenia)) + os.sep + 'examples' 
 


andreleblanc11:
The explanation makes sense. I tested the code on a dev environment.

```
# URP_DIR is referenced in the `directory` entries.
sarra/get-nurp1-dev-geotiff-6min.conf:declare env URP_DIR=urpd

# Get process ID
ps aux | grep get-nurp1-dev-geotiff-6min
sarra    29450  8.3  0.0 181588 64700 pts/1    S    17:57   0:00 /usr/bin/python3 /local/home/sarra/sr3/sarracenia/instance.py --no 1 start sarra/get-nurp1-dev-geotiff-6min

strings /proc/29450/environ
# URP_DIR does not show up. I thought it would've still showed up.


# However, we do see the directory being correctly set on download which shows that the variable setting works.

tail sarra_get-nurp1-dev-geotiff-6min_01.log
2026-06-30 18:01:58,327 [INFO] sarracenia.flowcb.log after_work downloaded ok: /apps/sarra/public_data/20260630/MSC-RADAR/urpd/GEOTIFF/COVERAGE/MMHR/20260630T1800Z_MSC_Radar-Coverage_MMHR-EPSG3978.json size: 39.0KiB (39933) rate: 4.0MiB/s (4161714.18 B/s)
```

andreleblanc11 (approved):

reidsunderland:
I didn't notice that when testing the change (I saw the wrong variables in `/proc/$PID/environ` before the fix but didn't check back after).

It's expected behaviour according to the man page, although it's still too bad because it's nice to be able to look there and see a processes' env vars.

```
/proc/[pid]/environ
       This file contains the initial environment that was set when the currently executing program was started via execve(2).  The entries are separated by
       null bytes ('\0'), and there may be a null byte at the end.  Thus, to print out the environment of process 1, you would do:

         $ strings /proc/1/environ

       If,  after  an  execve(2),  the  process  modifies its environment (e.g., by calling functions such as putenv(3) or modifying the environ(7) variable
       directly), this file will not reflect those changes.
```

Issue #1710, opened 2026-06-12
`sr3 declare` problems (it can run on disabled configurations)


andreleblanc11:
## How to recreate

```shell

# Turn off, disable and cleanup an existing configuration that has a queue declared on a broker.
sr3 stop sarra/get-radar-for-client
# Cleanup will remove the queue from the broker
sr3 cleanup sarra/get-radar-for-client
sr3 disable sarra/get-radar-for-client

grep queue sarra/get-radar-for-client.conf
queueName q_feeder.${PROGRAM}.${CONFIG}.ddsr


# Run an sr3 declare on this configuration
sr3 declare sarra/get-radar-for-client

# This will re-declare q_feeder.${PROGRAM}.${CONFIG}.ddsr on the target broker.
```

I don't think we should let sarracenia declare anything when a configuration is disabled.

reidsunderland:
Other things that need to be changed with declare:

- It should require `--dangerWillRobinson=?` when called for more than one config (either using no arguments or if more than one config is specified)
- Wishlist: for `--users declare`, it would be nice if it checked the existing permissions and refused to change them if they don't match (adding new users/permissions from scratch is fine, but if the user exists with incorrect permissions, then it shouldn't change them)

robjarawan:
- remove ability for declare to Frankenstein queues if a config is disabled temporarly

petersilva:
"start" does not require *--dangerWillRobinson* because starting configurations is not considered *dangerous*.
Is *declare* dangerous? It has a side effect (ie. messages could accumulate) but it isn't destructive, we don't lose anything from having declares around.  I doubt it makes sense for --dangerWillRobinson to be needed for start or declare to *work* almost every time they are invoked... it will just remove the whole safeguard effect of the option.

I would think the more logical way would be to ignore *disabled* configs for start (as they already are for stop) as well as declare.   We should not need --dangerWillRobinson to do harmless stuff.

reidsunderland:
`sr3 --users declare` is dangerous because it will overwrite any non-standard RabbitMQ permissions with what sr3 thinks is correct. In a perfect world we wouldn't have any non-standard permissions, but we do for now.

andreleblanc11:
Removing the `NextRelease` tag as the threat of `sr3 declare` running on disabled configurations has been removed. I'm also lowering the priority.

[...]

PR #1713, opened 2026-06-15
#1710 - sr3 declare, Add dangerWillRobinson and prevent execution for disabled configs

andreleblanc11:
- Prevents disabled configurations from being declared when declaring one or multiple configurations.
- Adds `dangerWillRobinson` dependency

```
sr3 --debug disable subscribe/test1 subscribe/test2 

sr3 --debug --dangerWillRobinson=2 declare subscribe/test1 subscribe/test2 subscribe/test3
declare: 2026-06-15 15:49:22,505 1650690 [ERROR] root validate_dangerWillRobinson specify --dangerWillRobinson=<number> of configs to cleanup (actual: 3, given: 2 ) when cleaning more than one

# Should only prevent declare on test1 and test2, which it does.
sr3 --debug --dangerWillRobinson=3 declare subscribe/test1 subscribe/test2 subscribe/test3
...
2026-06-15 15:49:27,422 1650697 [DEBUG] sarracenia.sr declare Configuration subscribe/test1 is disabled. Skipping
2026-06-15 15:49:27,422 1650697 [DEBUG] sarracenia.sr declare Configuration subscribe/test2 is disabled. Skipping
2026-06-15 15:49:27,422 1650697 [INFO] root declare looking at subscribe/test3
2026-06-15 15:49:27,422 1650697 [DEBUG] sarracenia.sr declare Configuration subscribe/test1 is disabled. Skipping
2026-06-15 15:49:27,422 1650697 [DEBUG] sarracenia.sr declare Configuration subscribe/test2 is disabled. Skipping
2026-06-15 15:49:27,422 1650697 [INFO] root declare looking at subscribe/test3
...
```

### DIFF
--- a/sarracenia/sr.py
+++ b/sarracenia/sr.py
@@ -1322,6 +1322,7 @@ class sr_GlobalState:
         self.please_stop=False
         self.users = opt.users
         self.declared_users = opt.declared_users
+        self.has_disabled_config = False
 
         signal.signal(signal.SIGTERM, self._stop_signal)
         signal.signal(signal.SIGINT, self._stop_signal)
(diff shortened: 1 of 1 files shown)


reidsunderland:
This mostly looks good, but it works differently than other commands that deal with disabled configs, e.g. `start`. If you try to start a group of configs, and at least one is disabled, then it will refuse to start *all* the configs:

https://github.com/MetPX/sarracenia/blob/79690cc8b280cec100688ac4cfc345673deb0019/sarracenia/sr.py#L2332-L2345

It's discussed here: https://github.com/MetPX/sarracenia/issues/1159

Declare should be the same as other actions.

Note: I'm not opposed to revisiting that discussion and changing the way *all* actions work with disabled configs. I often get annoyed that there's no way to start/stop/restart all configs of one component when there's at least 1 disabled config (e.g. `sr3 restart 'sender/*'` doesn't work if there's a disabled sender). But we should be consistent. (Maybe an `--ignore_disabled` flag or something could be added)

andreleblanc11:
That's interesting and is something I didn't even consider when writing the PR. In my head, this behaviour made the most sense, but you're right it's not consistent. I personally don't like that all configurations get blocked when a single 'disabled' configuration is found from a match. I'm sure there are other people that like this feature (apparently the behaviour is based on @iMacadan's feedback). Adding an option to work around this might be a good idea.

petersilva:
fwiw, I think you guys have more experience that anybody had at the time... we were working on what we thought would make sense, not having the feature available at all... You guys have experience with it, and would know better.
Based on what you've observed, the initial idea of rejecting whole thing when some are disabled is probably bad.   We had options of what to do, and we (I?) picked that, but I think it was just the wrong choice...We could just as easily have chosen to ignore the disabled ones.

So I don't think you should feel bound by what we thought then.  You have more and better data that we had.

andreleblanc11:
I've updated the `issue1710` branch to prevent `declare` from running when any configuration comes up as `disabled`.

My plan was to add the above change to this PR. Once this PR then gets merged, submit another PR to have disabled configurations be ignored when start/declare gets invoked.

I think we can keep this PR open.

andreleblanc11:
So based on the recent feedback, I've

- Changed `has_disabled_config` variable to `self.has_disabled_config` to avoid duplicate code.
- Removed ` if len(self.filtered_configurations) < len(self.all_configs):` at the start of the `declare` method. 
   - With the `if` condition present,  running `sr3 declare all-configs-except-1` would be different from `sr3 declare` which doesn't really makes sense considering what the logic is inside of the `if ` statement.
- Changed the behaviour to mimic what `start` does. Halt the `declare` command as soon as a `disabled` configuration is found.


```
sr3 --debug --dangerWillRobinson=3 declare subscribe/test[123]
...
declare: 2026-06-16 13:57:59,726 2065760 [ERROR] sarracenia.sr declare Config subscribe/test3 is disabled. It must be enabled before declaring.
2026-06-16 13:57:59,726 2065760 [ERROR] sarracenia.sr declare No configs have been declared due to disabled configurations.

sr3 --debug --dangerWillRobinson=41 declare
declare: 2026-06-16 14:03:22,838 2066596 [ERROR] sarracenia.sr declare Config subscribe/test3 is disabled. It must be enabled before declaring.
2026-06-16 14:03:22,838 2066596 [ERROR] sarracenia.sr declare No configs have been declared due to disabled configurations.

sr3 --debug --dangerWillRobinson=2 declare subscribe/test[12]

2026-06-16 14:03:59,093 2066645 [INFO] sarracenia.moth.amqp _queueDeclare queue declared q_feeder_subscribe.test1.myserver.SomethingSomethingTest.62504462
2026-06-16 14:03:59,093 2066645 [INFO] sarracenia.moth.amqp _queueDeclare queue declared q_feeder_subscribe.test2.myserver.SomethingSomethingTest.62504462
```

reidsunderland (approved):

PR #1715, opened 2026-06-16
Fix #1714 putAccelerated errors not being caught

### LINKED ISSUE #1714: accelerated upload errors are not handled correctly
When using a binary accelerator for uploads, if the binary returns a non-zero return code, the `putAccelerated` function returns `-1`:

https://github.com/MetPX/sarracenia/blob/79690cc8b280cec100688ac4cfc345673deb0019/sarracenia/transfer/sftp.py#L553-L554

But there's no logic to check for a `-1` result. The calling code either expects an Exception to be raised, or `0` to returned for an error (only when the message contains a non-zero size field though).

https://github.com/MetPX/sarracenia/blob/79690cc8b280cec100688ac4cfc345673deb0019/sarracenia/flow/__init__.py#L2898-L2900

I think the correct fix would be to have putAccelerated raise an exception.

getAccelerated is fine.

reidsunderland:
Issue #1714 describes the problem and has example logs. If an accelerated upload failed (`putAccelerated`) the error was not caught. Flow would think the transfer succeeded and move on.

I replaced the Popen call and return code checking inside the transfer classes with a wrapper method. This reduces some repeated code and captures stderr for better error logging.

`getAccelerated` worked fine, but I changed it to use the wrapper function too. Because the download code in Flow correctly deals with `-1` returned by getAccelerated, I kept that to minimize changes (rather than throwing the Exception up to Flow like putAccelerated does).

### DIFF
--- a/sarracenia/transfer/__init__.py
+++ b/sarracenia/transfer/__init__.py
@@ -28,6 +28,7 @@ import logging
 import os
 import random
 import signal
+import subprocess
 import stat
 import sys
 import time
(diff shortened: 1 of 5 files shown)


andreleblanc11:
Shouldn't we also add logic in `flow/__init__.py` so that -1 gets recognized as an error in the `send` method?

With the current code for your use case, we will log an error, however, sarracenia will still output that the file has sent correctly. This is based on the example you've provided in the issue. I tried to recreate this scenario with your branch, but the code would later fail at the `rename` portion of the flow so it's not the exact same scenario.

Otherwise the actual code changes look good.

reidsunderland:
That's true, I was testing with `inflight None` and that's when sr3 would have incorrectly reported that the file had been sent successfully (no error, no warning).

If you use a different inflight setting, then the upload will fail silently, then the subsequent rename will also fail and trigger error handling.

So the putAccelerated errors were mostly only a problem in cases with inflight None.

andreleblanc11 (approved):
Discussed and should be good to go

PR #1717, opened 2026-06-16
Make flow tests faster

reidsunderland:
In flow tests, we use `retryEmptyBeforeExit True` in `post` configs, otherwise they will terminate while there are messages in their retry disk queues waiting to be posted.

The current implementation is that the process will (effectively) sleep until the next scheduled housekeeping. I probably didn't know what I was doing when I wrote the implementation of that, but it's much faster to just change the next scheduled housekeeping time to be a time in the past, to force housekeeping to run right away.

### DIFF
--- a/sarracenia/flow/__init__.py
+++ b/sarracenia/flow/__init__.py
@@ -644,8 +644,8 @@ class Flow:
                         f"{self.metrics['retry']['msgs_in_post_retry']} messages in the post retry queue.")
                     # Messages can't be retried before housekeeping has run, so run it right now
                     next_housekeeping = now - 1
-                    # sleep for a bit, self.o.sleep is 0 (this probably isn't needed)
-                    current_sleep = 1
+                    # sleep for a bit (self.o.sleep is 0 so using a non-zero value here)
+                    current_sleep = 0.1
                 else:
                     self.runCallbacksTime('please_stop')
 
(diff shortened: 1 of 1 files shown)


andreleblanc11 (approved):

andreleblanc11:
I checked about half a dozen `flakey broker , AMQP` flow tests on GitHub prior to this PR and they all took around 32 to 36 minutes to complete. The flakey broker on this branch completed in 22 minutes so it seems to help with minimizing the duration of the tests.

reidsunderland:
There's still a lot of failures :( I didn't expect it to improve that, but I was hoping it might. I'll wait to merge it until I've had a chance to double check that the failures we're getting now aren't different than before

andreleblanc11:
I'm not convinced those are related to this PR, given past failures on previous branches.

reidsunderland:
Yeah me too, but I want to run the tests a couple more times locally to be sure.

reidsunderland:
They should be even faster with the latest commit to reduce the sleep time. I will trigger all of them again on GitHub.

andreleblanc11:
Flakey broker tests have now dropped to 11-12 minutes execution time which is fantastic

reidsunderland:
I don't think this is failing any worse than other branches. I plan to try to debug the tests a bit more to try to find out why they fail so often.

PR #1706, opened 2026-06-10
WIS2 Improvements - 1699

andreleblanc11:
These code changes were made based on tests that I've made with the Meteo France global cache MQTT broker (globalbroker.meteo.fr).

## What's changed

- `version` is the old field that was used to determine conformance. This was replaced by `conformsTo` field and the code was adapted for this.
- The JSON data is being extracted prior to executing an `importAny` in the MQTT MOTH. This is to be able to extract the necessary fields to check for proper conformance.
- We are using the `topic` field to determine the `relPath` value inside of the message.

I've tested message ingestions from multiple different sources (UKMET and ECCC) and I didn't encounter any problems.

Here is what a message looks like when being posted in WIS2 format. The `source` field is being kept in the message.. I think we should probably remove that but I'm not sure if keeping it would cause problems.

```
2026-06-10 13:19:18,671 [INFO] sarracenia.moth.amqp putNewMessage raw message body: version: wis type: <class 'str'> {"type": "Feature", "geometry": null, "properties": {"data_id": "wis2/ca-eccc-msc/data/core/weather/experimental/SXCN40_KWAL_101311___51355", "datetime": null, "global-cache": "kr-kma-global-cache", "integrity": {"method": "sha512", "value": "V+H1+imbC7oQL77SHtaBrbbwCtRdEPZM0Ftm8rG2PR5+uiz8u9cO6PobagN0G6B1/O4xaNeehtpXsibtCf6R0g=="}, "metadata_id": "urn:wmo:md:ca-eccc-msc:02fd8740-af64-4ce0-af42-3c17c92cf02b", "source": "tfeed", "pubtime": "2026-06-10T13:12:41Z"}, "conformsTo": ["http://wis.wmo.int/spec/wnm/1/conf/core"], "id": "58d8a1fb-cda5-43cd-8843-e3a26846e0b9", "links": [{"rel": "canonical", "href": "http://my-server//net/local/home/leblanca/wis2/ca-eccc-msc/data/core/weather/experimental/SXCN40_KWAL_101311___51355.43322", "length": 181, "type": "text/plain"}]}

# I know this is posting to AMQP which is wrong for WIS2 format..
[...]

Issue #1699, opened 2026-05-22
WIS PostFormat Updates


reidsunderland:
This is the current WIS2 message format definition: https://wmo-im.github.io/wis2-notification-message/standard/wis2-notification-message-STABLE.html 

https://github.com/MetPX/sarracenia/blob/development/sarracenia/postformat/wis.py

We need to ensure our implementation is correct.

A few relevant notes from our meeting with @tomkralidis:
- WIS2 messages must have at least one link. See https://wmo-im.github.io/wis2-notification-message/standard/wis2-notification-message-STABLE.html#_1_17_links
    - the links array must have only one link of relation `canonical` (new data), `update` (update to existing data) or `deletion`
    - we will want to use this as the `retrievePath` in the sr3 message (where the data is downloaded from)
- links won't have useful information for mirroring a directory/topic structure, so we will likely want to set the `relPath` based on the WIS2 message's `data_id` or topic.

PR #1718, opened 2026-06-16
#1699 - More corrections for WIS2 (2026/06/16)

andreleblanc11:
This PR addresses 2 things.

1. I made an error in https://github.com/MetPX/sarracenia/pull/1706/changes/d55997f4d2ac8bf66330a21fd9fcfbfd0837e203. That whole `if` block only gets invoked when `messageDebugDump` is used in the configuration.. and I was using the option while testing. I was confused why the `headers['content-type']` gets set inside of that `if` block so I just moved it all out and added debug logging.

2. There's a problem when we receive the `data_id` field from the WIS MQTT broker. Often times, it exceeds the 255 char limit invoked for AMQP headers and, when posting in v02 format, causes the post to fail and get retried. I changed the code to continue instead of failing a post when a header field went missing, but that logic might be there on purpose? The change I'm porposing might be doing more harm then good.. don't know.

Ideally, we'd want to forward the `data_id` field to downstream clients if possible. It would show them where the WIS2 data source for the product lies and we might be able to leverage it in other ways.

***

I just learned that in the sarra v2, the max char length imposed for AMQP headers in the code is set to 253 chars. So we'll actually need to drop the `data_id` field from the message at least until we can confirm the messages don't get forwarded to v2 post configs.

https://github.com/MetPX/sarracenia/blob/4b19896921460971049d43c8674d85ab78a89c7e/sarra/sr_message.py#L46

I created a plugin so that the field gets dropped.

```python3
import logging
from sarracenia.flowcb import FlowCB

logger = logging.getLogger(__name__)

class Remove_long_wis2_fields(FlowCB):

    """
    Remove the message fields that come in from WIS2 brokers that are too long to be posted in v2 format.

    Context:
        We already have a work-around in place in the sr3 source code for the v02 post format to handle header fields that are over the AMQP limit (255 chars).
        However, in sarra v2, there is no workaround. The code will truncate the field if it is to long AND will drop the message before it being posted.
           v2 source code reference: https://github.com/MetPX/sarracenia/blob/4b19896921460971049d43c8674d85ab78a89c7e/sarra/sr_message.py#L756-L769

        For the v3 post format, this isn't a problem because we aren't adding these additional options as headers in the AMQP messages.

        Current WIS2 fields that are being removed due too their length being an issue:
            data_id

        Once we know WIS2 messages don't get propogated to v2 sarra, we theoretically would be able to remove this plugin from the WIS2 data source.
    """

    def __init__(self, options, logger=logger):
        super().__init__(options,logger)

    def after_accept(self, worklist):
        for msg in worklist.incoming:
            if 'data_id' in msg:
                msg['_deleteOnPost'] |= set(['data_id'])

```

### DIFF
--- a/sarracenia/moth/amqp.py
+++ b/sarracenia/moth/amqp.py
@@ -786,7 +786,7 @@ class AMQP(Moth):
                 if (type(headers[k]) is str) and (len(headers[k]) >=
                                                       amqp_ss_maxlen):
                     logger.error(f"message header {k} too long, dropping")
-                    return False
+                    continue
 
         AMQP_Message = amqp.Message(raw_body,
                                         content_type=content_type,
--- a/sarracenia/moth/mqtt.py
+++ b/sarracenia/moth/mqtt.py
@@ -607,15 +607,16 @@ class MQTT(Moth):
            decode MQTT message (protocol specific thingamabob) into sr3 one (python dictionary)
         """
         headers = { 'topic' : mqttMessage.topic }
+
+        if hasattr(mqttMessage.properties, 'ContentType'):
+            logger.debug( f'Content-type: {mqttMessage.properties.ContentType}')
+            headers['content-type'] = mqttMessage.properties.ContentType
+        else:
+            logger.debug('message is missing content-type header')
+            mqttMessage.properties.ContentType = None
+
         if self.o['messageDebugDump']:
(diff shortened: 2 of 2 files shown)


reidsunderland (approved):
It looks good.

andreleblanc11:
I have a test flow running and I'm just letting it run for a while to see if I can catch some more issues.

PR #1719, opened 2026-06-17
flow_autoconfig.sh cleanup

reidsunderland:
- Removed `python-setuptools` (Python 2) which is failing to install (we have `python3-setuptools` in the list)
- Removed duplicate apt install line
- Moved other dependencies into one line
- Removed apt-key line because it wasn't doing anything (the add-apt-repository command seems to take care of key stuff, at least on recent Ubuntu versions)

### DIFF
--- a/travis/flow_autoconfig.sh
+++ b/travis/flow_autoconfig.sh
@@ -14,15 +14,14 @@ else
 fi
 
 # Install and configure dependencies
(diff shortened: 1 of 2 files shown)


reidsunderland:
I also fixed the ssh_localhost.sh script. It would get stuck asking for a password. This never caused problems on GitHub, but it didn't work when running interactively (e.g. setting up a new VM to run tests on locally).

andreleblanc11 (approved):

PR #1728, opened 2026-06-23
refactor key filter plugin

reidsunderland:
@petersilva pointed out some useful information in https://github.com/MetPX/sarracenia/pull/1697#issuecomment-4547408494

Sundew would generate keys based on the accept regex groups in the receiver config. I've updated my plugin to allow multiple regex groups to be used as the key (before it only allowed one possible group).

Using a config like this, you can replicate the Sundew key generation:

```
acceptUnmatched False

callback accept.key_filter
# key will have an _ between each regex group
key_filter_join_str _

accept_key txt:AQ_:AIRNOW:CMQ:ASCII

directory /${1}/

# for example filename: 20070515.txt:AQ1245Z:AIRNOW:CMQ:ASCII
# key generated would be: txt:AQ_:AIRNOW:CMQ:ASCII
accept  .*SOURCE/([0-9]{2})/.*(txt:AQ).*(:AIRNOW:.*:ASCII)  key_groups=2,3
```

Key derivation is also separated into a method, so someone could subclass this plugin and change the after_accept.

### DIFF
--- a/sarracenia/flowcb/accept/key_filter.py
+++ b/sarracenia/flowcb/accept/key_filter.py
@@ -12,8 +12,16 @@ https://github.com/MetPX/sarracenia/blob/development/sarracenia/flowcb/accept/su
 The key filtering applies *after* sr3's built-in regex filtering. This plugin assumes you will use a small number
 of accept/reject statements and potentially hundreds of accept_key/reject_key statements.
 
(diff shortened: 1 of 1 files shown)


andreleblanc11 (approved):
I'm not too familiar with the new plugin, but I assume the changes are good. If not, we can always overwrite the plugin in our own separate plugin repo.

PR #1730, opened 2026-07-07
use non-buffered file objects for pipelined SFTP uploads

reidsunderland:
creating the PR to trigger flow tests on GitHub. Without this change, I sometimes see the hang in the dynamic flow test. With the change, I have not seen it yet, but that doesn't mean it's fixed. The GitHub flow tests seem to be a lot more sensitive, so I'm creating this to trigger an Actions run.

### DIFF
--- a/sarracenia/transfer/sftp.py
+++ b/sarracenia/transfer/sftp.py
@@ -483,13 +483,19 @@ class Sftp(Transfer):
             length=0):
         logger.debug(' local_file=%s remote_file=%s local_offset=%s remote_offset=%s length=%s', local_file, remote_file, local_offset, remote_offset, length)
 
+        # do not use a Paramiko buffered file when using pipelined uploads. It seems to sometimes cause hangs.
+        if self.compat_mode:
+            bufSize = self.o.bufSize
+        else:
+            bufSize = 0
+
         # simple file
 
         alarm_set(2 * self.o.timeout)
 
         try:
            if length == 0:
-               rfp = self.sftp.file(remote_file, 'wb', self.o.bufSize)
+               rfp = self.sftp.file(remote_file, 'wb', bufSize)
                rfp.settimeout(1.0 * self.o.timeout)
 
            # parts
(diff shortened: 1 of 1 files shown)


reidsunderland:
(I clicked comment too soon, 1 of the 4 dynamic_flow tests failed, but the place it failed doesn't involve SFTP transfers so it's an unrelated failure imo. The problems Andre and I have been seeing have always been with the senders that use SFTP).

This is looking promising. ~~All four of the AMQP dynamic_flow tests passed on GitHub with this change.~~ I want to do a bit more research into using non-buffered files. From what I've found so far, it doesn't seem like it's a big deal, it might even be a good thing. We are already writing the files one chunk at a time, and either way Paramiko still has to divide up the chunks/buffer into packets that it can actually send over the SSH connection (usually about 32 KB).

Using the Paramiko BufferedFile seems like unnecessary overhead to me. We read the data from the local source file into a variable, then that data gets copied into the Paramiko file's buffer, then it gets sent out over the connection. That extra copy seems pointless.

Paramiko 2.10.0 also made a change that improved the performance of BufferedFiles: https://www.paramiko.org/changelog.html#:~:text=%5BBug%5D%20%23892,few%20megabytes%20in%20size. I was testing and hitting the problem on Paramiko 2.12.0 and 3.4.0. So that change isn't related to the hang that this PR appears to fix. But we still use older versions in most of our deployments (2.0.0 on Ubuntu 18, 2.9.3 on Ubuntu 22), so not using buffering might offer a performance improvement even without pipelining.

I will test more tomorrow.

reidsunderland:
A quick test uploading a 267 MiB file.

Ubuntu 22.04, Paramiko 5.0.0 (pip installed):

- Pipelining, non-buffered file: 67 MiB/sec
- Pipelining, buffered file: 42-58 MiB/sec (this is the case that sometimes causes hangs)
- No pipelining, non-buffered file: 1.1 MiB/sec
- No pipelining, buffered file: 1.1 MiB/sec

So using a non-buffered file doesn't seem to help the performance in the non-pipelined case. With pipelining enabled, it seems slightly better and at least in my tests so far avoids the hang problem

reidsunderland:
More testing with Ubuntu 22.04, default Paramiko 2.9.3 (before buffered file fixes in 2.10):

Pipelining, non-buffered file: ~20 MiB/sec
Pipelining, buffered file: ~20 MiB/sec (this is the case that sometimes causes hangs)
No pipelining, non-buffered file: 1.1 MiB/sec
No pipelining, buffered file:  1.1 MiB/sec

So there doesn't seem to be a performance difference between using a buffered or non-buffered file in these tests. Pipelining still makes a pretty big difference, but the latest version of Paramiko is a lot faster.

andreleblanc11:
I checked out and ran the dynamic flow test on our U18.04 testing server and it passed on first attempt. 

I then ran the dynamic flow test on our U20.04 testing server without the patch, and it failed with 35/37 of the checks passing.
I then tried again with the patch, and it passed on its first attempt.

andreleblanc11 (approved):

PR #1733, opened 2026-07-15
Persist failed messages to the retry store before acknowledging

robjarawan:
##### What
Failed messages could be acknowledged to the broker before they were durably written to the retry store. If the retry-store write raised, the callback error was swallowed and the delivery was lost for good.

##### Change
Persist to the retry store before the message is acknowledged, and stop swallowing the error. When the retry feature is not present, keep the previous ack-and-continue behaviour rather than stalling intake.

##### Test
New tests in tests/sarracenia/flow/work_retry_ack_test.py cover persist-before-ack ordering, deferred ack on a persistence failure, and the retry-absent path. They fail on the current code and pass with the change. Unit CI is green.

On a permanent store failure the flow deliberately holds rather than drop (no-loss by design); shutdown then waits on pending persistence.

### DIFF
--- a/sarracenia/flow/__init__.py
+++ b/sarracenia/flow/__init__.py
@@ -92,6 +92,8 @@ class Flow:
       * worklist.ok       --> successfully processed
       * worklist.rejected --> messages to not be further processed.
       * worklist.failed   --> messages for which processing failed.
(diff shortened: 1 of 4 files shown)


robjarawan:

these never got CI.

PR #1737, opened 2026-07-15
Do not silently drop a v2 on_file False message

robjarawan:
##### What
In the v2 wrapper, on_file returning False removed an already-acknowledged message from the ok list without adding it to failed or rejected, so it vanished with no record.

##### Change
Route the message to failed so it is logged, reported, and handled instead of dropped. Acknowledgement stays idempotent (no double-ack).

##### Test
A new test asserts the message lands in failed. It fails on the current code and passes with the change. Unit CI is green.

Automatic re-processing still requires post_broker to be configured; without it the change stops the silent drop but does not re-queue.


robjarawan:

these never got CI.

PR #1738, opened 2026-07-15
Read ordinary file downloads through EOF instead of the announced length

robjarawan:
##### What
File.get truncated ordinary downloads to the announced length. A 10-byte source announced as 3 returned only 3 bytes, which the flow could accept as complete.

##### Change
Read the whole file when exactLength is not set, and keep honouring the length for genuine ranged/multipart blocks (remote_offset is only nonzero when exactLength is true). This mirrors the existing read_writelocal behaviour.

##### Test
A new parametrized test covers the full-file and ranged cases. It fails on the current code and passes with the change. Unit CI is green.

### DIFF
--- a/sarracenia/transfer/file.py
+++ b/sarracenia/transfer/file.py
@@ -132,7 +132,7 @@ class File(Transfer):
             if self.sumalgo: self.sumalgo.set_path(remote_file)
 
             # download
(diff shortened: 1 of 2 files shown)


robjarawan:

these never got CI.

PR #1741, opened 2026-07-15
Key post retries by publisher identity so reconfiguration cannot misroute

robjarawan:
##### What
The post callback stored a positional publisher_index. Removing a publisher raised IndexError with no disposition, and reordering publishers silently sent a retry to the wrong broker.

##### Change
Tag retries with a stable publisher identity (broker, exchange, topicPrefix, format) and resolve to the current index on replay. A missing publisher routes the message to failed. The stored identity is password-masked, so no credential is written to the retry queue.

##### Test
New tests cover the removed-publisher and reordered-publisher cases. They fail on the current code and pass with the change. Unit CI is green.

A message persisted by older code before upgrading can misroute once if a reorder happened across the upgrade boundary; identity that was never stored cannot be recovered.

### DIFF
--- a/sarracenia/flowcb/post/message.py
+++ b/sarracenia/flowcb/post/message.py
@@ -37,6 +37,30 @@ class Message(FlowCB):
         #else:
         #    logger.error( f"no publishers for {self.o.component}/{self.o.config}")
 
+    @staticmethod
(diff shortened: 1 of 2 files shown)


robjarawan:

these never got CI.

PR #1742, opened 2026-07-15
Fix local file and symlink polling message handling

robjarawan:
##### What
Local-file polling returned a single Message where a list was expected, so extend added the message dict keys to the flow. The symlink path referenced an unbound variable and raised, discarding the poll cycle.

##### Change
Return a single-element list for a regular file, and fix the symlink path to use the bound message variable and the correct lowercase identity key.

##### Test
New tests cover the regular-file and symlink cases. They fail on the current code and pass with the change. Unit CI is green.


robjarawan:

these never got CI.

PR #1743, opened 2026-07-15
Bound the MQTT close wait for pending publishes

robjarawan:
##### What
On close, MQTT waited in an unbounded loop for pending publishes to complete. A permanently pending message id made a child process hang forever.

##### Change
Wait up to the configured timeout, then give up with a warning; call disconnect and loop_stop on both paths so no thread leaks. A timeout of 0 is treated as unset (falls back to the default) so it does not drop pending publishes at close.

##### Test
New tests cover the timeout-bound path and the normal-drain path. Unit CI is green.


robjarawan:

these never got CI.

PR #1744, opened 2026-07-15
Give malformed geometry an explicit disposition instead of bypassing the filter

robjarawan:
##### What
Invalid geometry JSON was re-raised before the incoming list was replaced. The normal dispatcher caught it and left the message in incoming (bypassing the filter); in debug mode it aborted the run.

##### Change
Catch the JSON or type error, route the message to failed, and continue so later messages are still evaluated and valid geometry still filters normally.

##### Test
A new test confirms a malformed message goes to failed while a following valid message is still processed. It fails on the current code and passes with the change. Unit CI is green.


robjarawan:

these never got CI.

PR #1749, opened 2026-07-24
add try/except to handle race conditions in check for existing inflight file

### LINKED ISSUE #1748: crash when inflight file gets deleted, race condition
```
sarra_get-ddsr-2_25.log.2026-07-21-2026-07-21 22:47:14,963 [INFO] sarracenia.flowcb.log after_accept accepted: (lag: 2.91 )  exchange: xpublic subtopic: 20260721.AMEC.RWIN.CMML.22 a file with baseUrl: http://fdsa relPath: /20260721/AMEC/RWIN/CMML/22/OB.NB.WSP.RNBBL.20260721224345.cmml_d,3de170ade24331b02d5ba7f3cacb531d sundew_extension: PULL:RWIN:AMEC:3:CMML:20260721224656 id: PeFwreJ size: 5.5KiB (5621)
sarra_get-ddsr-2_25.log.2026-07-21-Traceback (most recent call last):
sarra_get-ddsr-2_25.log.2026-07-21-  File "/usr/lib/python3/dist-packages/sarracenia/instance.py", line 283, in <module>
sarra_get-ddsr-2_25.log.2026-07-21-    i.start()
sarra_get-ddsr-2_25.log.2026-07-21-  File "/usr/lib/python3/dist-packages/sarracenia/instance.py", line 274, in start
sarra_get-ddsr-2_25.log.2026-07-21-    self.running_instance.run()
sarra_get-ddsr-2_25.log.2026-07-21-  File "/usr/lib/python3/dist-packages/sarracenia/flow/__init__.py", line 607, in run
sarra_get-ddsr-2_25.log.2026-07-21-    self.work()
sarra_get-ddsr-2_25.log.2026-07-21-  File "/usr/lib/python3/dist-packages/sarracenia/flow/__init__.py", line 1283, in work
sarra_get-ddsr-2_25.log.2026-07-21-    self.do()
sarra_get-ddsr-2_25.log.2026-07-21-  File "/usr/lib/python3/dist-packages/sarracenia/flow/__init__.py", line 1223, in do
sarra_get-ddsr-2_25.log.2026-07-21-    self.do_download()
sarra_get-ddsr-2_25.log.2026-07-21-  File "/usr/lib/python3/dist-packages/sarracenia/flow/__init__.py", line 2027, in do_download
sarra_get-ddsr-2_25.log.2026-07-21-    how_old = time.time() - os.path.getmtime(msg['new_inflight_path'])
sarra_get-ddsr-2_25.log.2026-07-21-  File "/usr/lib/python3.10/genericpath.py", line 55, in getmtime
sarra_get-ddsr-2_25.log.2026-07-21-    return os.stat(filename).st_mtime
sarra_get-ddsr-2_25.log.2026-07-21:FileNotFoundError: [Errno 2] No such file or directory: 'OB.NB.WSP.RNBBL.20260721224345.cmml_d,3de170ade24331b02d5ba7f3cacb531d.tmp'
sarra_get-ddsr-2_25.log.2026-07-21-2026-07-21 22:49:08,930 [INFO] sarracenia.moth.amqp putSetup exchange declared: xpublic (as: amqp://feeder@/)
```

https://github.com/MetPX/sarracenia/blob/4341536c4238a3a09e85cd9d798681de9604eaa9/sarracenia/flow/__init__.py#L2024-L2039

If the file gets deleted after sarra checks that the file exists, the next step can crash.

reidsunderland:
closes #1748

We've seen this race condition cause sr3 to crash during broker patching. When the broker connection is reset, any unacked messages will likely get redelivered to another instance, but the original instance still has the message in its local worklist. Both try to process the message, if the timing is right, one of them will crash.

This wraps the existing code in a try/except to catch FileNotFoundErrors that could be raised by `os.path.getmtime` or `os.unlink`.

### DIFF
--- a/sarracenia/flow/__init__.py
+++ b/sarracenia/flow/__init__.py
@@ -2021,12 +2021,12 @@ class Flow:
 
             # assert new_inflight_path is set.
 
-            if os.path.exists(msg['new_inflight_path']):
+            if self.o.inflight and os.path.exists(msg['new_inflight_path']):
 
-                if self.o.inflight:
+                try:
                     how_old = time.time() - os.path.getmtime(msg['new_inflight_path'])
-                    #FIXME: if mtime > 5 minutes, perhaps rm it, and continue? what if transfer crashed?
-                    #       Added this with fixed value, should it be a setting?
+                    # if file exists and mtime > 5 minutes, rm it, and continue. previous transfer likely crashed
+                    #FIXME: Added this with fixed value, should it be a setting?
                     if how_old > 300:
                         os.unlink( msg['new_inflight_path'] )
                         logger.info(
@@ -2034,8 +2034,13 @@ class Flow:
                     else:
                         logger.warning(
                             f"inflight file already exists. race condition, deferring transfer of {msg['new_path']}" )
-                    self.worklist.failed.append(msg)
-                    continue
+                # in case file gets deleted after the exists check but before the getmtime or unlink calls
+                except FileNotFoundError:
+                    logger.warning(
+                        f"inflight file already existed. race condition, deferring transfer of {msg['new_path']}")
+
+                self.worklist.failed.append(msg)
+                continue
                 # overwriting existing file.
 
             # FIXME: decision of whether to download, goes here.


andreleblanc11:
This error also occurs in the case where multiple upstream processes try to post the same file on the same exchange, without being duplicate suppressed upstream.

sarracenia/flow/__init__.py
```diff
@@ -2021,21 +2021,26 @@ def do_download(self) -> None:
...
                             f"inflight file is {how_old}s old. Removed previous attempt {msg['new_path']}" )
                     else:
                         logger.warning(
                             f"inflight file already exists. race condition, deferring transfer of {msg['new_path']}" )
-                    self.worklist.failed.append(msg)
-                    continue
+                # in case file gets deleted after the exists check but before the getmtime or unlink calls
+                except FileNotFoundError:
+                    logger.warning(
+                        f"inflight file already existed. race condition, deferring transfer of {msg['new_path']}")
```
andreleblanc11, line 2040: I think it would make sense to place a `continue` inside of the except statement. That way, the message will remain unacked and another process should pick it up, while also avoiding to get the message caught in an unecessary retry loop.

petersilva (approved):
looks good.. I know you just moved existing logic, but I'm wondering if queueing it for retry later is necessary... I guess when the retry happens, it notices the announced file is already here, and gets dropped then, or if it is different it repeats the *is this new?* logic... so it's extremely likely to be harmless... it could be useful if the other guy fails to download... but... if someone else is downloading... do we need to do it also? oh... it could be a different version of the file... so checking later does look like it might cover some cases... 

I'll approve this, but I don't think I'm going to merge anything any more.  I'm too far ;-)

reidsunderland:
Thanks :) 

I was kind of unsure about that as well. The queue for retry adds a delay (at least one housekeeping interval), so I decided to keep it. This race condition is usually triggered by having two instances trying to download the same file at the same time (due to a double post at the source, or when the connection to the broker gets interrupted and un-acked messages get redelivered). That delay will hopefully allow enough time for the conflict to resolve itself, and ideally the other instance will have successfully downloaded the file in the meantime. When "this" instance comes back and retries, it will likely see that the file on disk matches what it was going to download and skip the download. 

This is what we've observed in ops anyways, except the delay in that case is added by the instance crashing and being restarted later by sanity, not the retry code.

I'll let @andreleblanc11 decide when to merge it because he's working on the 3.03.00 release. I think we want this fix added to the next RC but he can coordinate the merges.

andreleblanc11:
I think we should merge it for this release because it can cause an instance to crash. Not everyone knows/runs `sr3 sanity` so its probably better to be safe.

andreleblanc11 (approved):

petersilva:
merging since André blessed it.

sarracenia/flow/__init__.py
petersilva: continue means looping while there is a conflict:

Using a continue:

I don´t know if hops far enough outside the logic ... like will it check if the file that the process wrote is different? I think it might just try to over-write it regardless, which would be unhelpful.

It also guarantee to slow the rate of all transfers down because it is going to keep trying to download until it gives up.  so no other file will be transferred while waiting for the conflict to be resolved.  queuing for retry allows other transfers to succeed and make it very unlikely that, the next time it checks, there is still a conflict.

If we want to try again, I think queueing for a retry is the better strategy for overall efficiency of all file delivery, though that file might be more slowly delivered in some rare cases.

reidsunderland: It falls through to the worklist.failed.append and continue a bit below :)

Issue #1756, opened 2026-08-05
transfers break on filenames containing spaces


robjarawan:
##### What

Every accelerated transfer command is assembled by string substitution and then split on
whitespace, so any path containing a space becomes two arguments. The escaping that was
meant to prevent this cannot work, because the split happens on the next line and
`subprocess.Popen` is handed a list, so there is no shell to interpret a backslash.

Affects all four accelerated protocols, not just sftp. Split out of #1753.

##### Reproduction

This is the code path, run standalone so the failure is visible without a server:

```python
import os
base = 'sarra@host'
arg1 = base + ':' + '/data' + os.sep + 'my file.dat'
arg1 = arg1.replace(' ', '\\ ')                      # transfer/sftp.py:414
cmd = '/usr/bin/scp %s %d'.replace('%s', arg1)
cmd = cmd.replace('%d', './my file.dat').split()     # transfer/sftp.py:419
print(cmd)
```

```
['/usr/bin/scp', 'sarra@host:/data/my\\', 'file.dat', './my', 'file.dat']
```

Five arguments where three were intended. scp sees three sources and a destination, and
either errors or writes to the wrong name.

In a running flow the same thing happens with `accelThreshold` set and a product whose
filename has a space in it.

##### Why it happens

`sarracenia/transfer/sftp.py:413-419`:

```python
arg1 = arg1.replace(' ', '\\ ')
...
cmd = self.o.accelScpCommand.replace('%s', arg1)
cmd = cmd.replace('%d', arg2).split()
```

The `.split()` splits on the same space the backslash was added to protect. Even without
the split, `runAccelCommand` calls `subprocess.Popen(cmd)` with a list and no `shell=True`,
so a backslash would be passed through to scp as a literal character rather than being
consumed as an escape.

The destination argument is not escaped at all, so a space in the local filename breaks it
regardless.

Same construction, unescaped, in:

```
sarracenia/transfer/sftp.py:565-571   putAccelerated
sarracenia/transfer/file.py:151-152   accelCpCommand
sarracenia/transfer/https.py:229-231  accelWgetCommand
sarracenia/transfer/ftp.py:363-364    accelFtpgetCommand
sarracenia/transfer/ftp.py:487-488    accelFtpputCommand
```

##### How likely is this in practice

Needs accelerated transfers enabled and a filename containing a space. Spaces are not
common in the meteorological products I have seen, but `sarracenia/__init__.py` goes to the
[...]

Issue #1755, opened 2026-08-05
accelerated sftp fails when baseUrl contains a path


robjarawan:
##### What

`getAccelerated` builds the scp source argument by stripping only the scheme from
`baseUrl`, so any path in the baseUrl ends up on the host side of the colon. `scp` then
sees no valid `host:path`, treats the whole thing as a local path, and fails.

Split out of #1753, where it turned up while testing something else. Independent of how
keys or ports are configured.

##### Reproduction

Environment: sr3 `development` (3.03.00rc1), an sftp server reachable as `user@host`.

Post with a baseUrl carrying a path, and a file large enough to cross `accelThreshold`:

```
post_baseUrl sftp://sarra@host/home/sarra
```

```
accelThreshold 1M
```

Observed:

```
[INFO]  sarracenia.transfer.sftp getAccelerated accel_sftp:
        /usr/bin/scp sarra@host/home/sarra:/home/sarra/data/big.dat ./big.dat
[ERROR] sarracenia.transfer.sftp getAccelerated failed:
        cp: cannot stat 'sarra@host/home/sarra:/home/sarra/data/big.dat': No such file or directory
```

The same baseUrl downloads correctly with `accelThreshold 0`, so it is specific to the
accelerated path.

##### Why it happens

`sarracenia/transfer/sftp.py:410-414`:

```python
base_url = msg['baseUrl'].replace('sftp://', '')
if base_url[-1] == '/':
    base_url = base_url[0:-1]
arg1 = base_url + ':' + self.pwd + os.sep + remote_file
```

`replace('sftp://', '')` removes the scheme and nothing else, so `base_url` is
`sarra@host/home/sarra` rather than `sarra@host`. Since the first `/` comes before the
first `:`, scp reads the argument as a local path, which is why the error mentions `cp`.

`putAccelerated` at line 565 builds `arg2` the same way from `self.o.sendTo`.

##### How likely is this in practice

Only reachable when accelerated transfers are enabled, which needs `accelThreshold` set
above zero, and only for baseUrls that carry a path. Both are ordinary enough. The failure
mode is unpleasant because the same config works below the threshold, so it presents as
"large files fail" rather than as a configuration error.

##### Possible approaches

1. Parse the baseUrl properly and use only the netloc for the host part, appending the path
   component to the remote path rather than the host.
2. Reject or warn when an accelerated transfer is asked for with a baseUrl that has a path,
   rather than building a command that cannot work.
[...]

PR #1757, opened 2026-08-05
doc #1753 sftp: prefer ~/.ssh/config over ssh_keyfile in credentials.conf

### LINKED ISSUE #1753: Accelerated SFTP transfers ignore `ssh_keyfile`
The accelerated SFTP code doesn't use the key file specified by `ssh_keyfile` in `credentials.conf`. 

If you are using a non-standard key file name (e.g. `id_ecdsa_for_specific_client`), it won't work.

You can workaround it by changing the command and hardcoding the key file path: `accelScpCommand /usr/bin/scp -i /local/home/sarra/.ssh/id_ecdsa_for-specific_client %s %d`, but it would be nice if a variable could be substituted instead.

robjarawan:
##### What

Closes #1753 as documentation, per the discussion there.

`ssh_keyfile` in `credentials.conf` is used by ordinary sftp transfers and silently ignored
by accelerated ones. Transfers over `accelThreshold` are handed to `scp`, which reads
`~/.ssh/config` and cannot see `credentials.conf`. The symptom is a configuration that works
until a file crosses the threshold, so it presents as "large files fail" rather than as a
credential problem.

Settings in `~/.ssh/config` are read by both paths: `scp` natively, and paramiko because
`Sftp.credentials()` looks the host up in `~/.ssh/config` itself and takes `HostName`,
`User`, `Port` and `IdentityFile` from it.

##### Change

Adds a `SFTP and ~/.ssh/config` section to `sr3_credentials.7.rst` documenting a `Host`
stanza with no `credentials.conf` entry as the way to configure sftp, and points to it from
the `ssh_keyfile` entry under Credential Details.

It also records two things that are easy to lose an afternoon to:

- a port belongs in the stanza, not the URL. A port in the URL is not passed to `scp`
  correctly, so an accelerated transfer to `sftp://user@host:2222/` fails while an ordinary
  transfer to the same URL succeeds.
- `~/.ssh/config` is only consulted when the credential does not already answer the question
  (`transfer/sftp.py:285-286`). An entry carrying a user and a password takes precedence and
  the stanza is not read, so omitting the entry is the reliable way to have it apply.

Documentation only. No code changes.

##### Verified

Every statement in the new section was run against a real sftp server on unmodified
`development`, not reasoned about. Lab account whose `authorized_keys` holds only a
deliberately odd named key, so nothing succeeds by accident, plus a second sshd on 2222 for
the port case. 2 MB file, `accelThreshold 1M`, md5 compared against the source on the
successful runs.

| setup | ordinary | accelerated |
|---|---|---|
| `ssh_keyfile` in credentials.conf | works | fails, `Permission denied (publickey)` |
| ssh_config stanza, no credentials.conf entry | works | works |
| ssh_config stanza with `Port 2222`, alias in URL | works | works |
| port in the URL, key in credentials.conf | works | fails, `2222:/path: No such file` |

RST checked with docutils. The only diagnostics are the pre-existing `|today|` and
`|release|` substitutions that sphinx supplies at build time.

##### Not included

The French `docs/source/fr/Reference/sr3_credentials.7.rst` carries the matching
`ssh_keyfile` entry and is not updated here. I would rather that was done by someone who
writes French properly than guessed at, but I can attempt it if you would prefer it in the
same change.

Two unrelated defects turned up while testing this and are filed separately, both with
reproductions and no patch: #1755 (accelerated sftp fails when baseUrl contains a path) and
#1756 (accelerated transfers break on filenames containing spaces, across all five
accelerated call sites).

##### Fork CI

Run on my fork first: https://github.com/robjarawan/sarracenia/pull/136 - 32 pass, 6 fail.

The six failures are runner flakiness, not this change. The clearest evidence is that the
duplicate workflow runs on the same commit disagree with themselves:

```
dynamic_flow on ubuntu-22.04     2 pass, 1 fail
dynamic_flow on ubuntu-24.04     2 pass, 1 fail
flakey_broker on ubuntu-22.04    2 pass, 1 fail
flakey_broker on ubuntu-24.04    2 pass, 1 fail
Maintenance test on ubuntu-22.04 1 pass, 1 fail
Maintenance test on ubuntu-24.04 1 pass, 1 fail
```

Every job that failed also passed on another run of the same commit. The same jobs are
also currently failing on untouched `development` upstream.

This branch changes one .rst file and a changelog line, with no python or config touched,
so the flow tests cannot be reached by it.

A passing dynamic_flow run: https://github.com/robjarawan/sarracenia/actions/runs/30972432267/job/92199482505

### DIFF
--- a/docs/source/Reference/sr3_credentials.7.rst
+++ b/docs/source/Reference/sr3_credentials.7.rst
@@ -72,7 +72,9 @@ You may need to specify additional options for specific credential entries. Thes
 
 Supported details:
 
-- ``ssh_keyfile=<path>`` - (SFTP) Path to SSH keyfile
+- ``ssh_keyfile=<path>`` - (SFTP) Path to SSH keyfile. For SFTP, prefer an entry in
+  ``~/.ssh/config`` instead, see `SFTP and ~/.ssh/config`_ below. ``ssh_keyfile`` is not
+  seen by the ``scp`` command used for accelerated transfers.
 - ``passive`` - (FTP) Use passive mode
 - ``active`` - (FTP) Use active mode
 - ``binary`` - (FTP) Use binary mode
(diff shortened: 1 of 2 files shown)


andreleblanc11:
The changes are good. Just need french changes as well.

docs/source/Reference/sr3_credentials.7.rst
```diff
@@ -101,6 +103,61 @@ Note::
...
+
+A port in the URL is not passed on to ``scp`` correctly, so an accelerated transfer to
+``sftp://sarra@host:2222/`` fails while an ordinary transfer to the same URL succeeds.
+
+Note::
+ Sarracenia only consults ``~/.ssh/config`` when the credential does not already answer
+ the question: when no user is known, or when neither a key nor a password was supplied.
+ A ``credentials.conf`` entry carrying a user and a password takes precedence and the
+ stanza is not read. Omitting the entry is the reliable way to have ``~/.ssh/config``
+ apply.
```
reidsunderland, line 158: I'm not 100% sure how this works, I think `.ssh/config` is always used, I don't know which overrides which (does credentials.conf override `.ssh/config` or is it the other way around?)

robjarawan:
I verified every claim in this doc change against `sftp.py` and `flow/__init__.py`. The two-path model and the port-in-stanza guidance check out. Three amendments before merge.

First, the subscriber stanza problem is real and the doc needs a warning. `options.sendTo` is overwritten by `msg['baseUrl']` (`flow/__init__.py:2259`), so on a subscriber the stanza `Host` must match the publisher's announced hostname, or there must be a matching `credentials.conf` entry. Aliasing local config alone has no effect on downloads, so please scope the wording that way.

Second, the precedence note must read user plus a secret. Per `sftp.py:285-286` the stanza lookup is skipped when a user plus a keyfile or password is known, so a user plus `ssh_keyfile` suppresses it too, not just user plus password.

Third, `IdentitiesOnly yes` affects `scp` only. Sarracenia reads just `HostName`, `User`, `Port`, `IdentityFile` (`sftp.py:288-296`), first `IdentityFile` only, so please note both.

French translation stays a follow-up. The port-mangling wording is accurate for both get and put paths.

PR #1758, opened 2026-08-06
some fixes for messages without size header.

### LINKED ISSUE #1754: Senders crash when using `fileSizeMax` and when a message is received without the size set
https://github.com/MetPX/sarracenia/blob/1e0e80d8497e4332e066fc8c597842f4f4fee441/sarracenia/flow/__init__.py#L3067-L3069

```
self.running_instance.run()
File "/usr/lib/python3/dist-packages/sarracenia/flow/__init__.py", line 607, in run
     self.work()
File "/usr/lib/python3/dist-packages/sarracenia/flow/__init__.py", line 1283, in work
     self.do()
File "/usr/lib/python3/dist-packages/sarracenia/flow/sender.py", line 31, in do
     self.do_send()
File "/usr/lib/python3/dist-packages/sarracenia/flow/__init__.py", line 3062, in do_send
     if self.o.fileSizeMax > 0 and msg['size'] > self.o.fileSizeMax:
 KeyError: 'size'
```

petersilva:
close #1754

Normally, a notification message should include the size of the file to be transferred. Following Postel's maxim, we want to handle it being missing elegantly:

* when the field is missing, don't crash #1754
* When we finish downloading a file, we know it's actual size, and if the field is missing it should be added.

### DIFF
--- a/sarracenia/flow/__init__.py
+++ b/sarracenia/flow/__init__.py
@@ -1376,7 +1376,8 @@ class Flow:
     def write_inline_file(self, msg) -> bool:
         """
            write local file based on a message with inlined content.
-
+           
+           side effect: msg['size'] will be set to the correct size of the downloaded file.
         """
         # make sure directory exists, create it if not
         if not os.path.isdir(msg['new_dir']):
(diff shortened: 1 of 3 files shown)


andreleblanc11:
The logic looks good to me. I think it also makes sense to set the message size if the source does not provide one.

reidsunderland (approved):
I agree, it looks good

PR #1759, opened 2026-08-18
AMQP1.0 tweak and NAV CANADA message format

reidsunderland:
- Adjusts the AMQP1.0 connection names (NAVCAN requested that we remove MetPX from the connection name, it will like cause confusion with their version of Sundew that they refer to as MetPX)
- Add NAV CANADA message format


sarracenia/postformat/navcanada.py
```diff
@@ -0,0 +1,303 @@
...
+                ISO:   2026-08-18T08:13:08.399Z or # 2026-08-18T08:13:08.399
+        """
+        if isinstance(time, str):
+            time = time.replace('Z', '')
+            if 'T' in time and ':' in time and '-' in time:
+                dt = datetime.strptime(time, '%Y-%m-%dT%H:%M:%S.%f')
+                return dt.strftime("%Y%m%dT%H%M%S.%f")[:-3]
+            else:
+                try:
+                    dt = datetime.fromtimestamp(int(mt), tz=timezone.utc)
```
andreleblanc11: Should `mt` be `time` instead? It doesn't look defined inside of the method.

sarracenia/postformat/navcanada.py
```diff
@@ -0,0 +1,303 @@
...
+                    return dt.strftime("%Y%m%dT%H%M%S.%f")[:-3]
+                except Exception as e:
+                    # ERROR message below will be logged
+                    logger.debug(f"{e}", exc_info=True)
+
+        elif isinstance(time, int):
+            dt = datetime.fromtimestamp(int(mt), tz=timezone.utc)
```
andreleblanc11: Should `mt` be `time` instead? It doesn't look defined inside of the method.

sarracenia/postformat/navcanada.py
```diff
@@ -0,0 +1,303 @@
...
+
+        # Set topic / DESTINATION
+        # Normally, this message format will be used in combination with a plugin that sets msg['topic']
+        # and topicDerive will return that value. When topic is returned from the message, the topicPrefix is
+        # *not* already in the topic, so we add it here. If topic is not in the message, it will be derived from
+        # the relPath or msg['subtopic'], and topicPrefix *will* be added. We have to handle both possibilities.
+
+        # topic must be uppercase, but the topic:// part can't be
+        post_topic = '/'.join(PostFormat.topicDerive(sr3_msg, options)).upper()
+        if post_topicPrefix.upper() in post_topic:
```
andreleblanc11, line 213: Why are we doing this if we are already doing a generalized `.upper()` on the whole post_topic on the line before? Seems redundant.

sarracenia/postformat/navcanada.py
robjarawan: I tried both timestamp forms listed in the docstring. This uses `mt`, which isn’t defined in this function. A numeric string gets caught and replaced with the current time, while an integer raises `NameError`. That means the sample `NCFILESHARE_FILE_MTIME` loses its actual timestamp.

Can we parse the value passed into the function and add tests for the numeric string, integer, and ISO forms? This one needs fixing before merge.

sarracenia/postformat/navcanada.py
```diff
@@ -0,0 +1,303 @@
...
+            logger.error("message missing MSG_PUBLISH_TIME, using current time")
+            msg['pubTime'] = sarracenia.nowstr()
+
+        # NOTE: we currently do not expect to receive messages with URLs from NC, inline data only
+        msg['relPath'] = ''
+        # build relPath from DESTINATION and NCFILESHARE_FILE_NAME, so we can at least have a file path and
+        # name to work with (mirror) when writing the data from the message.
+        if 'DESTINATION' in headers:
+            msg['relPath'] += headers['DESTINATION']
+        if msg['relPath'][-1] != '/':
```
robjarawan: `DESTINATION` is treated as optional just above, but when it’s missing, `relPath` is still empty and this index raises `IndexError`. Since `mine()` claims any recognized `MSG_TYPE`, this becomes an AMQP decode failure and the receiver acknowledges the message as bad.

Can we either validate `DESTINATION` as required before decoding, or safely build a filename-only path when it’s absent? A missing-header test would catch this.

sarracenia/postformat/navcanada.py
```diff
@@ -0,0 +1,303 @@
...
+            if decoded_payload:
+                msg['content'] = {
+                    'encoding': 'utf-8',
+                    'value': decoded_payload
+                }
+                # FIXME: sr3 bug: File "sarracenia/flow/__init__.py", line 1424, in write_inline_file
+                #                 if ((msg['size'] > 0) and len(data) != msg['size']):
+                #                 KeyError: 'size'
+                # inline data download does not work when size is not set
+                msg['size'] = len(decoded_payload)
```
robjarawan: This needs to store the byte length, not the number of decoded characters. I tried `é`: the incoming body is two bytes, but this records a size of one, so `write_inline_file()` rejects it after encoding it again. A non-UTF-8 body is currently returned without any content at all.

Could we preserve the decompressed bytes, use base64 for non-UTF-8 content like `Message.putContentInline()` does, and calculate `size` from those bytes? If NAV CANADA guarantees UTF-8 only, we should validate and test that explicitly.

sarracenia/postformat/navcanada.py
```diff
@@ -0,0 +1,303 @@
...
+        contentType = sr3_msg.get('contentType', 'text/plain')
+
+        # Set schema info when the data is IWXXM
+        if 'iwxxm' in raw_body:
+            ns_start = raw_body.find('xmlns:iwxxm="')
+            if ns_start > 0:
+                try:
+                    schema_ns = raw_body[ns_start:].split('"')[1]
+                    schema_ver = schema_ns.split('/')[-1]
+                    headers['MSG_NAMESPACE'] = schema_ns
```
robjarawan, line 288: The NAV CANADA input example in this file uses `MSG_SCHEMA_NAMESPACE`, but the exporter sends `MSG_NAMESPACE`. Which application-property name does NAV CANADA expect from us? Can we align this with the agreed format and lock it down with an outbound fixture?

robjarawan (changes requested):
Went through the AMQP1 change and exercised the new NAV CANADA format directly. The connection-name change itself looks reasonable. I left the runtime issues inline; I think those need fixing before this merges.

##### Tests

The full suite passes locally: 359 passed and 1 skipped. However, this PR doesn’t add any NAV CANADA tests, and the existing `amq1_test.py` collects zero tests. Can you add focused import/export tests covering the timestamp formats, required headers, UTF-8 and binary bodies, gzip, outbound application properties, and the new connection name?

A sanitized message fixture matching what NAV CANADA expects would be useful here. This is basically the contract for the new format and would have caught the timestamp and namespace problems.

##### Minor cleanup

`git diff --check` reports trailing whitespace in the new file, and `pycodestyle` finds the blank-line and indentation issues there. There’s also a `print()` in the gzip path that should use the logger, and its error message currently says it failed to read a SWIM message.

Since `post_format navcanada` is now selectable, could you also add a short config example showing the required `post_topicPrefix` and related message fields?

sarracenia/postformat/navcanada.py
```diff
@@ -0,0 +1,303 @@
...
+        else:
+            logger.error("message missing MSG_PUBLISH_TIME, using current time")
+            msg['pubTime'] = sarracenia.nowstr()
+
+        # NOTE: we currently do not expect to receive messages with URLs from NC, inline data only
+        msg['relPath'] = ''
+        # build relPath from DESTINATION and NCFILESHARE_FILE_NAME, so we can at least have a file path and
+        # name to work with (mirror) when writing the data from the message.
+        if 'DESTINATION' in headers:
+            msg['relPath'] += headers['DESTINATION']
```
robjarawan, line 105: I traced this through the real `Flow.filter()` path. With `mirror True` and `DESTINATION ../../escape/`, SR3 builds a target outside the configured directory. Since `DESTINATION` comes from the AMQP message, can we reject `.` and `..` path components and verify the final path stays under the configured root? I would add a containment test for this one too.

sarracenia/postformat/navcanada.py
```diff
@@ -0,0 +1,303 @@
...
+                logger.warning(f"failed to parse pubTime {sr3_msg['pubTime']}")
+                pubTime = datetime.now()
+        else:
+            pubTime = datetime.now()
+        headers['MSG_PUBLISH_TIME'] = pubTime.isoformat()[:-3] + 'Z'
+
+        headers['UUID'] = str(uuid.uuid4())
+
+        if 'content' in sr3_msg and sr3_msg['content']:
+            raw_body = sr3_msg['content']['value']
```
robjarawan, line 265: I tried the normal SR3 representation for binary inline content: `{'encoding': 'base64', 'value': '/wA='}` represents `b'\xff\x00'`. `exportMine()` sends the literal string `'/wA='` instead of the original bytes. Can we decode the SR3 content using `getContent()` or the equivalent before posting it? I would also fail the post when content is missing instead of warning and publishing an empty body, since the NAV CAN message body is the file.

sarracenia/postformat/navcanada.py
```diff
@@ -0,0 +1,303 @@
...
+            decompressed_payload = payload
+            decoded_payload = None
+            # Detect if the payload is gzipped
+            try:
+                # FIXME: i think we should support gzip as an inline content encoding and unzip somewhere else
+                # in the code, but leaving this here for now.
+                # content_encoding is only mandatory when compression is used
+                if 'amqp1_content_encoding' in headers and headers['amqp1_content_encoding'] == "gzip":
+                    if payload[:2] == b'\x1f\x8b':  # GZIP magic number
+                        decompressed_payload = gzip.decompress(payload)
```
robjarawan, line 147: I tried a highly compressible payload here. 4,892 compressed bytes became 5,000,000 bytes before SR3 got to `fileSizeMax` or filtering. Basically a small broker message can make the worker allocate a much larger buffer. Can we decompress with a hard expanded-size limit and reject anything over it? The 30 MB NAV CAN limit looks like the natural ceiling if that is the actual contract.

sarracenia/postformat/navcanada.py
robjarawan: One more edge from the adversarial pass: `body=b''` skips content entirely because of `if body`, leaving `baseUrl=none://` without content or size. A valid zero-byte file will then take the missing-download path. Please cover an empty payload explicitly too.

sarracenia/postformat/navcanada.py
reidsunderland: That's a good catch, I changed `mt` to `time`

sarracenia/postformat/navcanada.py
reidsunderland: This should be fixed now

robjarawan (changes requested):
I went through the two new commits. The timestamp forms work now, and a missing `DESTINATION` safely falls back to the filename. The UTF-8 size is also using the byte count now. I reran the full suite: 359 passed and 1 skipped.

I’m still able to reproduce the remaining inline points on `0ff06e02`: path traversal through `DESTINATION`, dropped binary and zero-byte bodies, literal base64 on export, and unbounded gzip expansion. The schema property also still needs confirmation, and there aren’t any NAV CANADA tests yet.

I think those still need fixing before merge. Ready for another look once they’re updated.

sarracenia/postformat/navcanada.py
reidsunderland: This transforms e.g. `TOPIC://ECCC/AZUREQA/WEATHER/TAC/V1/NT/CWAO/UA/01/` to `topic://ECCC/AZUREQA/WEATHER/TAC/V1/NT/CWAO/UA/01/`. NAV CANADA requires that the topic is uppercase, but that `topic://` prefix has to be lowercase.

reidsunderland:
- Path traversal through `DESTINATION`: 
  - e.g. with `..` in the DESTINATION/path.
  - This can happen with any PostFormat so I won't deal with it in the NavCanada code. 
  - It's a good discovery and a separate issue should be created. I think the correct solution would be changes to the main Flow code to handle it. There may be cases where it's useful/desirable, but I can see how there's potential for abuse, so I think an option would be needed to allow users to re-enable it.
- Fixed `mt` variable issue in time parsing.
- Waiting on NC to confirm the specifics for the inline content. Their document says: "Message payload is in appropriate encoding (UTF-8 or binary with correct x-opts)". I'm asking them if we can only expect to receive UTF-8 encoded messages, or to provide more info about non-UTF-8 cases.
- The schema stuff is correct.
- NAV CANADA has a validation/test suite to confirm the messages we are publishing conform to their requirements, and we passed their tests.

reidsunderland:
NC confirmed that we will only deal with UTF-8 data, so I removed the non-UTF-8 stuff and added better error handling for messages that fail to import or export

Issue #1768, opened 2026-08-28
clarify filtering documentation


reidsunderland:
https://metpx.github.io/sarracenia/Reference/sr3_options.7.html#:~:text=In%20the%20above%20snippet,higher%20in%20the%20file..

After upgrading to 3.03, which includes a fix where Sundew extensions are now included in the URL when filtering, we came across a few configs with accept statements that no longer matched the URLs with Sundew extensions because the accepts ended in `$`.

The documentation also has a FIXME note:

<img width="832" height="835" alt="Image" src="https://github.com/user-attachments/assets/67918f7f-c64e-4908-910b-963415d343e8" />

The case we ran into was:

```
accept .*.csv$
```

which worked on 3.02 when URLs were `http://something.com/2026.csv` but not once the sundew extension was added, e.g. `http://something.com/2026.csv:EXTENSION:...`.

---

The purpose of this issue is to:

1. Document clearly that Sundew extensions **are** included when filtering.
2. Resolve the FIXME
3. Potentially improve/clean up all parts of the sr3 Options document that relate to accept, reject and filtering

We use regex **match**:

> **`re.match(pattern, string, flags=0)`**
> If zero or more characters at the beginning of string match the regular expression pattern, return a corresponding [Match](https://docs.python.org/3/library/re.html#re.Match). Return None if the string does not match the pattern; note that this is different from a zero-length match.
> - https://docs.python.org/3/library/re.html#re.match

This means that `.*` is usually required at the beginning of an accept/reject statement, but not at the end. If you want to match something specific at the end of the line, you need to terminate with `$`.

```
accept .*csv
# accepts:
# ...something.csv
# ...something.csv_even_stuff_at_the_end

accept .*csv.*
# effectively the same as above

accept .*csv$
# only accepts ...something.csv
# not something.csv.with_stuff_at_the_end
```

This might be obvious to regex experts, but it is worth having it clearly documented for anyone who is not an expert, and for the sake of documenting the expected behaviour.

robjarawan:
I went through this against current development and both implementations. The main 3.02 to 3.03 point is right, but I think we should tighten the issue before someone picks it up.

The examples should use `.*\.csv` because the unescaped `.` matches any character. I’d also call this the string used for accept/reject filtering rather than saying the URL itself changes. The Python flow conditionally appends `:<sundew_extension>` to that matching string; it doesn’t modify the notification URL.

[...]

PR #1769, opened 2026-08-28
docs: clarify accept and reject filtering

mmaxjr:
Clarifies how accept/reject filtering is applied across Python flows and C components.

This PR documents that Python flows match against a temporary filtering string, while C components match their component-specific filtered values. It also explains first-match behavior, explicit end-of-pattern matching, and Sundew extension handling as legacy compatibility metadata.

The cleanup/reorganization remains tracked in #1768.

Checked with:

```sh
git diff --check
python -m pytest tests/sarracenia/flow/__flow___test.py
```

### DIFF
--- a/docs/source/Explanation/CommandLineGuide.rst
+++ b/docs/source/Explanation/CommandLineGuide.rst
@@ -971,7 +971,8 @@ accept, reject and accept_unmatch
 - **baseUrl_relPath   <boolean> (default: False)**
 
 The  **accept**  and  **reject**  options process regular expressions (regexp).
-The regexp is applied to the the notification message's URL for a match.
+They are interpreted in order, and the first matching **accept** or **reject**
+rule wins.
 
 If the notification message's URL of a file matches a **reject**  pattern, the notification message
 is acknowledged as consumed to the broker and skipped.
@@ -999,8 +1000,32 @@ sequence #2::
   reject .*\.gif
 
 
-In sequence #1, all files ending in 'gif' are rejected.  In sequence #2, the accept .* (which
-accepts everything) is encountered before the reject statement, so the reject has no effect.
+In sequence #1, all files whose filtered value matches ``.*\.gif`` are rejected. In sequence #2, the
+``accept .*`` rule is encountered first and accepts everything, so the later reject statement has no effect.
+
+Python flows build a temporary filtering string from ``baseUrl + relPath``. When
+a ``sundew_extension`` header is present, and the URL has fewer than three
+colons, that temporary filtering string may have ``:<sundew_extension>``
+appended for legacy compatibility. The notification URL itself is not modified.
+The C components do not append the separate ``sundew_extension`` header:
+``cpost`` filters the pathname and ``cpump`` filters ``relPath``.
+
+Because Python flow matching starts at the beginning of the filtering string,
+filters that look for text anywhere in that string usually begin with ``.*``.
+There is no implicit ``.*`` at the end of the pattern. For example::
+
+  accept .*\.csv$
+  # matches only a filtered value ending in .csv
+
+  accept .*\.csv:.*
+  # matches a Python flow filtering string with a Sundew extension after .csv
+
+  accept .*\.csv$|.*\.csv:
+  # matches either a plain .csv ending or a .csv followed by an extension boundary
+
+Prefer filtering on the path when possible. Treat ``sundew_extension`` as legacy
+compatibility metadata and include it in filters only when the path alone is not
+specific enough.
 
 It is best practice to use server side filtering to reduce the number of notification messages sent
 to the component to a small superset of what is relevant, and perform only a fine-tuning with the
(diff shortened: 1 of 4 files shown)


robjarawan (commented):
I think this needs a few changes before it merges.

##### Scope this to the component doing the filtering

The options page is shared by both implementations, so saying the patterns use Python `re.match` isn’t true for `cpost` and `cpump`. The Python flows compile with `re` and call `Pattern.match()`. The C components use POSIX `regexec()` instead; `cpost` filters the pathname, `cpump` filters `relPath`, and neither appends the separate `sundew_extension` header.

Could we document those separately so the leading `.*` advice isn’t applied to the C components?

##### Describe the filtering string, not a changed URL

For Python flows, the code builds a temporary `urlToMatch` from `baseUrl + relPath`. It conditionally appends `:<sundew_extension>` when the header exists and the URL has fewer than three colons. It doesn’t modify the notification URL itself.

Can we describe this as the string used for accept/reject filtering and avoid saying the extension is always included?

##### Keep the published copies in sync

This currently updates only `docs/source/Reference/sr3_options.7.rst`. The same FIXME and filtering guidance also appear in the English command-line guide and both French copies.

Could we update all four sources here? Otherwise we’ll publish conflicting explanations depending on which page someone reads.

One other contradiction: the new note correctly says `.*\.gif` also matches `image.gif2`, but the following paragraph still says the rule rejects files “ending in gif.” I’d also use `.*\.csv` and `.*\.csv$` consistently in the examples and state directly that the first matching accept/reject rule wins.

I ran `git diff --check`, the focused filtering tests, and the documentation build. The changed page renders correctly.

robjarawan:
Small correction to my earlier example after Peter’s note on #1768: `.*\.csv` is too loose because it also matches something like `.csvhithere.txt`.

If the filter is specifically for messages carrying a Sundew extension, `.*\.csv:.*` makes the colon boundary explicit. If it needs to accept both a plain `.csv` filename and `.csv:<sundew_extension>`, we can use `.*\.csv$|.*\.csv:`.

I’d also describe `sundew_extension` as legacy compatibility metadata and recommend avoiding it in filters when the path alone is enough. The Python/C and temporary match-string distinctions from my earlier review still apply.

mmaxjr:
Updated the PR to address the review feedback:

- scoped the regex wording so Python flows and C components are documented separately
- described the temporary Python filtering string instead of saying the notification URL is modified
- updated the English reference page, English command-line guide, and both French copies
- replaced the `.*\.gif` wording that implied an end-of-string match
- changed the CSV examples to use explicit `.csv`, `.csv:`, and `.csv$` boundaries
- noted that the first matching `accept` or `reject` rule wins

Validation:
- `git diff --check`
- `python -m pytest tests/sarracenia/flow/__flow___test.py`

I also tried a local Sphinx build. It starts, but this local environment is missing `pandoc`, so it stops while processing `Contribution/Philosophy/AboutTime.ipynb` before reaching these changed pages.

robjarawan:
Thanks, I went through the updated diff. The Python/C split, temporary filtering string, first-match behaviour, and four published copies are all covered now. The focused filtering tests pass and the docs build completes here.

I found one remaining consistency pass before I approve. The new text correctly explains that the patterns operate on a component-specific filtering value, but the same pages still say `accept` and `reject` match the URL in a few places. The command-line guides also still say `.*cap` means “ends with CAP”; that example needs `.*cap$`, since the new note correctly explains that there’s no implicit end-of-string match.

Could we replace the remaining URL references with “filtering string” or “filtered value” as appropriate, and change that ending example to `.*cap$` in both the English and French guides? Then I think this is ready.

mmaxjr:
Updated the consistency pass:

- replaced the remaining URL-based accept/reject wording in the edited sections with filtered value/filtering string wording
- changed the CAP ending examples to use `.*cap$` / `.*CAP$`
- kept the changes in sync across the English reference, English guide, French reference, and French guide

Checked with:

```sh
git diff --check
python -m pytest tests/sarracenia/flow/__flow___test.py
```

reidsunderland:
The changes so far also look good to me. There is duplication between the command line guide and sr3 options documents.

The other part of #1768 is to clean up the documentation and remove duplication. There's a lot of overlap between the command line guide and sr3 options related to filtering. I think we need to resolve that, but this PR can focus on clarification and we can create another one to work on cleanup/reorgination.

robjarawan (approved):
Looks good to me now. I checked `6b48694`; the four copies are in sync, the focused tests pass, `git diff --check` is clean, and the docs build completes here.

I agree with keeping the exhaustive details in `sr3_options` and handling the command-line guide duplication separately.

One housekeeping item before merge: the PR description still mentions only `re.match` and says Sundew extensions are included in filtered URLs. It also closes #1768, but I don’t see a separate issue tracking the cleanup yet. Could we update the summary and either create/link the cleanup issue or leave #1768 open?

The existing `.*?cap` wording can go into that cleanup pass. It isn’t a blocker for this PR.

mmaxjr:
Updated the PR description to reflect the current scope and avoid closing #1768.

docs/source/Explanation/CommandLineGuide.rst
```diff
@@ -999,8 +1000,32 @@ sequence #2::
...
+There is no implicit ``.*`` at the end of the pattern. For example::
+
+  accept .*\.csv$
+  # matches only a filtered value ending in .csv
+
+  accept .*\.csv:.*
+  # matches a Python flow filtering string with a Sundew extension after .csv
+
+  accept .*\.csv$|.*\.csv:
+  # matches either a plain .csv ending or a .csv followed by an extension boundary
```
andreleblanc11, line 1024: Are we sure this works? I'm 99% this should work when encaspulated in parantheses, i.e. `accept (.*\.csv$|.*\.csv:)`

andreleblanc11: I just did a manual test and ` accept .*\.csv$|.*\.csv:` works.

PR #1773, opened 2026-09-05
fix(flow): resolve sender path and inflight staging

### LINKED ISSUE #1772: Flow.send dry_run fails on files in local_dir and unbound len_written
##### What

Ran into this while testing a new sender configuration directly in stages with `dry_run True` before enabling actual uploads. `Flow.send` fails to transfer files in two ways:

1. Changing into `local_dir` is skipped when `dry_run` is set (`if (curdir != local_dir) and not self.o.dry_run:`). The subsequent check tests `os.path.exists(local_file)` using the relative basename against the daemon working directory rather than `local_dir`, logging `file <file> does not exist in local dir <local_dir>, can't send (baseDir not set?)` and aborting with return code 0.
2. When `inflight` is `None`, `inflight[0] == '.'`, or inplace blocks are used, the branch under `if not self.o.dry_run:` does not assign `len_written` on dry run. Line 2914 checks `len_written`, raising `UnboundLocalError`.

##### Reproduction

Environment: sr3 from `development` (3.03.00rc1), sender configuration with `dry_run True`.

```bash
mkdir -p /tmp/sender_repro/source /tmp/sender_repro/dest
echo "payload" > /tmp/sender_repro/source/data.txt

cat > ~/.config/sr3/sender/repro.conf <<EOF
sendTo file:///tmp/sender_repro/dest
baseDir /tmp/sender_repro/source
dry_run True
inflight None
EOF
```

Run from any directory other than `/tmp/sender_repro/source`:

```
sr3 foreground sender/repro
```

Observed:

```
[ERROR] sarracenia.flow send file data.txt does not exist in local dir /tmp/sender_repro/source, can't send (baseDir not set?)
```

If the daemon is started in `/tmp/sender_repro/source`, the chdir failure is bypassed, and it fails with:

```
[ERROR] sarracenia.flow send Delivery failed /tmp/sender_repro/dest/data.txt
[DEBUG] sarracenia.flow send Exception details:
  File "sarracenia/flow/__init__.py", line 2914, in send
    if msg['size'] > 0 and len_written == 0:
UnboundLocalError: local variable 'len_written' referenced before assignment
```

##### Why it happens

In `sarracenia/flow/__init__.py`:

```python
if (curdir != local_dir) and not self.o.dry_run:
    try:
        os.chdir(local_dir)
```

Skipping `os.chdir(local_dir)` leaves the process in whatever working directory the daemon was started in. Later:

```python
if not os.path.exists(local_file):
    logger.error(f"file {local_file} does not exist in local dir {local_dir}, can't send (baseDir not set?)")
```

`local_file` is `os.path.basename(local_path)`. Since the process cwd was not changed to `local_dir`, the relative path check fails.

Second, in the upload/inflight block:

```python
if inflight == None or (('blocks' in msg) and (msg['blocks']['method'] != 'inplace')):
    try:
        if not self.o.dry_run:
            if accelerated:
                len_written = self.proto[self.scheme].putAccelerated(msg, local_file, new_file)
            else:
                len_written = self.proto[self.scheme].put(msg, local_file, new_file)
```

There is no `else: len_written = msg['size']` branch (unlike `inflight == '.'` or `umask`). Later:

```python
if msg['size'] > 0 and len_written == 0:
```

`len_written` is checked without having been bound.

##### How likely is this in practice

Affects any sender daemon configured with `dry_run True` where `baseDir` differs from the daemon working directory, or where `inflight` is unset (`None`), `.ext`, or inplace. When staging and verifying new sender configurations in place before enabling live uploads, `dry_run` fails to simulate sending any files.

##### Possible approaches

1. Check `local_path` defensively, preserve caller working directory across live and dry-run modes, and assign `len_written = msg['size']` on dry run across all inflight branches.
2. Initialize `len_written = msg.get('size', 0) if self.o.dry_run else 0` beforehand, in addition to assigning it in each dry-run branch.

robjarawan:
When running sender with dry_run enabled, Flow.send skipped changing directory to local_dir. Relative path existence checks against local_file failed when the daemon working directory differed from local_dir, preventing test executions from completing successfully. Furthermore, staging directories were prepared during fileOp events, partitioned block transfers triggered invalid renames of non-existent temporary paths, short final blocks were rejected during inplace transfers, and non-string inflight values caused unhandled exceptions.

Fixes: https://github.com/MetPX/sarracenia/issues/1772
Refs: https://github.com/robjarawan/sarracenia/pull/138

#### What this adds

Flow.send resolves local_path using os.path.realpath and verifies that os.path.commonpath([base_dir, local_path]) == base_dir. Any attempted path traversal outside base_dir logs an error and returns -1.

Flow.send preserves the original process working directory file descriptor prior to directory modification and restores it in a finally: block across live and dry run modes.

Inflight options enforce string validation. Directory staging paths starting with / or ending in / take precedence over file suffixes. Absolute staging paths disable accelerated transfer and are rejected on object storage backends. Staging directory creation and rename capability checks are deferred to the payload transfer section so fileOp events (remove, rename, directory, link) bypass staging operations.

Partitioned block transfers bypass staging and temporary rename steps. Inplace block transfers extract offsets and lengths from msg['blocks'] manifests, take precedence over inflight is None, and use msg['size'] as the expected written length for short final blocks. In sarracenia/transfer/sftp.py, file truncation during partial block writes is removed so out-of-order block assembly succeeds.

Written byte counts returned by transfer backends are validated against expected_length before executing renames or writing success report codes.

#### Blast radius and rollback

This change modifies transfer staging, directory switching, and write validation logic in sarracenia/flow/__init__.py and sarracenia/transfer/sftp.py.

Destructive operations:
- Staging directory preparation creates remote directories using cd_forced when configured with inflight directory paths.
- Temporary files are renamed to final destination paths upon payload transfer completion.
- Files and directories are removed during remove fileOp events.

If transfer validation fails, files remain in temporary staging paths without replacing existing target files. Rollback is accomplished by reverting the commit on the branch. Reviewers should focus on the staging directory preparation placement in sarracenia/flow/__init__.py lines 2735-2755 and block transfer precedence in lines 2820-2850.

#### How to verify

Execute the flow test suite:

    python3 -m pytest tests/sarracenia/flow/

Verify formatting and line length:

    git diff origin/development..HEAD | pycodestyle --diff --max-line-length=119

#### Evidence

<details><summary>Platform results</summary>

```
============================= test session starts ==============================
platform linux -- Python 3.10.12, pytest-9.0.2, pluggy-1.6.0
rootdir: /net/local/home/jarawanr/src/sarracenia/tests
configfile: pytest.ini
plugins: mock-3.15.1, depends-1.0.1, anyio-4.13.0
collected 13 items

tests/sarracenia/flow/__flow___test.py ....                              [ 30%]
tests/sarracenia/flow/callback_dispatch_test.py ........                 [ 92%]
tests/sarracenia/flow/metrics_leak_test.py .                             [100%]

======================== 13 passed, 2 warnings in 1.65s ========================
```

</details>

### DIFF
--- a/sarracenia/flow/__init__.py
+++ b/sarracenia/flow/__init__.py
@@ -3,6 +3,7 @@ import glob
 import importlib
 import logging
 import os
+import posixpath
 import re
 
 # v3 plugin architecture...
@@ -2531,11 +2532,11 @@ class Flow:
         """
 
         self.o = options
-        sendTo=self.o.sendTo 
-        start_time=time.perf_counter()
+        sendTo = self.o.sendTo
+        start_time = time.perf_counter()
         logger.debug('%s_transport sendTo: %s', self.scheme, sendTo)
         logger.debug('%s_transport send %s %s', self.scheme, msg['new_dir'], msg['new_file'])
- 
+
         if len(self.plugins['send']) > 0:
             ok = False
             for plugin in self.plugins['send']:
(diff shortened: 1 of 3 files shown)


robjarawan:
I gave this PR a thorough review. Two early concerns did not hold up on verification (`local_offset` is always 0 on senders, so the old code KeyError'd; over-length tolerance is required by real FTP drivers), so I am dropping those.

My sign-off: mergeable with two small conditions, both one-line comments. At `flow/__init__.py:3012`, the rename-fail `return 0` is retryable by contract, so do not escalate it to -1, just note the retry intent. At `:2560`, note the parent-dir-only realpath resolve still blocks dir and `..` escapes via the `commonpath` check at `:2567`.

One explicit drop: do not restore the blanket `rfp.truncate()` in `sftp.py` `put`. Verification showed it fires per block and re-breaks the out-of-order assembly this PR fixes, and `transfer/__init__.py:397-398` already covers the short final. The shrink-overwrite tail case stays open as a follow-up, gated on final-block-only truncation.

Riskiest remaining item: `flow/__init__.py:3000` returns 0 on persistent short writes with only a 0.01s sleep path elsewhere, so a stuck writer hot-retries with no poison cap. Worth a follow-up issue, not a blocker.

PR #1776, opened 2026-09-08
Restore the working directory after Flow work

### LINKED ISSUE #1774: SR3 Flow retains payload directory as process cwd while idle, blocking safe retention
#### Summary

Long-running SR3 Flow workers can retain a payload/output directory as the process current working directory after message processing has finished. The worker can remain idle in that directory for days.

This caused a production DDI-to-SAN migration to fail its safety fence twice on 2026-09-06/07. The migrator is just a set of scripts that uses `rsync` to move each dated tree from `public_data` into SAN space for temporary archiving. It gets stuck when a process is still attached to the directory through its `cwd`, so the dated tree is not moved. The payload had copied successfully and passed a full checksum comparison both times, but the migration could not safely rename and retire the local date tree while an SR3 process still held that tree as its `cwd`.

This is not an open-file leak: the affected process had zero file descriptors and zero memory mappings under the date tree. The retained process `cwd` was the only namespace anchor.

#### Environment

- SR3: `3.03.00rc2`
- Component: long-running `sarra` Flow worker
- Platform: Linux
- Upstream `development` checked at `446a826bc3c30e6523775809dd85d1544b583c67`
- Downstream consumer: DDI-to-SAN retention migrator for `/apps/sarra/public_data/YYYYMMDD`

#### Production evidence

The worker entered the dated output directory while processing a message:

```text
2026-09-04 19:54:42,368 [DEBUG] sarracenia.flow do_download chdir /apps/sarra/public_data/20260904/.../19
2026-09-04 19:54:42,571 [DEBUG] sarracenia.flow do_download chdir /apps/sarra/public_data/20260904/.../19
```

The same worker was still alive and anchored there more than 57 hours later:

```text
PID:          9298
Started:      2026-08-12 12:42:05 UTC
Observed:     2026-09-07
Command:      SR3 sarra Flow worker
/proc/9298/cwd -> /apps/sarra/public_data/20260904/.../19
FDs under /apps/sarra/public_data/20260904:   0
maps under /apps/sarra/public_data/20260904: 0
```

The first DDI-to-SAN attempt copied the multi-terabyte date tree and completed a checksum comparison, then refused the unsafe cutover:

```text
2026-09-06T04:32:28Z rsync_start day=20260904
2026-09-06T08:12:08Z rsync_done elapsed=13180s rc=0
2026-09-06T15:47:47Z verify: rsync comparison clean checksum=1 day=20260904
2026-09-06T15:50:53Z fence: process namespace anchor permits writes after rename anchor=/proc/9298/cwd target=/apps/.ddi_public_data_migrator/quarantine/20260904/.../19
2026-09-06T15:50:53Z fence: restored canonical source after pre-promotion failure day=20260904
2026-09-06T15:50:53Z STATUS subcommand=migrate result=source_busy exit=13
```

The automatic retry updated the staging copy and again passed checksum verification. It was blocked by the same PID and the same retained `cwd`:

```text
2026-09-06T16:01:15Z rsync_start day=20260904
2026-09-06T16:46:57Z rsync_done elapsed=2742s rc=0
2026-09-07T00:08:21Z verify: rsync comparison clean checksum=1 day=20260904
2026-09-07T00:11:18Z fence: process namespace anchor permits writes after rename anchor=/proc/9298/cwd target=/apps/.ddi_public_data_migrator/quarantine/20260904/.../19
2026-09-07T00:11:18Z fence: restored canonical source after pre-promotion failure day=20260904
```

The migration's refusal is intentional. A Linux process whose `cwd` is inside a directory tree retains a reference to that directory across a rename. A later relative-path write can therefore land in the renamed/quarantined source tree instead of the new canonical path. Ignoring the anchor would risk losing late data when the quarantined source is removed.

This is not isolated to one worker. A read-only `/proc/*/cwd` scan on the same node found SR3 workers anchored in four different date trees:

```text
1   20260904
2   20260905
3   20260906
215 20260907
```

Archived DEBUG logs recorded 129,196 SR3 `chdir` operations into the current date tree on that node on 2026-09-04. The problem becomes visible when a low-volume worker receives no later accepted message and therefore never happens to change directory again.

#### Current upstream code

The current `development` branch changes the process-wide working directory in several Flow paths and does not restore a neutral/stable directory before returning to the long-running loop:

- [`do_download`: `os.chdir(msg['new_dir'])`](https://github.com/MetPX/sarracenia/blob/446a826bc3c30e6523775809dd85d1544b583c67/sarracenia/flow/__init__.py#L1977-L1985)
- [download path: `os.chdir(new_dir)`](https://github.com/MetPX/sarracenia/blob/446a826bc3c30e6523775809dd85d1544b583c67/sarracenia/flow/__init__.py#L2240-L2255)
- [send path: `os.chdir(local_dir)`](https://github.com/MetPX/sarracenia/blob/446a826bc3c30e6523775809dd85d1544b583c67/sarracenia/flow/__init__.py#L2576-L2586)

The first call directly matches the production DEBUG record and retained `/proc/9298/cwd` value.

#### Reproduction

1. Start a long-running SR3 `sarra` configuration with DEBUG logging and a destination below a disposable path such as `/tmp/sr3-cwd-repro/20260904/product/19`.
2. Deliver one accepted message and wait for `sarracenia.flow do_download chdir /tmp/sr3-cwd-repro/20260904/product/19`.
3. Allow the worker to become idle without delivering another accepted message.
4. Resolve its process working directory:

   ```bash
   readlink /proc/PID/cwd
   ```

5. Observe that it remains `/tmp/sr3-cwd-repro/20260904/product/19` rather than a stable SR3 runtime/configuration directory.
6. Rename `/tmp/sr3-cwd-repro/20260904` to `/tmp/sr3-cwd-repro/20260904.quarantine` and resolve `/proc/PID/cwd` again. The process remains anchored below the renamed tree and retains the ability to make relative-path writes there.

Control: after the worker exits, or after it explicitly changes to a neutral directory, the `/proc/PID/cwd` anchor is gone and the retention fence can safely proceed.

#### Expected behavior

An SR3 worker should not retain a payload, source, or destination data directory as its process `cwd` between messages. When a work item finishes, fails, is rejected, or raises through a callback, the worker should return to a stable runtime directory.

#### Suggested direction

Prefer eliminating process-global `chdir()` from transfer paths and using absolute paths. If relative-path transfer APIs require it, save a stable working directory and restore it in `finally` blocks on every success, reject, retry, and exception path.

A regression test should run the real Flow lifecycle, process a message, let the worker idle, and assert that `/proc/self/cwd` is not within the message's payload tree. The test should cover successful download, successful send, rejected work, and callback/transfer exceptions.

#### Operational impact

- A checksum-verified DDI-to-SAN migration was blocked twice by the same idle SR3 worker.
- The safe migrator restored the canonical source both times, so there is no evidence of data loss or corruption.
- Local retention could not reclaim the old date while the anchor remained, increasing `/apps` capacity risk and forcing repeated multi-hour verification/retry work.
- Restarting the affected worker clears the immediate anchor, but this is only a workaround; any SR3 Flow using these paths can recreate it.

robjarawan:
I changed `Flow.work()` to restore its starting working directory after the work cycle. Transfers currently leave the worker inside the payload directory, so an idle worker can stay attached to an old dated tree and block the retention checks described in #1774.

The existing work and acknowledgement order stays inside a `try/finally`. If the starting directory is no longer available, the worker falls back to `cfg_run_dir`, then filesystem root. Recovery preserves the original processing exception.

The [reproduction config, callback and commands](https://github.com/MetPX/sarracenia/issues/1774#issuecomment-5565675301) launch a real `sr3 foreground` worker, write an inline payload, leave it idle, and rename the dated tree. On base `24de015c`, the worker stays inside that tree and follows the rename. With the fix, it returns to its runtime directory and stays outside the renamed tree.

The [isolated DEV-pump check requested by Andre](https://github.com/robjarawan/sarracenia/pull/171#issuecomment-5591066286) confirmed that behavior at `2b92e2c4`. A second inline message also wrote the correct content, the first payload remained intact, and cleanup completed. This was one worker without a broker; the full migrator and a long-running soak were not tested.

I prepared this submission on upstream `development` at `321272b3`, carrying only the cwd fix and regression tests from robjarawan/sarracenia#171. The Flow implementation and test file are byte-identical to the DEV-tested fork head. No subscription or CI changes from the fork are included.

The seven focused Flow tests pass on this submission branch. The combined Flow, v2wrapper, logging and retry test selection reports 16 passed and 9 skipped. The earlier fork unit run recorded 378 passed and 1 skipped; [fork PR #171](https://github.com/robjarawan/sarracenia/pull/171) retains the broader flow results and outstanding CI failures. Those earlier results are not a claim that upstream CI on this new branch has passed.

For an already-affected worker, start from a stable directory outside payload storage and use a full `sr3 stop` / `sr3 start`. The in-place resource restart preserves its working directory, so a package upgrade alone does not clear an existing attachment.

#### Update, 8cb27aa

Reworked the restore loop after review. The `for`/`else` is replaced with a plain `restored` flag, and the `raise` is gone: `run()` does not catch exceptions from `work()`, so raising out of a `finally` would have killed the instance without `close()` and left its pidfile behind. An unrestorable working directory now logs and calls `runCallbacksTime('please_stop')`, matching how `run()` already handles `messageCountMax`. `work_succeeded` only existed to gate that raise and is removed.

The `restore_error is None` case that prompted the review was not reachable — `fallback_dir` is never None, since a `cfg_run_dir` that is not an absolute existing directory is replaced by the filesystem root, so at least two candidates are always attempted and the failure branch is only reached when every one raised. The block was hard to read regardless.

Focused Flow tests: 7 passed, and they still fail 5 of 7 when the restore is disabled.

Fixes #1774

### DIFF
--- a/sarracenia/flow/__init__.py
+++ b/sarracenia/flow/__init__.py
@@ -1280,6 +1280,14 @@ class Flow:
 
     def work(self) -> None:
 
+        work_dir = os.getcwd()
+        try:
+            self._work()
+        finally:
+            os.chdir(work_dir)
+
+    def _work(self) -> None:
+
         self.do()
 
         # need to acknowledge here, because posting will delete message-id
--- a/sarracenia/flow/__init__.py
+++ b/sarracenia/flow/__init__.py
@@ -1280,48 +1280,76 @@ class Flow:
 
     def work(self) -> None:
 
-        work_dir = os.getcwd()
+        fallback_dir = getattr(self.o, 'cfg_run_dir', None)
+        if not isinstance(fallback_dir, str) or not os.path.isabs(fallback_dir) \
+                or not os.path.isdir(fallback_dir):
+            fallback_dir = os.path.abspath(os.sep)
+
         try:
-            self._work()
+            work_dir = os.getcwd()
+        except OSError as ex:
+            work_dir = None
+            logger.warning("working directory is unavailable before work: %s; using %s", ex, fallback_dir)
+            try:
+                os.chdir(fallback_dir)
+            except OSError as fallback_error:
+                fallback_dir = os.path.abspath(os.sep)
+                logger.warning("failed to use configured working directory: %s; using %s", fallback_error,
+                               fallback_dir)
+                os.chdir(fallback_dir)
+
+        work_succeeded = False
+        try:
+            self.do()
+
+            # need to acknowledge here, because posting will delete message-id
+            self.ack(self.worklist.ok)
+            self.ack(self.worklist.rejected)
+            self.ack(self.worklist.failed)
+
+            # adjust message after action is done, but before 'after_work' so adjustment is possible.
+            post_messages=[]
+
+            for m in self.worklist.ok:
+                if len(self.o.publishers) <= 1: # save creation of new messages (a lot of space & time savings.)
+                    self.work_message_adjust(m)
+                    m['publisher_index'] = 0
+                else: # replace output messages with 1 per publishing destination.
+                    i=0
+                    for p in self.o.publishers:
+                        new_m=sarracenia.Message()
+                        new_m.copyDict(m)
+                        new_m['publisher_index'] = i
+                        new_m.updatePaths( self.o, m['new_dir'], m['new_file'], i )
+                        self.work_message_adjust(new_m)
+                        post_messages.append(new_m)
+                        i += 1
+                m['_deleteOnPost'] |= set(['publisher_index'])
+
+            if len(self.o.publishers) > 1:
+                self.worklist.ok=post_messages
+
+            self._runCallbacksWorklist('after_work')
+
+            self.ack(self.worklist.rejected)
+            self.worklist.rejected = []
+            self.ack(self.worklist.failed)
+            work_succeeded = True
         finally:
-            os.chdir(work_dir)
-
-    def _work(self) -> None:
-
-        self.do()
-
-        # need to acknowledge here, because posting will delete message-id
-        self.ack(self.worklist.ok)
-        self.ack(self.worklist.rejected)
-        self.ack(self.worklist.failed)
-
-        # adjust message after action is done, but before 'after_work' so adjustment is possible.
-        post_messages=[]
-
-        for m in self.worklist.ok:
-            if len(self.o.publishers) <= 1: # save creation of new messages (a lot of space & time savings.)
-                self.work_message_adjust(m)
-                m['publisher_index'] = 0
-            else: # replace output messages with 1 per publishing destination.
-                i=0
-                for p in self.o.publishers:
-                    new_m=sarracenia.Message()
-                    new_m.copyDict(m)
-                    new_m['publisher_index'] = i
-                    new_m.updatePaths( self.o, m['new_dir'], m['new_file'], i )
-                    self.work_message_adjust(new_m)
-                    post_messages.append(new_m) 
-                    i += 1
-            m['_deleteOnPost'] |= set(['publisher_index'])
-                    
-        if len(self.o.publishers) > 1:
-            self.worklist.ok=post_messages
-    
-        self._runCallbacksWorklist('after_work')
-
-        self.ack(self.worklist.rejected)
-        self.worklist.rejected = []
-        self.ack(self.worklist.failed)
+            restore_error = None
+            for restore_dir in [work_dir, fallback_dir, os.path.abspath(os.sep)]:
+                if restore_dir is None:
+                    continue
+                try:
+                    os.chdir(restore_dir)
+                    break
+                except OSError as ex:
+                    restore_error = ex
+                    logger.warning("failed to restore working directory to %s: %s", restore_dir, ex)
+            else:
+                if work_succeeded:
+                    raise restore_error
+                logger.error("failed to restore working directory after work failed: %s", restore_error)
 
 
 
--- a/sarracenia/flow/__init__.py
+++ b/sarracenia/flow/__init__.py
@@ -1298,7 +1298,6 @@ class Flow:
                                fallback_dir)
                 os.chdir(fallback_dir)
 
-        work_succeeded = False
         try:
             self.do()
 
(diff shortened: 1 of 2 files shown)


sarracenia/flow/__init__.py
```diff
@@ -1280,40 +1280,76 @@ def work_message_adjust(self,m):
...
+                    continue
+                try:
+                    os.chdir(restore_dir)
+                    break
+                except OSError as ex:
+                    restore_error = ex
+                    logger.warning("failed to restore working directory to %s: %s", restore_dir, ex)
+            else:
+                if work_succeeded:
+                    raise restore_error
```
reidsunderland: maybe add a check for restore_error is None? inside the else? I'm not sure if it's possible to reach that, but at least ask AI to be sure :)

andreleblanc11: Could also just pass a basic default exception.

```
>>> restore_error = Exception("Something went wrong.")
>>> raise ex
Exception: Something went wrong.
>>>
```

robjarawan: It is not reachable. `fallback_dir` is never None — a `cfg_run_dir` that is not an absolute existing directory is replaced by the filesystem root — so at least two candidates are always attempted, and the failure branch is only reached when every attempt raised, which means `restore_error` was always bound.

Changing it anyway. Three people have stopped on this block now, so the construct is the problem. Replaced the `for`/`else` with a plain `restored` flag and dropped the `raise`: `run()` does not catch exceptions from `work()`, so raising out of a `finally` would kill the instance without `close()` and leave its pidfile behind. It now logs and asks the flow to stop cleanly, the same way `run()` already handles `messageCountMax`. `work_succeeded` only existed to gate that raise, so it is gone too.

Pushed as 8cb27aa. Focused flow tests still 7 passed, and I checked they still fail 5 of 7 when the restore is disabled.

sarracenia/flow/__init__.py
```diff
@@ -1280,40 +1280,78 @@ def work_message_adjust(self,m):
...
+                    break
+                except OSError as ex:
+                    restore_error = ex
+                    logger.warning("failed to restore working directory to %s: %s", restore_dir, ex)
+            if not restored:
+                # Raising here would escape a finally, and run() does not catch
+                # it, so the instance would die without close() and leave its
+                # pidfile behind. Ask for an orderly stop instead.
+                logger.error("failed to restore working directory: %s", restore_error)
+                self.runCallbacksTime('please_stop')
```
andreleblanc11, line 1354: I don't think invoking a please_stop here is better than raising an error.

Invoking please_stop means we are expecting the process to die gracefully, which also means the configuration will likely be in a stopped state. This might be adding confusion for an analyst trying to figure out what is the root cause of an issue.

This would also mean sr3 sanity wouldn't pick up the process as it wouldn't be in a 'missing' or 'hung' state. Leaving the pid file intact is exactly what allows sr3 sanity to determine when a a process is supposed to be running when it has crashed.

reidsunderland: I agree, a clean shut down is not a good way to handle an error or unexpected state here. We'd rather crash and let sanity restart it.

reidsunderland:
I'm also not sure about the idea of setting a fallback dir to `/`. I feel like that's risking putting files in a weird place. 

It would be quite disastrous for `getattr(self.o, 'cfg_run_dir', None)` to not return a valid directory. I don't really think that's a possibility you should be accommodating for. If that returns nothing, I'd rather have it log an error and not attempt any chdiring at all. 

Imo, if cfg_run_dir is not set then the process is in a bad state alreadyand cd'ing to `/` only puts in it a more abnormal state.

PR #1779, opened 2026-09-10
Bound HTTP requests in built-in poll callbacks

### LINKED ISSUE #1778: Built-in HTTP poll callbacks do not bound requests
Several built-in poll callbacks make HTTP requests without an explicit timeout. I found this while creating a poll that uses the EUMETSAT website, then checked the other built-in HTTP poll implementations for the same pattern.

On current `development` at `321272b3866f41823d662f5312e6b702378e4204`, direct `requests.get()` calls without `timeout=` occur in:

- [`poll.eumetsat`](https://github.com/MetPX/sarracenia/blob/321272b3866f41823d662f5312e6b702378e4204/sarracenia/flowcb/poll/eumetsat.py#L149-L171): browse and product-detail requests
- [`poll.odata`](https://github.com/MetPX/sarracenia/blob/321272b3866f41823d662f5312e6b702378e4204/sarracenia/flowcb/poll/odata.py#L188-L196): catalogue and pagination requests
- [`poll.nasa_cmr`](https://github.com/MetPX/sarracenia/blob/321272b3866f41823d662f5312e6b702378e4204/sarracenia/flowcb/poll/nasa_cmr.py#L270-L278): catalogue requests, plus [checksum metadata requests](https://github.com/MetPX/sarracenia/blob/321272b3866f41823d662f5312e6b702378e4204/sarracenia/flowcb/poll/nasa_cmr.py#L340-L346)
- [`poll.copernicus_marine_s3`](https://github.com/MetPX/sarracenia/blob/321272b3866f41823d662f5312e6b702378e4204/sarracenia/flowcb/poll/copernicus_marine_s3.py#L89-L115): product and dataset STAC requests
- [`poll.airnow`](https://github.com/MetPX/sarracenia/blob/321272b3866f41823d662f5312e6b702378e4204/sarracenia/flowcb/poll/airnow.py#L34-L44): hourly file request

There are also direct `urllib.request.urlopen()` calls without `timeout=` in the USGS, NOAA hydrometric, and NEXRAD poll callbacks. Those should be included when deciding the scope of a common fix.

Requests does not impose a timeout unless one is supplied. SR3 has a `timeout` option, with a default of 300 seconds, but these callbacks do not pass it to their HTTP operations. A `stop_requested` check between requests cannot interrupt a request already waiting for a response.

#### Reproduction

The following bounded reproduction uses `poll.eumetsat`, where the problem was first observed. The local endpoint accepts the request and deliberately does not send a response. The test calls the real `Eumetsat.poll()` method and confirms that it reached the endpoint.

```python
#!/usr/bin/env python3
import multiprocessing
import threading
import time
from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer
from types import SimpleNamespace

import requests
from sarracenia.flowcb.poll.eumetsat import Eumetsat


class StalledHandler(BaseHTTPRequestHandler):
    def do_GET(self):
        self.server.request_seen.set()
        time.sleep(30)

    def log_message(self, *args):
        pass


def run_poll(url):
    callback = object.__new__(Eumetsat)
    callback.o = SimpleNamespace(
        timeNowMinus=0,
        pollUrl=url + "/---COLLECTION_ID---/",
        post_baseUrl=url + "/download/",
        acceptMediaType=[],
    )
    callback._encoded_collectionIds = ["TEST"]
    callback._cid_placeholder = "---COLLECTION_ID---"
    callback.stop_requested = False
    callback.poll()


server = ThreadingHTTPServer(("127.0.0.1", 0), StalledHandler)
server.request_seen = threading.Event()
threading.Thread(target=server.serve_forever, daemon=True).start()
url = f"http://127.0.0.1:{server.server_port}"

process = multiprocessing.Process(target=run_poll, args=(url,))
process.start()
assert server.request_seen.wait(2), "poll did not reach the local endpoint"
process.join(2)
print(f"built-in poll returned within 2 seconds: {not process.is_alive()}")
if process.is_alive():
    process.terminate()
    process.join()

started = time.monotonic()
try:
    requests.get(url + "/control", timeout=0.25)
except requests.exceptions.ReadTimeout:
    print(f"timeout control returned in {time.monotonic() - started:.2f} seconds")
finally:
    server.shutdown()
```

Observed on the revision above:

```text
built-in poll returned within 2 seconds: False
timeout control returned in 0.25 seconds
```

This establishes the failure without depending on an external provider or waiting for a production outage.

#### Expected behavior

Every built-in poll HTTP operation should have a bounded, configurable connect/read timeout. The common SR3 `timeout` option may be suitable if it is intended to govern poll HTTP requests; otherwise the callbacks need a clearly named common option. Timeout exceptions should be handled without terminating the flow.

A regression should exercise at least one `requests` callback and one `urllib` callback through their poll entry points, and verify that the configured value reaches every request path. Individual callbacks with multiple paths, such as EUMETSAT browse/details and NASA CMR catalogue/checksum, also need coverage for each path.

A per-request timeout fixes the indefinite wait. A total poll deadline may still be useful for callbacks that paginate or make one detail request per product, but that is a separate bound and does not need to be part of the first fix.

robjarawan:
## What

Built-in poll callbacks can wait indefinitely when an HTTP provider or network path stops responding. This was found while configuring an EUMETSAT poll and confirmed across the AirNow, Copernicus Marine, EUMETSAT, NASA CMR, NEXRAD, NOAA hydrometric, OData, and USGS callbacks.

## Change

Pass the existing SR3 `timeout` setting to each direct `requests.get()` and `urllib.request.urlopen()` call in those callbacks. The change stays in the built-in polls and does not add a new option, retry policy, batching behavior, or shared abstraction.

Closes #1778.

## Validation

Nine regression tests cover all 14 changed request paths, including EUMETSAT browse/details, OData pagination, NASA catalogue/checksum, both Copernicus STAC requests, both USGS modes, and all NOAA hydrometric requests.

The tests fail against the base development commit `1c738c8ae` because the requests receive no timeout, and pass at `439b13dc8`.

- `pytest -q tests/sarracenia/flowcb/poll tests/sarracenia/config_test.py tests/sarracenia/flow/poll_test.py`: 25 passed
- `pytest -q tests/sarracenia/flowcb --ignore=tests/sarracenia/flowcb/wistree_test.py`: 132 passed, 11 skipped
- `git diff --check`: passed
- `python3 -m compileall -q sarracenia/flowcb/poll tests/sarracenia/flowcb/poll`: passed

A complete local `pytest tests` collection also requires the optional `GTStoWIS2` and `proton` modules, which are not installed in this checkout.


andreleblanc11 (changes requested):
These are not the same changes as in https://github.com/robjarawan/sarracenia/pull/173...

They include some? of the changes in https://github.com/MetPX/sarracenia/pull/1776.. which has not been merged in the main line code yet and still has pending issues to be addressed. It also adds separate functionality changes in the subscription.py configuration class.

Please adjust your branch so that it adds the same functionality changes as in https://github.com/robjarawan/sarracenia/pull/173.

If your PR adds more unit tests from your source repositories' source branch, that's likely not a problem. Added unit tests are welcome.

But adding functionality changes that are already addressed in the other PRs and that are not part of the same scope of your original changes can't be merged via this PR.

PR #1781, opened 2026-09-15
Issue1764

vik164:


andreleblanc11 (changes requested):
I think the logic looks good. I just have one suggestion.

This change will make it so `_tag_progress` returns `True` on a successful operation and `False` on a failed operation. Could you please add a comment in the methods description to describe this change?

Also, adding unit tests is encouraged (for your case, you could add them in `tests/sarracenia/sr_test.py`). AI is pretty good at creating unit tests 🙂

vik164:
Updated the tag_progress method description and added tests for both altered fucntions

tests/sarracenia/sr_test.py
```diff
@@ -1,6 +1,169 @@
...
+        }
+    }
+
+    marker = tmp_path / "subscribe" / "amis" / "starting"
+    marker.parent.mkdir(parents=True)
+    marker.write_text("timestamp")
+
+    assert sr._tag_progress("subscribe", "amis", "starting", False) is False
+
+def test_tag_progress_returns_false_when_removing_missing_marker(tmp_path):
```
andreleblanc11, line 60: The title is a bit confusing. We're actually not removing the missing marker because it doesn't exist, but returning False in this scenario is correct.

andreleblanc11 (approved):

PR #1786, opened 2026-09-22
Issue1782

vik164:
Added methods and tests


sarracenia/sr.py
```diff
@@ -466,10 +466,11 @@ def _read_state_dir(self):
...
                     #print( f" {self.configs[c][cfg]['statehost']=} " )
-                    if 'options' in self.configs[c][cfg] and self.configs[c][cfg]['options'].statehost:
-                        state_dir=self.user_cache_dir + os.sep + self.hostdir + os.sep + c + os.sep + cfg
-                    else:
-                        state_dir=self.user_cache_dir + os.sep + c + os.sep + cfg
+                    # if 'options' in self.configs[c][cfg] and self.configs[c][cfg]['options'].statehost:
+                    #     state_dir=self.user_cache_dir + os.sep + self.hostdir + os.sep + c + os.sep + cfg
+                    # else:
+                    #     state_dir=self.user_cache_dir + os.sep + c + os.sep + cfg
+                    state_dir=self._statehost_dir(c, cfg)
```
andreleblanc11: `_statehost_dir` as the method name is a bit vague.. `_set_state_dir` makes more sense IMO. `statehost` makes reference mostly to the host that is running the configuration in sr3 options terms.

vik164: I agree. I will update it

sarracenia/sr.py
andreleblanc11, line 2344: The new method should be called from here as well. Doing an `sr3 start` with a disabled configuration with `statehost on` is currently not ignoring disabled status.

andreleblanc11, line 896: Your new method should also be called here as well. Otherwise the status doesn't get properly reported back via `sr3 status`.

I did a `%s/os.path.exists(self.user_cache_dir + os.sep + c + os.sep + cfg + os.sep/os.path.exists(state_dir + os.sep/g` in `vim` and then tested and that seemed to have done the trick.

